PR #2, opened 2020-07-27
Add Dockerfile

shaananc:
I've taken the liberty of making a Dockerfile that allows for one-push testing of head scale.

It includes both Postgres and headscale (a questionable decision, but it was the right one at least for my use case)

Not super high quality code, but for the state of headscale at the moment it's probably appropriate.

Thanks!


juanfont:
Merged. Cheers!

PR #6, opened 2021-04-21
Add additional locations to look for the config.json file

cure:
Make headscale look for its config file in 

  /etc/headscale
  $HOME/.headscale

in addition to the current directory.


juanfont:
Awesome! Thanks :)

PR #8, opened 2021-04-23
Handle relative paths in private_key_path and derp_map_path

cure:
If the configuration file contains relative paths (e.g. in the values for private_key_path or derp_map_path), interpret them as relative to the directory where the config.json file was read from.


juanfont:
Makes sense :) Thanks!

PR #9, opened 2021-04-24
Add support for TLS to the web service.

cure:
Add supporrt for TLS to the web service (automatic with Let's Encrypt, or via externally managed certificate/key files). 

This PR als has a small bugfix for absPath, added error checking after calling the Serve() function, and a new configuration reference section in the README.md file.


juanfont:
This is fantastic!!

PR #11, opened 2021-04-24
fix: remove db twice

xpzouying:
* fix: remove db twice.
* better for new error, use errors.New.
* Error strings should not be capitalized or end with punctuation.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings


juanfont:
Looks good! Thanks!

PR #13, opened 2021-04-25
Add a Makefile with a few targets.

cure:
The default is 'build'. The build target calls the new version-at-commit.sh script which will automatically populate the version variable inside the Headscale binary.

Once we start tagging releases on the git tree, that will come in handy.

The Makefile also has a 'test' target (does nothing yet, no tests yet) and a 'dev' target, which runs linters, tests, and finally builds.


juanfont:
Fantastic, as always!

PR #16, opened 2021-04-28
Reorganise the CLI structure

juanfont:
`cmd/headscale/headscale.go` was getting already a bit out of control. These commits changes a bit the structure for the sake of tidiness, but dumping into the `cli` package the CLI funcs.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -1,13 +1,9 @@
 package headscale
 
 import (
-	"encoding/json"
-	"errors"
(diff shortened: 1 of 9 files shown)


cure:
Oh, yes, thank you! I was thinking along the same lines when I was adding tests. Thanks for this cleanup!

juanfont:
Thanks for the input!

Issue #14, opened 2021-04-26
anyway to collect nodes status ?


changchichung:
is there any way to get nodes status from headscale ?

juanfont:
@changchichung hey! I will rework soon the CLI to expose more info on the general status.

I will let you know ASAP.

PR #19, opened 2021-05-01
List nodes in namespace + register under node command 

juanfont:
Implements #14 (need to add more info + json output, but basics are there). Also `register` cmd is now under `node` (I reckon makes more sense)


cure:
This looks good, thanks!

PR #20, opened 2021-05-02
Preauthkeys improvements

cure:
This fixes a bug in preauthkeys and adds tests. The tests use sqlite3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,5 +20,5 @@ lint:
 	golangci-lint run
 
 compress: build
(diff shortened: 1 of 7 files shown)


juanfont:
This looks awesome. My only concern is how we can handle the `postgres.Jsonb` type https://github.com/juanfont/headscale/blob/main/machine.go#L32

I am merging it anyway, as supporting sqlite is definitely a good idea...

PR #24, opened 2021-05-09
Add a DestroyNamespace command and tests for the Namespace functions.

cure:
This also reorganizes the test infrastructure a bit, moving the shared functions to app_test.go


juanfont:
Looks good!

PR #25, opened 2021-05-09
Added the first integration test with the official Tailscale client

juanfont:
Ok, so these series of commits add the first integration test with the official Tailscale client.

Nothing too complex:

- Install tailscale
- Setup headscale, create a namespace, and an authkey
- Register the tailscale client using the authkey
- Check that headscale reports one, and only one, node

I am aware that this is very basic, but sets the foundations to evolve this :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,20 +107,20 @@ jobs:
       run: |
         ./headscale serve &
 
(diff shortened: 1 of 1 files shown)


juanfont:
I am closing this PR. This is an evolutionary dead end ;)

PR #26, opened 2021-05-12
Add more tests

cure:
Nothing special, just adding some more test coverage.


juanfont:
Um, there is something wrong with the go.sum (?)

cure:
That was interesting! It's fixed now, the commit explains what happened.

PR #28, opened 2021-05-15
Adding SQLite support

juanfont:
This WIP PR adds SQLite support in preparation for the integration tests, as described in https://github.com/juanfont/headscale/pull/25#issuecomment-837643844.

Both Postgres and SQLite will remain supported.


app.go
```diff
@@ -60,11 +62,22 @@ func NewHeadscale(cfg Config) (*Headscale, error) {
...
 	}
 	pubKey := privKey.Public()
+
+	var dbString string
+	switch cfg.DBtype {
+	case "postgres":
+		dbString = fmt.Sprintf("host=%s port=%d dbname=%s user=%s password=%s sslmode=disable", cfg.DBhost,
+			cfg.DBport, cfg.DBname, cfg.DBuser, cfg.DBpass)
+	case "sqlite3":
+		dbString = cfg.DBpath
```
cure, line 72: Question - what do we want the default path to be for the sqlite db? Should we adopt the same logic as for the config file here and wrap this in an absPath() call? It's a little weird to have database files in a config directory though...

README.md
```diff
@@ -55,7 +55,12 @@ Suggestions/PRs welcomed!
   ```shell
   wg genkey > private.key
   wg pubkey < private.key > public.key  # not needed
-  cp config.json.example config.json
+
```
cure, line 58: There is a trailing space here.

cure (commented):
This is great!! What should we do about the default location of the sqlite3 db (also what is in the config.json.sqlite.example file)? Having a ./ in there seems not ideal.

cure (approved):
Looks good, thanks for making absPath more portable!

juanfont:
Thanks!

PR #29, opened 2021-05-23
Add support for ephemeral nodes via a special type of pre-auth key.

cure:
Also add some tests for that feature.

Other fixes: clean up a few typos in comments. Fix a bug that caused the tests to run four times each. Be more consistent in the use of log rather than fmt to print errors and notices.


juanfont:
This is just awesome! I am learning so much :)

PR #33, opened 2021-06-05
Handle client sending new NodeKey 

juanfont:
This commit changes a bit how we handle the registration workflow. Hopefully solves #32.

### DIFF
--- a/api.go
+++ b/api.go
@@ -136,7 +136,7 @@ func (h *Headscale) RegistrationHandler(c *gin.Context) {
 
 	// The NodeKey we have matches OldNodeKey, which means this is a refresh after an key expiration
 	if m.NodeKey == wgcfg.Key(req.OldNodeKey).HexString() {
-		log.Printf("[%s] We have the NodeKey in the database. This is a key refresh", m.Name)
+		log.Printf("[%s] We have the OldNodeKey in the database. This is a key refresh", m.Name)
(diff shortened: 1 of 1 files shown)


api.go
```diff
@@ -390,66 +428,43 @@ func (h *Headscale) getMapKeepAliveResponse(mKey wgcfg.Key, req tailcfg.MapReque
...
-		m.IPAddress = ip.String()
-		m.NamespaceID = pak.NamespaceID
-		m.AuthKeyID = uint(pak.ID)
-		m.RegisterMethod = "authKey"
-		m.Registered = true
-		db.Save(&m)
-
-		resp.MachineAuthorized = true
-		resp.User = *pak.Namespace.toUser()
+	pak, err := h.checkKeyValidity(req.Auth.AuthKey)
```
cure, line 433: Can we add a check before this line that req.Auth is not nil? Otherwise, if it ever were nil, we'd get a runtime panic.

cure (approved):
Just one small nit, looks good otherwise, thanks!

api.go
juanfont: But req.Auth cannot be nil https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L602

cure: OK that's fair, never mind then thanks.

PR #35, opened 2021-06-13
Use goreleaser to generate binary builds

juanfont:
For the time being only on new tags...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - "*" # triggers only if push new tag version
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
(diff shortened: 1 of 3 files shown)


cure (commented):
I think we're missing a .goreleaser.yml file in the repo. I propose one with these contents:

```
builds:
  - main: cmd/headscale/headscale.go
    mod_timestamp: '{{ .CommitTimestamp }}'
    goarch:
      - amd64
      - arm64
    goos:
      - linux
```
Only the line that sets `main` is strictly necessary. The `mod_timestamp` line is nice so that we make reproducible builds. The `goarch` and `goos` values are there to limit the number of targets we build for.

I tested this in a dummy repo by playing around with `goreleaser --snapshot --skip-publish --rm-dist` and after setting a local tag, `goreleaser --skip-publish --rm-dist`.

juanfont:
Brilliant move on my side, I left the most important file behind :(

cure (approved):
LGTM!

PR #38, opened 2021-06-17
Add k8s deployment, standalone app Dockerfile.

cmars:
Tested with Rancher k3s and a couple of NixOS containers running tailscale clients against it.


juanfont:
This looks excellent to me! 

@cure, what do you think?

juanfont:
Let's start with this :)

PR #44, opened 2021-07-04
Add support for Policy ACLs

juanfont:
This series of commits adds support for Tailscale policy ACLs, as described in https://tailscale.com/kb/1018/acls

I am opening the PR for reviewing purposes, as it is not yet ready to merge. Documentation is still missing too.


juanfont:
Gonna move this forward, so we can jump :)

PR #46, opened 2021-07-10
Update DERP server definitions

Teteros:
The repo link referenced in derp.yaml is invalid and is missing the new DERP addresses.
This PR updates the file to use the newest tailscale DERP servers and updates the link in the comment.

Out of curiousity is `stunonly` always `false`, and `derptestport` always `0`?
`stunport`, `stunonly`, and `derptestport` have no references in the code so not sure if these fields are still needed.


juanfont:
Looks good! I didn't actually include all official Tailscale DERP servers because people should be running their own DERP infra. But it does not make sense to just include some of them...

The fields you are mentioning are just part of the DERPNode struct https://github.com/tailscale/tailscale/blob/main/tailcfg/derpmap.go#L87

Issue #53, opened 2021-07-16
Allow to use letsencrypt on port other than 443


viq:
I want to run headscale as a dedicated user, which precludes it from running on privileged port. I forward port 443 to 8000 using firewall, thus allowing verification from letsencrypt.

juanfont:
Implemented!

PR #55, opened 2021-07-17
Turn the combination of TLS-ALPN-01 and listen_addr on a port other than

cure:
443 into a warning, not an error, refs #53.


juanfont:
Looks great!

PR #57, opened 2021-07-19
Added `init` subcommand to generate base config

juanfont:
`headscale init sqlite` generates a working headscale environment (i.e., config.json, derp.yaml and wireguard private.key)

This helps my life a lot when developing in a reusable Vagrant environment.

### DIFF
--- a/cmd/headscale/cli/init.go
+++ b/cmd/headscale/cli/init.go
@@ -1,0 +1,75 @@
+package cli
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/juanfont/headscale"
+	"github.com/spf13/cobra"
+	"tailscale.com/types/wgkey"
+)
+
+var InitCmd = &cobra.Command{
+	Use:   "init",
+	Short: "Creates a basic Headscale env",
+}
+
+var InitSqliteCmd = &cobra.Command{
+	Use:   "sqlite",
+	Short: "Creates a headscale env using SQLite as database",
+	Run: func(cmd *cobra.Command, args []string) {
+		force, _ := cmd.Flags().GetBool("force")
+		if !force {
+			if _, err := os.Stat("config.json"); err == nil {
+				fmt.Println("config.json already exists")
+				return
+			}
+			if _, err := os.Stat("private.key"); err == nil {
+				fmt.Println("private.key already exists")
+				return
+			}
+			if _, err := os.Stat("derp.yaml"); err == nil {
+				fmt.Println("derp.yaml already exists")
+				return
+			}
+		}
+
+		fmt.Println("Creating config.json")
+		cfg := `{
+		"server_url": "http://127.0.0.1:8000",
+		"listen_addr": "0.0.0.0:8000",
+		"private_key_path": "private.key",
+		"derp_map_path": "derp.yaml",
+		"ephemeral_node_inactivity_timeout": "30m",
+		"db_type": "sqlite3",
+		"db_path": "db.sqlite",
+		"tls_letsencrypt_hostname": "",
+		"tls_letsencrypt_cache_dir": ".cache",
+		"tls_letsencrypt_challenge_type": "HTTP-01",
+		"tls_cert_path": "",
+		"tls_key_path": "",
+		"acl_policy_path": ""
+	}`
+		err := os.WriteFile("config.json", []byte(cfg), 0644)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


viq:
You know you can quite easily have Vagrant pre-configure the machine for you? ;) Not negating this feature, but if you want I could help you with the Vagrant config.

juanfont:
> You know you can quite easily have Vagrant pre-configure the machine for you? ;) Not negating this feature, but if you want I could help you with the Vagrant config.

I know! :) But for convenience I'd like to merge it. It would be a common solution for both Vagrant or classic environments. Also, removes the dependency on `wireguard-tools` packages, as you can generate the private.key with headscale CLI

cmd/headscale/headscale.go
```diff
@@ -39,9 +39,10 @@ https://gitlab.com/juanfont/headscale`,
 func main() {
 	err := cli.LoadConfig("")
 	if err != nil {
-		log.Fatalf(err.Error())
```
cure, line 42: How will this affect normal operation, i.e. not the `init` path? Do we have another check further in the code for a non-existent config?

cure (approved):
Conceptually, I see no issues, this PR makes sense. Just one comment about the change in behavior when no config file is found.

cmd/headscale/cli/init.go
```diff
@@ -0,0 +1,75 @@
...
+			if _, err := os.Stat("derp.yaml"); err == nil {
+				fmt.Println("derp.yaml already exists")
+				return
+			}
+		}
+
+		fmt.Println("Creating config.json")
+		cfg := `{
+		"server_url": "http://127.0.0.1:8000",
+		"listen_addr": "0.0.0.0:8000",
```
ohdearaugustin, line 41: Why by default listen on *? I rather would prefer a more strict default, like 127.0.0.1. Even thou it is also the default in the config examples.

cmd/headscale/cli/init.go
```diff
@@ -0,0 +1,75 @@
...
+		fmt.Println("Generate the Wireguard private.key")
+		privk, err := wgkey.NewPrivate()
+		if err != nil {
+			log.Fatal(err)
+		}
+		err = os.WriteFile("private.key", []byte(privk.String()), 0644)
+		if err != nil {
+			log.Fatal(err)
+		}
+
```
ohdearaugustin, line 68: Maybe also generate public key here and write it.

cmd/headscale/headscale.go
ohdearaugustin: I would also see this change as important, because you won't be able to even run the help command without a proper config setup up. 
I would agree to @cure some more check in the code would be needed to have a valid configuration for the other commands, beside help, version and init.

ohdearaugustin (commented):
Further more I would like to open up a discussion here if it is not maybe a better idea to use setting defaults in viper to make a runable version of headscale. Adding the functionality of init would make it even harder to run headscale without the need of config file.

I would love to be able to configure headscale just via environment vars. which would make it even easier to run it with kubernetes or other containerization. As you mentioned Vagrant it would also be a nice way to configure headscale their only via environment vars making the whole app more selfcontained.

Hope this is not the wrong place for those thoughts.

juanfont:
@cure @ohdearaugustin 

You are very right. I do agree that the idea of this PR makes sense, and its quite convenient (in particular removing the need for `wireguard-tools` to be installed to generate the wg key). However, how the configuration is defined in a string + removing the check on a valid config at Headscale start is suboptimal / crap 😅

I do agree with @ohdearaugustin on dropping the need for a `config.json` file altogether. Running only on ENV vars is perfectly feasible, and pretty easy with viper.

I am closing this PR, and will think on a better way to implement this.

viq:
Not that there's much secrets in configuration besides the database
password, but oftentimes files are easier to keep secret than environment
of a process, so if possible I'd rather keep the ability to feed it a
config file.

ohdearaugustin:
@viq I didn't meant to replace the ability to use a config file. This should still be kept. I rather see it as addition to make headscale only configurable from the environment. As a lot of containerization take use of it and it is easier to configure. File permission are in my opinion rather harder to get right with containerization. For security reason you can use for example systemd and lock it down really a lot, so the environment is also secured.

Therefore move the default setting into the program itself. I was thinking about using viper with default values to make the executable runable without any additional configuration file needed.

PR #59, opened 2021-07-23
Add a 'tls_letsencrypt_listen' config option

qbit:
Currently the default (and non-configurable) Let's Encrypt listener will
bind to all IPs. This isn't ideal if we want to run headscale on a specific
IP only.

This also allows for one to set the listener to something other than
port 80. This is useful for OSs like OpenBSD which only allow root to
bind the lower port ranges (and don't have `setcap`) as we can now run
`headscale` as a non-privileged user while still using the baked in ACME
magic. Obviously this configuration would also require a reverse proxy
or firewall rule to redirect traffic. I attempted to outline that in the
README change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,12 +157,14 @@ Headscale can be configured to expose its web service via TLS. To configure the 
 
 ```
     "tls_letsencrypt_hostname": "",
+    "tls_letsencrypt_listen": ":http",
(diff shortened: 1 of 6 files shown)


cure:
Thank you, I've merged this. I will followup with some more README changes.

Issue #52, opened 2021-07-16
Ability to remove a node


viq:
Currently nodes can only be added. I'd like to be able to remove nodes that don't exist anymore (without going into SQL myself).

juanfont:
Done! Although you might have to reload headscale for changes to take effect. We will be working on that next.

viq:
Seems also the same is true for ephemeral nodes - they get cleaned up only on headscale restart.

juanfont:
That really sounds like something is being cached. Are you serving this
through Cloudflare by any chance?

viq:
No, purely my infra: couple OpenBSD boxes on libvirt on a dedicated server,
and all the pieces are running on them. The only possible caching is via
nginx, and I thought I disabled it in config.

juanfont:
Can you make sure all the requests arriving to nginx are fwded to
Headscale? It sounds very weird...

viq:
I just verified with `tailscaled` on a couple machines connecting directly to `headscale`'s port `8000`, and even when `headscale` logged removal of the ephemeral node, all other machines still show it listed in `tailscale status`.

viq:
Though I'll add this is with `0.3.4`, I didn't try building `master`. Would
you expect different result there?

juanfont:
I will tag a new release as soon as https://github.com/juanfont/headscale/pull/63 is merged. Apologies for the late response, busy week :(

viq:
Thanks :)
No need to apologise, those are very quick responses for my standards ;)

juanfont:
This is fixed now :)

djkazic:
I'm seeing this behavior on headscale `0.22.3`.

I'd register a node A, then once I see it in `headscale nodes list`, remove it with `headscale nodes delete -i 5`. Confirmed it's gone from the list, but `tailscale status` on my other machines would still have it there, and it could continue to network.

Only when I restart the tailscale daemon does the connection drop. If I restart the tailscale app of the removed node, similarly, it updates then.

I looked at the code and there doesn't seem to be any polling now, but rather usage of a mapper. Does that mapper suffer from the same issue as what was seen here?

Context: I wanted to write some code that would be able to eject a node from the tailnet, but this doesn't seem possible.

PR #63, opened 2021-07-25
Added communication between Serve and CLI using KV table 

juanfont:
The in-memory `clientsPolling` map is only available from the Headscale process serving the API. This means that some changes performed by the CLI (which is a different process) are not propagated immediately to the polling clients. 

This is what is happening in #52. Nodes are successfully removed from the DB, but the connected machines do not receive an updated node `Map` until their client decides to close the long poll and reconnect. 

I have added a new worker that polls the KV table we had since the beginning in Headscale DB schema. Namespaces whose nodes should be sent an updated map are added to the `namespaces_pending_updates` key. 

The hardcoded polling time is 5000 msecs.

### DIFF
--- a/app.go
+++ b/app.go
@@ -139,6 +139,20 @@ func (h *Headscale) expireEphemeralNodesWorker() {
 	}
 }
 
+// WatchForKVUpdates checks the KV DB table for requests to perform tailnet upgrades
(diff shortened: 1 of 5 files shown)


juanfont:
I am not 100% happy with this, as it looks a bit convoluted. But it works(tm).

PR #69, opened 2021-07-30
Use 8080 as default port in the example config

### LINKED ISSUE #68: Docker port issue or by design?
Thanks for the new dockerfile :)

I wondered why you are exposing 8080/tcp from the docker but use 8000/tcp in the sample config files.

Are both ports used or we should we update and use 8080 as the default?

Thanks.

juanfont:
We had :8000 in the examples, while the Dockerfile and Kubernetes YAMLs use :8080. 

This PR changes everything to :8080, as it is the port that had actually running software pointing at it.

This fixes #68.


juanfont:
I'll go forward with the merge, as it is rather trivial.

PR #72, opened 2021-08-02
Make IP Prefix configurable and available ip deterministic

kradalby:
Hi

This PR contains two commits were the first (making ip prefix configurable) one can be cherry picked if the second (rework get available ip) is not desirable.

Please see the individual commit messages for details.

The code has some tests, but real world testing help would be appreciated.

If it is not desirable to change the IP allocation, then I can rework it into an alternative to the original, so users can opt in/out.


juanfont (approved):
This is just fantastic, and a great idea!

Thank you so much!

PR #77, opened 2021-08-06
Remove more deadlocks

kradalby:
This PR removes more deadlocks, one I forgot and on (from keepAlive) that I could not understand why was there.

Ideally we need some wireup tests to test for locks and check actual tailscale(d) behaviour over time.


juanfont:
Excellent!

PR #81, opened 2021-08-08
Add Integration tests

kradalby:
This PR builds the basis for integration testing the built `headscale` binary and against the official Tailscale client.

please review commit by commit, as each commit describe the given change.

These has been tested with Docker for mac, so would appreaciate someone testing with Linux:

Via `make`:

```bash
make test_integration
```

Or via `go` (with a lot more verbose output, which I prefer):

```bash
go test -tags integration
```

The tests are as the command above shows, flag guarded so they will not run without being explicitly invoked.


Dockerfile.tailscale
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:latest
+
+RUN apt-get update \
+    && apt-get install -y gnupg curl \
+    && curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/focal.gpg | apt-key add - \
+    && curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/focal.list | tee /etc/apt/sources.list.d/tailscale.list \
+    && apt-get update \
+    && apt-get install -y tailscale \
```
juanfont, line 8: Um. We could have here an `ARG` to do integration tests with multiple Tailscale versions, right?

kradalby: Yes, I would argue that could be an improvement for when it is hooked up with ci, so we can do a matrix

juanfont:
This is awesome! And tests are passing :)

PR #84, opened 2021-08-13
Improve logic to keep nodes up to date with the network state

kradalby:
This PR is a two parter:

- Hook up the integration tests
  - Checking that all nodes is a part of the `tailscale status` of all other nodes
  - Ping all nodes from each node directly
- Fix the issues making the integration tests fail

This PR addresses a series of issues with the update logic making us able to ensure that nodes stay up to date when the network mesh changes. Previously it managed to push out changes sporadically and not consistently. 

This is done by:
- Ensuring updates are issued on omitPeer=true, stream=false (nodes use that to update their addresses)
- Ensure the update channel actually is consumed by the stream function
- Periodically issue an update request to all nodes

In addition, to not send excess amount of traffic to all nodes with unnecessary update, a simple "state" has been introduced with a timestamp the last time relevant changes was made. Updates will only be sent if the nodes have not had a successful update written to their long poll after the last update timestamp.

This PR should bring headscale and the integration tests into a state were we can reliably detect regressions that make the application not work as intended for our users :). Hopefully we can make changes with more confidence.

### DIFF
--- a/app.go
+++ b/app.go
@@ -60,8 +60,7 @@ type Headscale struct {
 
 	clientsUpdateChannels sync.Map
 
-	lastStateChangeMutex sync.RWMutex
-	lastStateChange      time.Time
+	lastStateChange sync.Map
 }
 
 // NewHeadscale returns the Headscale app
@@ -88,13 +87,12 @@ func NewHeadscale(cfg Config) (*Headscale, error) {
 	}
 
 	h := Headscale{
-		cfg:             cfg,
-		dbType:          cfg.DBtype,
-		dbString:        dbString,
-		privateKey:      privKey,
-		publicKey:       &pubKey,
-		aclRules:        &tailcfg.FilterAllowAll, // default allowall
-		lastStateChange: time.Now().UTC(),
+		cfg:        cfg,
+		dbType:     cfg.DBtype,
+		dbString:   dbString,
+		privateKey: privKey,
+		publicKey:  &pubKey,
+		aclRules:   &tailcfg.FilterAllowAll, // default allowall
 	}
 
 	err = h.initDB()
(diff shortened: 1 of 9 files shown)


machine.go
```diff
@@ -251,21 +260,67 @@ func (m *Machine) GetHostInfo() (*tailcfg.Hostinfo, error) {
...
+
+func (h *Headscale) requestUpdate(m *tailcfg.Node) error {
+	pUp, ok := h.clientsUpdateChannels.Load(uint64(m.ID))
+	if ok {
+		log.Info().
+			Str("func", "requestUpdate").
+			Str("machine", m.Name).
+			Msgf("Notifying peer %s", m.Name)
+
+		if update, ok := pUp.(chan struct{}); ok {
```
juanfont, line 294: what happens if the machine closes the channel here? just after the `Load` from the map?

kradalby: It can, and during testing, I get it occasionally, from testing this far it happens about 1/100 integration test runs.

Which is still to much, so this branch will have to remain draft for now, potentially we have to merge, to get some benefits, and then create an issue to solve it.

1/100 is better than the 97/100 failed runs I had when I started :P

kradalby:
I think this PR is more or less ready, please review (carefully).

I think the integration tasts _can_ still fail with two things:
- Nodes not managing to sync up properly, needing some more cycles to sort that
- tailscale ping fails because of docker networking

However, after running it a few hundred times, I see roughly 1% fake failures and I think the benefits might outweigh the potentials for false positives.

machine.go
```diff
@@ -211,6 +213,13 @@ func (h *Headscale) GetMachineByID(id uint64) (*Machine, error) {
 	return &m, nil
 }
 
+func (h *Headscale) UpdateMachine(m *Machine) error {
```
juanfont, line 216: What is this function used for? It seems quite unused / no-op

kradalby: It is quite used:
https://github.com/kradalby/headscale/blob/integration-tests-ci/poll.go#L240
https://github.com/kradalby/headscale/blob/integration-tests-ci/poll.go#L286
https://github.com/kradalby/headscale/blob/integration-tests-ci/poll.go#L350
https://github.com/kradalby/headscale/blob/integration-tests-ci/poll.go#L350

It is the discovery (and initial mitigation) of #93

juanfont (approved):

juanfont:
This is just awesome. The code and how well the PR was structured.

app.go
```diff
@@ -165,9 +168,18 @@ func (h *Headscale) Serve() error {
 	r.POST("/machine/:id", h.RegistrationHandler)
 	var err error
 
+	timeout := 30 * time.Second
```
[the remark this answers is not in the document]
kradalby: I dont have any tests to "prove it", but observing the run of the tests, it does not seem like it breaks that often. 

However, I would appreciate if someone pulls it down and has a look

How I looked at it:

The timeout is 30s,
Which if it breaks longpolling should mean that every other keepAlive should break on _all_ nodes. I don't think it does, but please help me confirm. 

We can adjust it accordingly as we learn more. At the moment, I would argue that its better to have all the clients in sync than the _most_ efficient implementation.

PR #94, opened 2021-08-21
Split lint and test CI files

kradalby:
This commit splits the lint and test steps into two different jobs in
github actions.

Consider this a suggestion, the idea is that when we look at PRs we will
see explicitly which one of the two types of checks fails without having
to open Github actions.

image


ohdearaugustin (approved):
It looks good to me and absolutely makes sense to split it.

PR #96, opened 2021-08-24
Fix setting of version

qbit:
Make version-at-commit.sh work on systems that don't have bash in `/bin/bash` while here.


juanfont:
Great! Thanks!

PR #107, opened 2021-09-12
Remove trace lines about NO_COLOR.

qbit:
Currently there is a warning for every command run when NO_COLOR is set. For example:

```
$ headscale nodes 
{"level":"trace","time":"2021-09-12T07:31:48-06:00","message":"NO_COLOR is set, disabling colors"}
Manage the nodes of Headscale

Usage:
  headscale nodes [command]

Available Commands:
  delete      Delete a node
  list        List the nodes in a given namespace
  register    Registers a machine to your network
  share       Shares a node from the current namespace to the specified one

Flags:
  -h, --help               help for nodes
  -n, --namespace string   Namespace

Global Flags:
  -o, --output string   Output format. Empty for human-readable, 'json' or 'json-line'

Use "headscale nodes [command] --help" for more information about a command.
$
```

This disables that.

### DIFF
--- a/cmd/headscale/headscale.go
+++ b/cmd/headscale/headscale.go
@@ -22,14 +22,12 @@ func main() {
 		colors = true
 	default:
 		// no color, return text as is.
-		log.Trace().Msg("Colors are not supported, disabling")
 		colors = false
 	}
 
 	// Adhere to no-color.org manifesto of allowing users to
 	// turn off color in cli/services
 	if _, noColorIsSet := os.LookupEnv("NO_COLOR"); noColorIsSet {
-		log.Trace().Msg("NO_COLOR is set, disabling colors")
 		colors = false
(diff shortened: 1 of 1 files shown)


kradalby:
I don't have strong opinions of this, iirc this might be set up before the log level?

If it can disabled by setting lower than tracing, I would suggest that rather than removing.

qbit:
I am using the default log level - and the output happens on the command line.

Personally - I think it's annoying. I know I have color disabled, I set NO_COLOR explicitly and any thing that is informing me of that is just adding noise.

qbit:
by "default log level" I mean I have nothing explicitly set in my config.json that increases or decreases the logging.

kradalby (approved):
Your right, the lines are printed before we set the log level, let's just remove them

PR #108, opened 2021-09-13
Dockerfile: add golang tag

### LINKED ISSUE #95: Dockerfile Latest Version
I would never use the latest tag in a docker file, because it is unstable.
You will never now with which version you end up. For example ubuntu:latest is for now ubuntu:focal.
I would prefer to use explicit versions.

Same for the golang container. Suddenly the container updated and headscale maybe won't build anymore.

Good blog about why latest is not so great:
https://vsupalov.com/docker-latest-tag/

ohdearaugustin:
Closes #95

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:latest AS build
+FROM golang:1.17.1-bullseye AS build
 ENV GOPATH /go
 
 COPY go.mod go.sum /go/src/headscale/


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:latest AS build
+FROM golang:1.17.1-bullseye AS build
```
kradalby, line 1:
```suggestion
FROM golang:1.17-bullseye AS build
```

How about only tagging minor so we automatically get security/patch fixes?

ohdearaugustin: would be an idea, but we would not be able anymore to rebuild old containers as they are. Reproducible builds.

ohdearaugustin: Maybe add something like depenendbot or something like that.

kradalby: Dependabot or renovate would make sense regardless.


I am not sure how go and go.mod works to the detail of reproducible builds?

ohdearaugustin: As far as I know are the dependencies  in go.mod have versions and they are not automatically updated. Also with the tags in the repo. The version of the dependency of the tag would be used.

ohdearaugustin:
Anything on that?

juanfont:
Apologies for the delay. Merged!

ohdearaugustin:
Thx (=

PR #109, opened 2021-09-14
Update to Tailscale 1.14

juanfont:
This PR updates our dependencies to Tailscale v1.14. 

Consider it as WIP, as I am still reviewing Tailscale's code looking for changes like https://github.com/tailscale/tailscale/issues/2107


kradalby:
Sounds like this could benefit from expanding the integration tests to cover all the changes cases we come across, what do you think?

Also extend them for testing the new shared node.

kradalby:
Could you merge master into this branch so we can run it against the new integration tests?

Issue #58, opened 2021-07-20
Mobile clients


ptman:
Tailscale has clients for linux, Windows, macOS, Android and iOS. Linux, Windows and macOS can be told to connect to a headscale server. I'm not aware of a way to do that on Android or iOS.

Can the (open source) android client be compiled with another server or be patched to allow server selection? What about iOS?

outbackdingo:
sounds more like a tailscale question then headscale

ptman:
Sort of, but it doesn't affect users of tailscale that don't use headscale.

juanfont:
Selecting the control server will be coming in the future for the official mobile clients (https://twitter.com/dave_universetf/status/1415046381996167170).

For the time being is possible to compile the Android client and change the hardcoded `ipn.Prefs` when calling the library.

juanfont:
I will close this for the time being...

kradalby:
For people who wonder how to tell macOS to connect to `headscale`:

Install from App Store:

```bash
/Applications/Tailscale.app/Contents/MacOS/Tailscale up -login-server <YOUR HEADSCALE> --accept-routes
```

tombh:
I wonder if it might also be useful to mention in the README that Headscale doesn't yet support mobile clients natively? Just took me a bit of digging to get here and so realise that Headscale wasn't going to be so useful in my case.

kradalby:
> I wonder if it might also be useful to mention in the README that Headscale doesn't yet support mobile clients natively? Just took me a bit of digging to get here and so realise that Headscale wasn't going to be so useful in my case.

Would be a good idea, would you be able to submit a PR?

We could have a table with os, client combos

boehs:
Could tailscale android be patched to introduce a configuration option?

Akruidenberg:
I have the same question.

artemklevtsov:
I managed to automate the process using GitLab CI.

`.gitlab-ci.yml`:
```yaml
variables:
  GIT_TAG: "1.23.148-t7c7f37342-gb4f8e7f90a4"
  GIT_REPO: "tailscale/tailscale-android"

stages:
  - docker
  - build
  - upload

.fetch-repo: &fetch-repo
  - curl -o tailscale-android.zip https://codeload.github.com/${GIT_REPO}/zip/refs/tags/${GIT_TAG}
  - unzip -q tailscale-android.zip
  - mv -n tailscale-android-${GIT_TAG}/* .
[...]

PR #113, opened 2021-09-19
Apple macOS profile support

### LINKED ISSUE #106: Usage with Tailscale on MacOS
Hey

Could add some clue how to use this with the MacOS TailScale client? It seems the GUI client needs to be loaded and logged in before I can use the CLI command, but then this mean I cannot run the following, as it is already authenticated to TailScales hosted controller.

/Applications/Tailscale.app/Contents/MacOS/Tailscale up -login-server <> --authkey <KEY>

If I run the following, it forces the client to logout, but it doesn't reconnect:

/Applications/Tailscale.app/Contents/MacOS/Tailscale up -login-server <> --authkey <KEY> --reset --force-reauth

Thank you

kradalby:
This PR will add new HTTP endpoints that can generate ~iOS~ and macOS profiles which will allow users to use Tailscales official app with their headscale server.

Currently, a new endpoint called `/apple` has been added with documentation on how to use the official Tailscale.app with macOS and override the Control plane url so you can use Headscale.

I was originally hoping that I would figure out iOS for this PR as well, but it has proven more difficult than anticipated. 

I need some more time, and I probably need to set up and understand more of Apple MDM management so I will split this to a separate PR.

This will resolve ~the iOS part of #58~ and resolves #106 (macOS)

Better macOS support will probably be useful for some users :)


jilyaluk:
I think that iOS app could be configured using a profile only if it's managed. It is not trivial to find a proof for that in docs, however [some answer on SO](https://stackoverflow.com/a/43124181) mentions that, or answer in [this thread on Apple Developer Forums](https://developer.apple.com/forums/thread/672938) states that you need a MDM server to push configuration to device.

I tried installing Tailscale on my device using a Jamf, and configuring it like that:
```
<dict>
<key>ControlURL</key>
<string>http://...</string>
</dict>
```

However, the app is still trying to login via tailscale.com. Maybe the setting is named differently, or I made a mistake in the deployment, I'll try investigating that further.

In theory, it is possible to embed MDM server (e.g. https://github.com/micromdm/micromdm) to install the app as managed and push configuration.

kradalby:
Hmm yes, I am currently split between if I am just hitting the wrong key with my testing, having the same with the PayloadType, iOS instead of macOS seems like it would make sense which makes me think you are on to something.

If you are right that it's an iOS restriction, that would be a bummer.

Atm I am thinking: not ready to give up yet, and at least the macOS setup got neater.

I'm gonna continue to hunt among possible keys, but I guess I should do a better dive into mdm vs non mdm.


I guess one thing that makes me think I just got the wrong key is that while on macOS it shows the change in the profile, iOS shows "unknown payload".

Please keep me updated with your findings.

kradalby:
After you pointed out that this kind of change would require a mdm setup, I was thinking you might be on to something. 

It sounds like something (almost sensible) would do to prevent user from ending up being tricked to override "safe" app settings...

You mentioned you used Jamf, is there a "home setup" one can use to test? Or is it because you have an enterprise setup?

jilyaluk:
I do have an enterprise setup, however, I mentioned https://github.com/micromdm/micromdm earlier, I think you could deploy it for a home setup (I am not familiar with it, but it should support managed app installation and configuration).

kradalby:
Yep I saw, it looks like it still [requires an Apple Enterprise Dev cert ($299/year)](https://github.com/micromdm/micromdm/blob/master/docs/user-guide/introduction.md#apple-requirements).

I'll have a play around to see if there is a local setup or something

jilyaluk:
Ah, that's a bummer. However, that makes sense, since MDM is tightly coupled with Apple's push notification system.

There are a plenty of MDM service providers out there, I hope some of them provide personal plans.

jilyaluk:
Btw, on macOS you could achieve the same result without mobileconfig whatsoever:
`defaults write io.tailscale.ipn.macos ControlURL http://...`

outbackdingo:
so is there a write up which better explains how to accomplish this on OS X ?

kradalby:
> Btw, on macOS you could achieve the same result without mobileconfig whatsoever:
> `defaults write io.tailscale.ipn.macos ControlURL http://...`

Yep I noticed this as well, in terms of the feature of this PR, the macOS was a "bi-product" of trying to figure out how to fix the iOS part.

My current plan is to leave the profile for macOS but improve it with a "proper how-to" and also have the defaults alternative in the documentation (as in the /apple endpoint).

> so is there a write up which better explains how to accomplish this on OS X ?
Yep, as mentioned above, I just have not had any time to work on it lately


I have some vague ideas on what to try for iOS, but if I cant figure it out the next time I sit down, I might split this PR into a "macOS first" one and then leave the iOS as draft until I figure it out.

outbackdingo:
> > Yep, as mentioned above, I just have not had any time to work on it lately

how about a quick and dirty i have an immediate need to add a MAC OS X client to my headscale network, they have no linux :(

kradalby:
> > > Yep, as mentioned above, I just have not had any time to work on it lately
> 
> how about a quick and dirty i have an immediate need to add a MAC OS X client to my headscale network, they have no linux :(

You can do what @jilyaluk suggest with defaults. 

But please keep support requests off the pull requests discussion, it makes it hard for us to filter out relevant comments on how to solve this feature.

juanfont (approved):
Go for it!

PR #115, opened 2021-09-20
Topic/fix docu

ohdearaugustin:
Fixed the docu as the docker part, which was actually written by me (Facepalm), was not working at all. Now also tested probably.


juanfont:
hahah appreciated!

Issue #118, opened 2021-09-21
Enable taildrop support from headscale?


mannp:
Hi

I wondered if it would be possible to enable taildrop support from within headscale at all?

Cheers

russkel:
More information here:
https://tailscale.com/blog/2021-06-taildrop-was-easy/

Looks like a useful addition.

juanfont:
Just landed in v0.9.1 https://github.com/juanfont/headscale/releases/tag/v0.9.1

Cheers!

mannp:
> Just landed in v0.9.1 https://github.com/juanfont/headscale/releases/tag/v0.9.1
> 
> Cheers!

Thanks for the very prompt implementation :) is taildrop enabled automatically, or needs a env or config setting ro enable like it does in tailscale?

Cheers

juanfont:
Automatically enabled. Actually, hardcoded 😅

mannp:
Thanks for the quick clarification 👍🏻 presume its selectable in tailscale as its currently showing as an alpha grade feature .... likely you know its good, but might be worth a switch until its final code in tailscale....just a thought :-)

PR #122, opened 2021-09-24
Add support for Taildrop (file sharing)

juanfont:
A very low-hanging fruit here. This PR just adds support for Taildrop (tailscale filesharing, https://tailscale.com/kb/1106/taildrop/)

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -565,22 +565,44 @@ func (s *IntegrationTestSuite) TestTailDrop() {
 						// So curl!
 						peerAPI, ok := apiURLs[ip]
 						assert.True(t, ok)
-						command := []string{
-							"curl",
-							"--retry",
-							"10",
-							"-X",
-							"PUT",
-							"--upload-file",
-							fmt.Sprintf("/tmp/file_from_%s", hostname),
-							fmt.Sprintf("%s/v0/put/file_from_%s", peerAPI, hostname),
+
+						// We still have some issues with the test infrastructure, so
+						// lets run curl multiple times until it works.
+
+						attempts := 0
+						var err error
+						for {
+							command := []string{
+								"curl",
+								"--retry-connrefused",
+								"--retry-delay",
+								"30",
+								"--retry",
+								"10",
+								"--connect-timeout",
+								"60",
+								"-X",
+								"PUT",
+								"--upload-file",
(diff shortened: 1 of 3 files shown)


juanfont:
Implements #118

qbit:
FWIW you can do HTTP stuff at a UNIX domain socket with `net/http`:


```
        socClient := http.Client{
                Transport: &http.Transport{
                        DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
                                return net.Dial("unix", "/tmp/sockpost.sock")
                        },
                },
        }

        var res *http.Response
        res, err := socClient.Post("http://unix/place/stuff", "application/octet-stream", fileReader)
        if err != nil {
                log.Fatal(err)
        }
```

This would eliminate the need for an external dependency on curl.

juanfont:
> This would eliminate the need for an external dependency on curl.

I am afraid we can't. We execute `curl` _inside_ the container where we run Tailscale.

integration_test.go
```diff
@@ -529,6 +540,110 @@ func (s *IntegrationTestSuite) TestSharedNodes() {
...
+			command := []string{"touch", fmt.Sprintf("/tmp/file_from_%s", hostname)}
+			_, err := executeCommand(
+				&tailscale,
+				command,
+				[]string{},
+			)
+			assert.Nil(s.T(), err)
+			for peername, ip := range ips {
+				s.T().Run(fmt.Sprintf("%s-%s", hostname, peername), func(t *testing.T) {
+					// We currently cant send files to so skip that
```
kradalby, line 560: This comment sounds a bit unfinished

integration_test.go
```diff
@@ -529,6 +540,110 @@ func (s *IntegrationTestSuite) TestSharedNodes() {
...
+						// Under normal circumstances, we should be able to send a file
+						// using `tailscale file cp` - but not in userspace networking mode
+						// So curl!
+						peerAPI, ok := apiURLs[ip]
+						assert.True(t, ok)
+
+						// We still have some issues with the test infrastructure, so
+						// lets run curl multiple times until it works.
+
+						attempts := 0
```
kradalby, line 572: Lets add a `TODO(juanfont):` here to remove the retry loop when we get that sorted.

kradalby (approved):

kradalby:
Looks good

PR #126, opened 2021-09-26
Initial work on OIDC (SSO) integration

unreality:
This pull adds initial support for SSO. If you add `oidc_endpoint`, `oidc_client_id` and `oidc_client_secret` to config.json, when a new machine attempts to register, the OIDC endpoint will be used for authentication. A new namespace is created using the email address of the user, and the machine is added to that namespace.

Expiry times probably need to be implemented/used so that Logging out works correctly, and i think further work needs to be done on the register endpoint to become secure.

`oidc_endpoint` needs to be the base url of the OIDC endpoint (ie if your well-known configuration is at `http://localhost:8080/auth/realms/master/.well-known/openid-configuration` your endpoint url should be `http://localhost:8080/auth/realms/master/` -- it does not do proper path joining at the moment, so the training `/` is required


kradalby:
Hi @unreality This is great!

I have not come around to starting to review the code yet, but I have built it and played a bit with it.

On the first attempt I got the following panic when the callback came back from my OIDC (I use [dex](https://dexidp.io/)):

```go
GET /oidc/callback?code=xatlvybq3mfbs3l6vm4jfdr57&state=0c22a99162dfaacbbe31287a666f41e4 HTTP/2.0
Host: headscale.example.no
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Encoding: gzip, deflate, br
Accept-Language: en-us
Referer: https://id.example.no/dex/approval?req=zepbtkldlpe7h3nrlhh7emot5
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Safari/605.1.15

runtime error: index out of range [0] with length 0
/usr/lib/go-1.17/src/runtime/panic.go:90 (0x439054)
        goPanicIndex: panic(boundsError{x: int64(x), signed: true, y: y, code: boundsIndex})
/root/git/headscale_oidc/oidc.go:81 (0xbdc3dd)
/root/git/headscale_oidc/oidc.go:256 (0xbde16b)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/context.go:165 (0x977081)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/recovery.go:99 (0x97706c)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/context.go:165 (0x9762e6)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/logger.go:241 (0x9762c9)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/context.go:165 (0x97581d)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/gin.go:489 (0x9754a5)
/root/go/pkg/mod/github.com/gin-gonic/gin@v1.7.4/gin.go:445 (0x975004)
/usr/lib/go-1.17/src/net/http/server.go:2878 (0x785bba)
        serverHandler.ServeHTTP: handler.ServeHTTP(rw, req)
/usr/lib/go-1.17/src/net/http/server.go:3479 (0x787944)
        initALPNRequest.ServeHTTP: h.h.ServeHTTP(rw, req)
/usr/lib/go-1.17/src/net/http/h2_bundle.go:5832 (0x75af77)
        (*http2serverConn).runHandler: handler(rw, req)
/usr/lib/go-1.17/src/runtime/asm_amd64.s:1581 (0x46bc40)
        goexit: BYTE        $0x90        // NOP

```

I have not had any time to dig into what can cause this and will have a look tomorrow if no one beats me to it. Heck, it could even be me having misconfigured something.

Other than that I think it is a good start, but I would make a little list of things to discuss before merging (@juanfont, @qbit):

- We would probably need to handle the log out/expiry before merger
- We need to understand what kind of work needs to be done for it to be secure
- Automatically making namespace based on email probably make sense, but I need to think a bit about it

unreality:
That trace is probably a bug on my part, I have made an assumption about the JWKs returned by the OIDC server. I'll see if I can spin up a Dex server and do some testing. All my testing has been done using Keycloak.

* I agree that logout/expiry should probably be implemented before merging. I dont think it would be too hard to implement basic expiry.
* My main concern with making it secure is the expiry times, and also adding a confirmation when new machines are added. It also looks like there is also a state in which registration is requested, but since the machine is already registered, it skips any change to the machine in the database. I think Machine.Registered usage needs to be clarified/cleaned up?
* An additional todo is also to grab the groups and/or other data from the OIDC server and make a decision on namespace based on that, but i think email address would be sufficient for now?

kradalby:
> That trace is probably a bug on my part, I have made an assumption about the JWKs returned by the OIDC server. I'll see if I can spin up a Dex server and do some testing. All my testing has been done using Keycloak.

That would be good, we want to ensure that we are compliant if we are advertising OpenID Connect.

> * I agree that logout/expiry should probably be implemented before merging. I dont think it would be too hard to implement basic expiry.

Sounds good

> * My main concern with making it secure is the expiry times, and also adding a confirmation when new machines are added. 

Do you mean expiry as in on the server-side?

> It also looks like there is also a state in which registration is requested, but since the machine is already registered, it skips any change to the machine in the database. I think Machine.Registered usage needs to be clarified/cleaned up?

There is quite a lot of "hard-to-follow" stuff with the registration and all the database calls all over the place. It is on my cleanup list, but time. 

I'll try to have a look at that while reviewing.

> * An additional todo is also to grab the groups and/or other data from the OIDC server and make a decision on namespace based on that, but i think email address would be sufficient for now?

I agree, I think email _should_ be good, or at least a good default. 

My concern is that I would concern headscale a _hobbyist_ project, which means that a setup might have users from all over the place, domain-wise.
And if you use gmail.com, and your friend uses gmail.com, this is a problem if you share a headscale instance.

I dont have a good solution as of now, maybe I can think of something while reviewing the code.

unreality:
This new build should work with Dex now, i replaced my code with go-oidc. I'll try to tackle nodekey expiration etc as well, unless you want to separate that into a separate pull request?

kradalby:
Let's make it feature complete before we merge :), so implementing expiry and logout makes sense.

Great work on using the upstream lib! I'll take a look. Sorry for the delay I was hunting some other bugs and got distracted.

api.go
```diff
@@ -133,8 +133,13 @@ func (h *Headscale) RegistrationHandler(c *gin.Context) {
...
 			Msg("Not registered and not NodeKey rotation. Sending a authurl to register")
-		resp.AuthURL = fmt.Sprintf("%s/register?key=%s",
-			h.cfg.ServerURL, mKey.HexString())
+
+		if h.cfg.OIDCIssuer != "" {
+			resp.AuthURL = fmt.Sprintf("%s/oidc/register/%s", h.cfg.ServerURL, mKey.HexString())
+		} else {
+			resp.AuthURL = fmt.Sprintf("%s/register?key=%s",
+				h.cfg.ServerURL, mKey.HexString())
+		}
```
kradalby, line 142:
```suggestion
		if h.cfg.OIDCIssuer != "" {
			resp.AuthURL = fmt.Sprintf("%s/oidc/register/%s", strings.TrimSuffix(h.cfg.ServerURL, "/"), mKey.HexString())
		} else {
			resp.AuthURL = fmt.Sprintf("%s/register?key=%s",
				strings.TrimSuffix(h.cfg.ServerURL, "/"), mKey.HexString())
		}
```

Lets handle the possible scenario of users having a serverUrl with a trailing `/`

api.go
```diff
@@ -199,8 +204,12 @@ func (h *Headscale) RegistrationHandler(c *gin.Context) {
...
 		Str("machine", m.Name).
 		Msg("The node is sending us a new NodeKey, sending auth url")
-	resp.AuthURL = fmt.Sprintf("%s/register?key=%s",
-		h.cfg.ServerURL, mKey.HexString())
+	if h.cfg.OIDCIssuer != "" {
+		resp.AuthURL = fmt.Sprintf("%s/oidc/register/%s", h.cfg.ServerURL, mKey.HexString())
+	} else {
+		resp.AuthURL = fmt.Sprintf("%s/register?key=%s",
+			h.cfg.ServerURL, mKey.HexString())
+	}
```
kradalby, line 212:
```suggestion
	if h.cfg.OIDCIssuer != "" {
		resp.AuthURL = fmt.Sprintf("%s/oidc/register/%s", strings.TrimSuffix(h.cfg.ServerURL, "/"), mKey.HexString())
	} else {
		resp.AuthURL = fmt.Sprintf("%s/register?key=%s",
			strings.TrimSuffix(h.cfg.ServerURL, "/"), mKey.HexString())
	}
```

oidc.go
```diff
@@ -0,0 +1,196 @@
...
+			log.Error().Msgf("Could not retrieve OIDC Config: %s", err.Error())
+			c.String(http.StatusInternalServerError, "Could not retrieve OIDC Config")
+			return
+		}
+
+		oauth2Config = &oauth2.Config{
+			ClientID:     h.cfg.OIDCClientID,
+			ClientSecret: h.cfg.OIDCClientSecret,
+			Endpoint:     oidcProvider.Endpoint(),
+			RedirectURL:  fmt.Sprintf("%s/oidc/callback", h.cfg.ServerURL),
```
kradalby, line 56:
```suggestion
			RedirectURL:  fmt.Sprintf("%s/oidc/callback", strings.TrimSuffix(h.cfg.ServerURL, "/")),
```

oidc.go
```diff
@@ -0,0 +1,196 @@
...
+	c.Redirect(http.StatusFound, authUrl)
+}
+
+// OIDCCallback handles the callback from the OIDC endpoint
+// Retrieves the mkey from the state cache and adds the machine to the users email namespace
+// TODO: A confirmation page for new machines should be added to avoid phishing vulnerabilities
+// TODO: Add groups information from OIDC tokens into machine HostInfo
+// Listens in /oidc/callback
+func (h *Headscale) OIDCCallback(c *gin.Context) {
+
```
kradalby, line 93:
```suggestion
```

oidc.go
```diff
@@ -0,0 +1,196 @@
...
+type IDTokenClaims struct {
+	Name     string   `json:"name,omitempty"`
+	Groups   []string `json:"groups,omitempty"`
+	Email    string   `json:"email"`
+	Username string   `json:"preferred_username,omitempty"`
+}
+
+var oidcProvider *oidc.Provider
+var oauth2Config *oauth2.Config
+var stateCache *cache.Cache
```
kradalby, line 28: Could you move these three from global into the `Headscale` object?

oidc.go
```diff
@@ -0,0 +1,196 @@
...
+
+		oauth2Config = &oauth2.Config{
+			ClientID:     h.cfg.OIDCClientID,
+			ClientSecret: h.cfg.OIDCClientSecret,
+			Endpoint:     oidcProvider.Endpoint(),
+			RedirectURL:  fmt.Sprintf("%s/oidc/callback", h.cfg.ServerURL),
+			Scopes:       []string{oidc.ScopeOpenID, "profile", "email"},
+		}
+
+	}
```
kradalby, line 60: This section looks like it can have the potential for a race condition, as per comment about moving it from global to the `Headscale` struct. Can this be instantiated once at startup so we are sure we dont end up there?

oidc.go
kradalby: And maybe the stateCache can be named oidcMachineKeyCache or something explicit.

kradalby:
Could you see if you can resolve the conflicts so I can have that as part of the next review?

kradalby (approved):

PR #128, opened 2021-09-27
Added version checker on Headscale startup

juanfont:
This small PR just adds an updated version checker on startup. 

It uses https://github.com/tcnksm/go-latest to fetch the latest release in GitHub.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 version = $(shell ./scripts/version-at-commit.sh)
 
 build:
-	go build -ldflags "-s -w -X github.com/juanfont/headscale/cmd/headscale/cli.version=$(version)" cmd/headscale/headscale.go
+	go build -ldflags "-s -w -X github.com/juanfont/headscale/cmd/headscale/cli.Version=$(version)" cmd/headscale/headscale.go
 
 dev: lint test build
 
--- a/cmd/headscale/cli/version.go
+++ b/cmd/headscale/cli/version.go
@@ -2,11 +2,12 @@ package cli
 
 import (
 	"fmt"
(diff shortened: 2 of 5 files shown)


kradalby (approved):

qbit:
Would it be possible to add a config option to disable this? At least in OpenBSD ports land it's typical that these kind of checks are disabled so as not to cause confusion for people that installed it via `pkg_add`.

The alternative would be that I have to carry patches in the OpenBSD port to disable the check.

kradalby (changes requested):
Actually @qbit has a good point.

Also having something call the internet is not optimal without being able to disable it.

How does that library handle lack of internet?

juanfont:
@qbit @kradalby indeed, that's hell of a point. Would make sense to limit the checks to `runtime.GOOS == "linux"` and  `runtime.GOOS == "darwin"` (the OS that we have binaries for) ?

qbit:
That makes sense to me, I do still think it would be good to have an option to completely disable it for people that don't want to tell github what software they are running.

kradalby:
> @qbit @kradalby indeed, that's hell of a point. Would make sense to limit the checks to `runtime.GOOS == "linux"` and  `runtime.GOOS == "darwin"` (the OS that we have binaries for) ?

No OS checks, those operating systems also have package managers.

It should be possible to turn off for all package maintainers

juanfont:
> This small PR 

Famous last words 😂😂

Indeed, those OS might end up having Headscale packages. We leave it then as a viper config setting (via the `config.json` or ENV) ?

qbit:
LGTM!

kradalby (approved):

PR #129, opened 2021-09-27
Add support for MagicDNS

juanfont:
This PR adds support for MagicDNS. 

The relevant part of the config.json looks like this:

```json
  ...
    "dns_config": {
        "nameservers": [
            "1.1.1.1"
        ],
        "magic_dns": true
    }
```

I plan to add docs + integration tests, but I am having issues testing this in my computers. I get `Temporary failure in name resolution` unless I hardcode `100.100.100.100` in `/etc/resolv.conf`. In that case it actually works.

Looks like Tailscale does not like my systemd-resolve/NetworkManager setup. 

@kradalby can you have a look?

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -28,7 +28,7 @@ import (
 
 // From the netmask we can find out the wildcard bits (the bits that are not set in the netmask).
 // This allows us to then calculate the subnets included in the subsequent class block and generate the entries.
-func generateMagicDNSRootDomains(ipPrefix netaddr.IPPrefix, baseDomain string) (*[]dnsname.FQDN, error) {
+func generateMagicDNSRootDomains(ipPrefix netaddr.IPPrefix, baseDomain string) ([]dnsname.FQDN, error) {
 	base, err := dnsname.ToFQDN(baseDomain)
 	if err != nil {
 		return nil, err
@@ -69,5 +69,5 @@ func generateMagicDNSRootDomains(ipPrefix netaddr.IPPrefix, baseDomain string) (
 		}
 		fqdns = append(fqdns, fqdn)
 	}
-	return &fqdns, nil
+	return fqdns, nil
 }
--- a/app.go
+++ b/app.go
@@ -114,7 +114,7 @@ func NewHeadscale(cfg Config) (*Headscale, error) {
 			return nil, err
 		}
 		h.cfg.DNSConfig.Routes = make(map[string][]dnstype.Resolver)
-		for _, d := range *magicDNSDomains {
+		for _, d := range magicDNSDomains {
 			h.cfg.DNSConfig.Routes[d.WithoutTrailingDot()] = nil
 		}
 	}
--- a/dns_test.go
+++ b/dns_test.go
@@ -11,7 +11,7 @@ func (s *Suite) TestMagicDNSRootDomains100(c *check.C) {
 	c.Assert(err, check.IsNil)
 
 	found := false
-	for _, domain := range *domains {
+	for _, domain := range domains {
 		if domain == "64.100.in-addr.arpa." {
 			found = true
 			break
@@ -20,7 +20,7 @@ func (s *Suite) TestMagicDNSRootDomains100(c *check.C) {
 	c.Assert(found, check.Equals, true)
 
 	found = false
-	for _, domain := range *domains {
+	for _, domain := range domains {
 		if domain == "100.100.in-addr.arpa." {
 			found = true
 			break
@@ -29,7 +29,7 @@ func (s *Suite) TestMagicDNSRootDomains100(c *check.C) {
 	c.Assert(found, check.Equals, true)
 
 	found = false
-	for _, domain := range *domains {
+	for _, domain := range domains {
 		if domain == "127.100.in-addr.arpa." {
 			found = true
 			break
@@ -44,7 +44,7 @@ func (s *Suite) TestMagicDNSRootDomains172(c *check.C) {
 	c.Assert(err, check.IsNil)
 
 	found := false
-	for _, domain := range *domains {
+	for _, domain := range domains {
 		if domain == "0.16.172.in-addr.arpa." {
 			found = true
 			break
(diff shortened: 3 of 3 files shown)


cmd/headscale/cli/utils.go
```diff
@@ -108,8 +108,10 @@ func GetDNSConfig() (*tailcfg.DNSConfig, string) {
 			dnsConfig.Domains = viper.GetStringSlice("dns_config.domains")
 		}
 
-		if viper.IsSet("dns_config.magic_dns") {
-			dnsConfig.Proxied = viper.GetBool("dns_config.magic_dns")
+		if len(dnsConfig.Nameservers) > 0 {
```
kradalby, line 111: This should probably emit an error rather than fail silently?

api.go
```diff
@@ -241,18 +241,28 @@ func (h *Headscale) getMapResponse(mKey wgkey.Key, req tailcfg.MapRequest, m Mac
...
+		if err != nil {
+			return nil, err
+		}
+		dnsConfig = h.cfg.DNSConfig.Clone()
+		for _, ns := range *namespaces {
+			dnsConfig.Domains = append(dnsConfig.Domains, fmt.Sprintf("%s.%s", ns.Name, h.cfg.BaseDomain))
+		}
+	} else {
+		dnsConfig = h.cfg.DNSConfig
+	}
```
kradalby, line 258: Can this snippets be made into a function in preparation for fixing that todo?

Sounds like it could belong in `dns.go`

cmd/headscale/cli/utils.go
```diff
@@ -108,10 +108,23 @@ func GetDNSConfig() *tailcfg.DNSConfig {
...
+			if viper.IsSet("dns_config.magic_dns") {
+				dnsConfig.Proxied = viper.GetBool("dns_config.magic_dns")
+			}
+		}
+
+		var baseDomain string
+		if viper.IsSet("dns_config.base_domain") {
+			baseDomain = viper.GetString("dns_config.base_domain")
+		} else {
+			baseDomain = "headscale.net" // does not really matter when MagicDNS is not enabled
```
kradalby, line 121: Since we dont own (I think?) this domain, can we swap it out for headscale.font.eu or something? I feel like we should control the placeholder.

dns.go
```diff
@@ -0,0 +1,30 @@
...
+	"tailscale.com/util/dnsname"
+)
+
+func (h *Headscale) generateMagicDNSRootDomains() (*[]dnsname.FQDN, error) {
+	base, err := dnsname.ToFQDN(h.cfg.BaseDomain)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(juanfont): we are not handing over IPv6 addresses yet
```
kradalby, line 15:
```suggestion
	// TODO(juanfont): we are not handing out IPv6 addresses yet
```

machine.go
```diff
@@ -52,7 +52,7 @@ func (m Machine) isAlreadyRegistered() bool {
 
 // toNode converts a Machine into a Tailscale Node. includeRoutes is false for shared nodes
 // as per the expected behaviour in the official SaaS
-func (m Machine) toNode(includeRoutes bool) (*tailcfg.Node, error) {
+func (h *Headscale) toNode(m Machine, includeRoutes bool) (*tailcfg.Node, error) {
```
kradalby, line 55: We really need to stop passing the whole application into a function (or make it the parent) just because we need one field.

In this case you have access to the parts of `Machine` you need on the outside, so you can compute the hostname and do:
```suggestion
func (m *Machine) toNode(hostname string, includeRoutes bool) (*tailcfg.Node, error) {
```

or, you can pass just the dns configuration and the base name:

```suggestion
func (m *Machine) toNode(baseDomain string, dnsConfig struct..., includeRoutes bool) (*tailcfg.Node, error) {
```

If we continue with this, it will be impossible to rename or change code, or mutate objects because they are used and are available all over the place.

machine.go
```diff
@@ -147,10 +147,17 @@ func (m Machine) toNode(includeRoutes bool) (*tailcfg.Node, error) {
 		keyExpiry = time.Time{}
 	}
 
+	var hostname string
+	if h.cfg.DNSConfig != nil && h.cfg.DNSConfig.Proxied { // MagicDNS
+		hostname = fmt.Sprintf("%s.%s.%s", m.Name, m.Namespace.Name, h.cfg.BaseDomain)
+	} else {
+		hostname = m.Name
+	}
+
```
kradalby, line 156: As mentioned above, this could be computed outside of this function, possible it can also be its own function.

machine.go
kradalby: Resolving this comment will make merging with master (slightly) easier.

cmd/headscale/cli/utils.go
juanfont: I do own that domain !! haha

machine.go
juanfont: You are so utterly right in this one...

api.go
juanfont: I have reviewed the code, and I reckon we just should send the search domain of the current namespace. Shared nodes need to use the FQDN - as per tailscale.com SaaS. Fixed in the latest commit.

machine.go
juanfont: Does it make sense to make a function out of 5 lines? I rather leave it for the time being there.

dns.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(juanfont): we are not handing out IPv6 addresses yet
+	// and in fact this is Tailscale.com's range (note the fd7a:115c:a1e0: range in the fc00::/7 network)
+	ipv6base := dnsname.FQDN("0.e.1.a.c.5.1.1.a.7.d.f.ip6.arpa.")
+	fqdns := []dnsname.FQDN{base, ipv6base}
+
+	// TODO(juanfont): This only works for the 100.64.0.0/10 range.
```
kradalby, line 20: This needs a warning or error if the sets magic DNS to true and use another range

PR #130, opened 2021-10-01
Allow preauthkeys to specify a subnet for machine to join

jccit:
This PR updates the preauthkey create command with a new subnet argument. If a machine joins the network using this authkey it will be assigned the next IP in the subnet provided.

I've found this quite useful for organising IP addresses of nodes, hopefully someone else will too

Update 2021-10-08: There is also an IP option which can be used for non-reusable keys. This accepts an IP address, converts it into CIDR format and passes it into the create key function. This allows you to pre-assign an IP address for a single node


joy4eg:
I would also extend this command (or the whole topic) to be able to specify the exact IP address to use :)

jccit:
Hey sorry about the delay on this. I've pulled in the new changes and improved some of the error handling. Let me know if it needs improving

diseq:
just to catch your thoughts: this might be useful for advertised tags as well.
with an acl policy attached, one might not want to have the client decide which tags it uses.

jccit:
Closing this for now. This PR was developed to support my employer's headscale deployment. We've since switched to a paid tailscale account and sadly I don't have much time to work on this

PR #131, opened 2021-10-01
fix some typos in README

awoimbee:
Small PR, simply fixing some issues I stumbled upon while testing headscale.
Here is the extent of my small test, if that's any use: https://gist.github.com/awoimbee/0602089974d75bfb2852068ae772fcc7


kradalby (approved):

juanfont:
Thanks!

Issue #97, opened 2021-08-25
Update channel loop bug


busimus:
I noticed that headscale v0.7.0 enters a loop sometimes – it constantly receives requests for update and checks that the client is up to date. It snaps out of it a minute later. 
I'm attaching a log (with slight modifications) of the entire process from program start to the loop ending. And bisecting the repo led me to find the offending commit: 88d7ac04bf7a78f378cd3015c1b1bb083ba54cb3

[hs_88d7ac04b_bug.log](https://github.com/juanfont/headscale/files/7046105/hs_88d7ac04b_bug.log)

jsiebens:
for the record, I've seen the same behaviour, causing high CPU usage during the loop

kradalby:
I think I have noticed this as well, I will take a dive into it.

Based on the PR from that commit, it has ended up being a trade off of ensuring clients are up to date vs not getting all updates. For the time being I would say its better, but I will dive into it when I have some time.

I suspect its a timing and channel/lock issue.

kradalby:
@busimus @jsiebens  Hi, I am sure there is something, and that it is causing high cpu load, but I am struggling to see the loop in the example log provided.

The updates, and the "check if update is necessary" seem to be working as intended. I will agree that the update is somewhat aggressive, but I have not managed to spot the loop.

I might have read the wrong parts of the log, since it is quite comprehensive, would you manage to give me some pointers to how you understand it?

Meanwhile, I'll have a think if it can be made a bit less aggressive.

busimus:
I truncated most of the loop and replaced it with `...`, but I left a few iterations of the repeating five lines before and after that (e.g. lines from 220 to 224).

kradalby:
Thats the stupidest I have been in a long time, I hit the wrong key and did not scroll far enough down 🤦‍♀️

kradalby:
I think this is resolved in #142, could any of you take it for a spin?

PR #136, opened 2021-10-02
Code, pointer, variable cleanups. And metrics!

kradalby:
This PR was suppose to be cleaning up some database calls that were all over the code, but it ended up being a chase for the cause of #97.

It contains a couple of things:

- A lot of fixed type, pointer and variable inconsistencies 
- Initial work on adding metrics (available at `/metrics`) for chasing #97 
- More unit tests
- Improving a bit of the setup of integration tests

I have narrowed down what is causing the occasional storm of update requests, but not sure yet, so will continue to dig.

However this PR is of a size that it should be handled before it become even bigger (worse).


juanfont (approved):

integration_test/etc/config.json
```diff
@@ -7,5 +7,5 @@
   "db_type": "sqlite3",
   "db_path": "/tmp/integration_test_db.sqlite3",
   "acl_policy_path": "",
-  "log_level": "debug"
+  "log_level": "trace"
```
juanfont, line 10: And phasers to stun!

PR #137, opened 2021-10-03
Fix goreleaser version

cure:
The goreleaser builds were not populating the headscale version (accessible via the `version` subcommand on the cli) properly, because the flag wasn't set correctly. This commit fixes that.


cure:
err, sorry I didn't pull to the latest first, I will make a new PR

PR #139, opened 2021-10-03
Make sure that goreleaser uses the appropriate version string when

cure:
building the headscale executable.

The goreleaser builds we re not populating the headscale version (accessible via `headscale version`) properly, because the flag wasn't set correctly. This commit fixes that.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -19,7 +19,7 @@ builds:
     flags:
       - -mod=readonly
     ldflags:
-      - -s -w -X main.version={{.Version}}
+      - -s -w -X github.com/juanfont/headscale/cmd/headscale/cli.Version=v{{.Version}}
   - id: linux-armhf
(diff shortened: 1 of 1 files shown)


qbit:
I don't know much about `goreleaser`, but would it make sense to have it call `scripts/version-at-commit.sh`? That way we only have to set the ldflags and what not once?

cure:
@qbit The ldflags consume the `{{.Version}}` variable, which is calculated by goreleaser from the git tag of the commit (afaik scripts/verstion-at-commit.sh is not involved here, only when building from source with `make`), so that it can be injected as the value of the appropriate variable at build time.

I agree it would be nice if we could have just one set of ldflags defined, but it seems that goreleaser wants it specified for each build target.

kradalby (approved):

juanfont:
Thanks!! :)

PR #141, opened 2021-10-04
Document client OS support in a table

ptman:
Would this be a good documentation improvement as discussed in #58 ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Headscale implements this coordination server.
 | --- | --- |
 | Linux | Yes |
 | OpenBSD | Yes |
-| macOS | Yes |
+| macOS | Yes (see `/apple` on your headscale for more information) |
 | Windows | Yes |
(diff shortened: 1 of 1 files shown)


qbit:
Needs a line for OpenBSD! :D

ptman:
I just looked at https://tailscale.com/download for a list

README.md
```diff
@@ -30,6 +30,16 @@ Headscale implements this coordination server.
 - [x] Share nodes between ~~users~~ namespaces
 - [ ] MagicDNS / Smart DNS
 
+## Client OS support
+
+| OS | Supports headscale |
+| --- | --- |
+| Linux | Yes |
+| macOS | Yes |
```
kradalby, line 38:
```suggestion
| macOS | Yes (see `/apple` on your headscale for more information) |
```

kradalby:
I think this looks like a good start :), I added some comments to point users in the right direction.

Without having tested, I assume Windows users can override the control URL via som regedit/HKEY magic, would be great if someone who figure that out can add a note about that :)

kradalby (approved):

Issue #181, opened 2021-10-19
Clean up and convert example configuration to YAML


kradalby:
JSON its quite hard to read and does not allow comments, to make it easier for users to check the example and to reduce the burden to update I suggest that we:

- Convert the configuration examples to YAML
- Remove the specific postgres/sqlite file, make postgres default, and have the sqlite entry as a comment.
- Add comments to explain each section/configuration option.

PR #184, opened 2021-10-19
Move documentation away from README and use YAML everywhere

juanfont:
This PR moves the documentation we had in README.md into separate files in `docs/`.

It also converts the examples into YAML, as per #181.


kradalby:
Will there be a followup pr converting the SQLite and Postgres example to YAML as well?

PR #185, opened 2021-10-20
fix apple configuration profile links

derelm:
when headscale is mounted in a subpath of a domain these links pointed to potentially invalid urls.
another potential fix would be to use relative links (dropping `{{.Url}}/` in all urls).


kradalby:
@derelm in your case you would have a URL like:

`https://example.no/headscale` ?

So you want it to be `https://example.no/headscale/apple`

Issue #148, opened 2021-10-11
Nodes IP missing in tailscale status


mhhakim:
Hi,

Here is my config.json
<pre>
{
    "server_url": "https://sub.domain.tld",
    "listen_addr": "0.0.0.0:443",
    "ip_prefix": "10.0.1.0/24",
    "private_key_path": "/etc/headscale/private.key",
    "derp_map_path": "/etc/headscale/derp.yaml",
    "ephemeral_node_inactivity_timeout": "30m",
    "db_type": "sqlite3",
    "db_path": "/etc/headscale/db.sqlite",
    "tls_cert_path": "/etc/headscale/crt.pem",
    "tls_key_path": "/etc/headscale/key.pem",
    "acl_policy_path": "",
    "dns_config": {
        "nameservers": ["1.1.1.1", "8.8.8.8"],
        "domains": [],
        "magic_dns": true,
        "base_domain": "local.net"
    }
}
</pre>

Created namescpace. Joined 3 nodes successfully. They can reach each others. The issue is, when I run **tailscale status**, it doesn't show headscale IP on node list. I get the following output.

<pre>
hakim@node-2:~$ sudo tailscale status
10.0.1.2        node-2                main         linux   -
                node-1                main         linux   active; direct PUBLICIP:41641, tx 11524412 rx 288104
                node-3                main         linux   active; direct PUBLICIP:41641, tx 3092 rx 2556
</pre>

I can ping using node name like **ping node-1** because of Magic DNS but can't use tailscale ping node-1 or Taildrop using node name. I have to use node's tailscale IP address. Headscale log seems fine. I am running v0.10.0

mhhakim:
This is my first time setup.

Here's the outputs.

#### ip addr
<pre>
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc mq state UP group default qlen 1000
    link/ether 02:00:17:01:b6:92 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
[...]

PR #208, opened 2021-10-31
Support for IPv6 prefixes in namespaces

enoperm:
I'm sending an MR to initiate a discussion about this initial implementation.

I have found that specifying an IPv6 prefix for `ip_prefix` caused the Headscale server to crash, because `getAvailableIP` assumed an IPv4 address by calling `As4()`.

While I was at it, I also tidied up address generation a bit, because the comment within was inaccurate (a network/broadcast address is one where the host parts of the address are all zero/one bits, not ones that end with eights consecutive zero/one bits), and if I interpret the `netaddr` API reference correctly, `IsZero()` and `IsLoopback()` should never return true for the same address, so I assume the use of `&&` probably had been a typo here.

I also found that `machine.go` also assumed an IPv4 representation and sent `/32` routes to nodes, which `tailscaled` refused to use, even though `tailscale ping` managed to resolve the correct destination node.

These changes were enough to ICMPv6 ping working both against namespace addresses, as well as advertised IPv6 routes. As far as I can see, the changes did not break any of the established tests that use IPv4, but I have not yet added any IPv6 specific test coverage - If I read the code correctly, there is a single unit under test preconfigured with an IPv4 prefix, and I'm not sure about the optimal way to handle the situation.

I have also separately tested with the default IPv4 prefix as well and things seem to still work that way.
I'm not sure why yet, but I was only able to access IPv4 advertised routes when I also used an IPv4 prefix for the namespace, and only able to access IPv6 advertised routes with an IPv6 prefix configured for the namespace. Accessing IPv4 advertised routes from an IPv6 prefix or the other way around does not seem to work, and I have yet to perceive any error messages anywhere, so far I can only observe the lack of packets.


enoperm:
I have had a little bit of time to investigate this branch again, and I have found that I could get IPv6 routes with IPv4-first tailnets or the other way aroudn working. The issue was that none of the hosts had any IPv6 *source addresses* that was included in their `AllowedIPs`. By advertising a `/128` source address from the source node and manually configuring it on the `tailscale0` interface, I was able to get ICMP between a tailscale node and an advertised IPv6 address working with the `ip_prefix` still set to the default IPv4 subnet. As long as addresses are assigned both an IPv4 and an IPv6 address, fully automatic dual-stack tailnets should be possible.

While assigning addresses from multiple pools would be a simple solution, it does raise the question as to what to do if any of the pools are out of addresses, should a node join fail completely, or should it be assigned addresses from a subset of the available pools? Is it possible to push custom messages to tailscale clients in response to a join request?

enoperm:
As of 09c755137e4e296653d54087cee8676cc6ce2473, by updating my config to include a list of prefixes under the config key `ip_prefixes`, instead of a single `ip_prefix`, I have been able to assign both an IPv4 and an IPv6 address to multiple nodes fully automatically, and the nodes could communicate with each other by both addresses. For some reason, only the tailscale IPv4 ranges show up for peers on the output of `tailscale status --json`, but I think this is probably related to #148.

kradalby:
Apologise for the delay, I'm very for this pr, but I had (and have) some refactors to get out of the way before I can take a proper look.

In the mean time, would you be able to add some integration test cases showing of ipv6?

enoperm:
I was thinking of adding unit tests but I don't mind playing around with integration tests either (though admittedly, I haven't run those yet).

enoperm:
Though I guess I have a few hundred commits worth of rebasing to do first. Resolving the merge conflicts after adding even more modifications sounds harder than doing them before. :)

kradalby:
I saw some unit tests and that's good, but we need the integration ones to ensure we are and remain compatible with tailscale. 

So we need both :)

And having it automatic to verify that multiple IPs and ipv6 works as intended is important.

Since this will also change the database schema, we need to test it a bit extra, potentially keep the old field for compatibility.

enoperm:
Yeah, for now I was only focused on working the simplest possible way, so schema compatibility was not even accounted for.  From what I could see, "schema upgrades" happened by GORM automatically creating a new column, and being unable to drop the old one due to sqlite limitations. There's also a bunch of code I haven't tested yet at all, such as the IPv6 magic DNS stuff. It is also not nice of me to squeeze multiple values into a single column they way I currently do.

enoperm:
Oh, I have noticed something else now that I joined on android as well: For the newly joined phone, the displayed address is the IPv6 one, for the others, it is the IPv4 one (the latter is as expected from the output of `tailscale status --json` on my laptop). However, clicking on the name of another note in the android app copies the IPv6 address instead of the displayed IPv4 ones. This is not a particularly useful observation, but thought I would mention it while this tab is still open.

enoperm:
It seems to me that it is possible to retrieve the IPv6 address of peer nodes by running `tailscale ping`. Whether it returns the IPv6 address because I set that first in my config, or the client has some preference for IPv6, I can not say. The same trick might be a workaround for IPv4 prefixes outside of `100.64.0.0/10` in #148.

enoperm:
Okay, from what I can see, taildrop via curl is *not* going to work in integration tests as is in the IPv6 case. Regardless of whether I try to connect to the IPv4 or the IPv6 address on the peer API port, the receiving node seems to forward the connection to `127.0.0.1`. This is fine in the IPv4 case, but apparently (I haven't investigated it thoroughly so far), when IPv6 addresses are involved in the tunnel, `tailscaled` binds that port on IPv6 **only**, which basically breaks taildrop entirely. I am not sure if it is a side effect of running `tailscaled` in docker containers without *any* IPv6 addresses (including loopback). I can override the relevant knob (`net.ipv6.conf.all.disable_ipv6=0`) on the docker command line to get at least a loopback address without changing the docker daemon config, but from what I gather [dockertest](https://godocs.io/github.com/ory/dockertest/v3#RunOptions) does not currently expose the `sysctls` parameter.

enoperm:
Okay, having an IPv6 loopback address is not enough in itself to make `tailscaled` use it as a destination. I still do not see *why* the node winds up binding to an IPv6 only port yet. Still, with a tun device present this whole problem shouldn't exist.

enoperm:
I'm not familiar with github actions, is it possible to provide a `/dev/net/tun` to the containers, and set the equivalent of `--cap-add=NET_ADMIN` on the command line to let `tailscaled` create a new interface?

enoperm:
Okay, as of now I have a branch (https://github.com/enoperm/headscale/tree/ipv6-f) where the integration tests all run fine even on IPv6, but it does rely on the containers being able to instantiate tun devices.
Pros:
* I imagine this to be a more common setup, so it might make the test environment more "realistic".
* Avoids the problem with the peerapi being unaccesible in userspace-networking mode on dualstack(/ipv6-only? haven't tested that) configurations.  (possibly an upstream `tailscaled` bug?)
* Allows us to use `tailscale file cp`, among other things.
Cons:
* I am not sure how the integration test is set up with the CI/CD environment (if at all?), and I do not know whether this could cause any problems.
* Some tests seemed to log source/destination IP addresses, which was only made possible by having a single overlay address for each node. An example would be the magicdns test that uses `tailscale ping` - it does print an address, but it does not allow us to choose which address we use when we let it resolve the target for itself. A possible workaround might be to query `A` or `AAAA` records specifically from `100.100.100.100`, and eliminate the tailscale commands from the test case entirely.

I'll see if I can tidy up the branch into clean little patches later on and update this PR. The changes so far can be categorized as follows:
* IP handling cleanup (can now process IPv6 prefixes, also more efficient for IPv4 prefixes as a side effect).
* MagicDNS IPv6 support for arbitrary prefixes (not explicitly tested, yet).
* Support for multiple IP prefixes per namespace (this one probably deserves a cleaner data model than what I have used so far).
* Some stability fixes in `PollNetMapStream` to avoid crashing the control server on machine joins.

enoperm:
I have cleaned up the branch used in this PR. I have added tests that query `A` records through the DNS interface of `tailscaled` rather than testing MagicDNS with `tailscale ping`. It *should* be trivial to query `AAAA` ones as well, but that did not work on my tailnet when I queried manually. This may be a side effect of using `100.64.0.0/10` as my IPv4 prefix, while using a custom ULA as my IPv6 one. I'll have to check later on whether using Tailscale's own IPv6 prefix makes it work. If it were to work, then I think it would be safe to assume there is some sort of filtering in `tailscaled` and the problem does not lie with this version of Headscale.

enoperm:
Okay, [apparently](https://github.com/tailscale/tailscale/issues/1152), `AAAA` records will only be served by MagicDNS if the server has *nothing* but IPv6 addresses. There seems to be some interest in adding them, but for now (and probably for older versions currently part of the integration test suite), it is not going to happen. I'll go look into which version the `tailscale ip` command, and its `-6` appeared in, and if it seems viable, I'll refactor the tests to use that instead of DNS queries.

enoperm:
Refactored the MagicDNS integration test and split the contents of the `gen/` dir into a separate commit. Assuming I did not forget about anything, PR should be ready for a review.

kradalby:
Hi @enoperm This looks like good thorough work and I will try to review it soon, but currently in need of a little break. Hopefully should be able to get around to it the week after tomorrow.

machine.go
```diff
@@ -31,7 +33,7 @@ type Machine struct {
 	MachineKey  string `gorm:"type:varchar(64);unique_index"`
 	NodeKey     string
 	DiscoKey    string
-	IPAddress   string
+	IPAddresses MachineAddresses
```
juanfont, line 36: This is a breaking change, needs to be documented in the changelog.

cmd/headscale/cli/utils.go
```diff
@@ -221,10 +221,20 @@ func getHeadscaleConfig() headscale.Config {
 	dnsConfig, baseDomain := GetDNSConfig()
 	derpConfig := GetDERPConfig()
 
+	configuredPrefixes := viper.GetStringSlice("ip_prefixes")
```
juanfont, line 224: This needs to be documented, and added in the `config-example.yml`

dns.go
```diff
@@ -34,14 +34,25 @@ const (
...
+func generateMagicDNSRootDomains(ipPrefixes []netaddr.IPPrefix) []dnsname.FQDN {
+	fqdns := make([]dnsname.FQDN, 0, len(ipPrefixes))
+	for _, ipPrefix := range ipPrefixes {
+		var generateDnsRoot func(netaddr.IPPrefix) []dnsname.FQDN
+		switch ipPrefix.IP().BitLen() {
+		case 32:
+			generateDnsRoot = generateIPv4DNSRootDomain
 
+		default:
+			panic(fmt.Sprintf("unsupported IP version with address length %d", ipPrefix.IP().BitLen()))
```
juanfont, line 54: IPv6 is not supported in MagicDNS?

utils.go
```diff
@@ -133,16 +133,32 @@ func encode(
...
+			return
+		}
+		ips = append(ips, *ip)
+	}
 
+	return
+}
+
+// TODO: Is this concurrency safe?
+// What would happen if multiple hosts were to register at the same time?
```
juanfont, line 151: Good point. We might need a lock here.

juanfont (commented):
Added some comments.

dns.go
enoperm: It was stubbed in this commit (which only prepares for multiple namespaces), but IPv6 support was added in a follow-up one.

enoperm: I thought it'd make it easier to review the individual changes this way, but I can squash changes if you'd prefer them that way.

utils.go
enoperm: I think it'd be actually better to put a lock around the callers, it is entirely possible that we get a free IP, but end up not using it because of an error somewhere else before persisting the machine in question - for example, by being able to allocate an IPv6 one but running out of IPv4 ones, or receiving a database error (such as "disk full", "connection lost"...).

dns.go
kradalby: Yes, better this way, but maybe mark it with todo or something and tag with you name so we have some pointers to who and when it will be sorted

utils.go
```diff
@@ -141,36 +141,32 @@ func (h *Headscale) getAvailableIP() (*netaddr.IP, error) {
 		return nil, err
 	}
 
+	ipPrefixNetworkAddress, ipPrefixBroadcastAddress := func() (netaddr.IP, netaddr.IP) {
```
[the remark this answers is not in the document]
enoperm: I personally like using anonymous scopes named functions to to hide away temporary variables, so as not to accidentally refer to them later. I *could* see retrieving the network and broadcast addresses of an IP range being useful elsewhere, though, so I guess I'll extract it.

dns.go
enoperm: I mean, it is *already* sorted, the comment is on an intermediate commit, not on the top of this branch. :)

proto/headscale/v1/machine.proto
```diff
@@ -18,7 +18,7 @@ message Machine {
     string machine_key  = 2;
     string node_key     = 3;
     string disco_key    = 4;
-    string ip_address   = 5;
+    repeated string ip_addresses   = 5;
     string name         = 6;
```
enoperm, line 22: I do not have much experience with gRPC, but I *suppose* this might not be the best way to handle this change. Should I have allocated a new field number? I went with this because I was just trying to get it to work in the simplest possible way. Though even if a new number is allocated, the API will still not be compatible between versions. Do clients transmit enough metadata for things to fail without any strange misparsing if I just leave it like this?

utils.go
enoperm: Ah, I forgot to write about the *actual* point of putting the lock around the whole flow. If the lock was here, it'd only cover the part of the control flow where the next free address is fetched - if the potentially allocated address is not saved into the database before another client enters this function, it may find the same address to be free. So the mutex must last at least long enough to cover both the `fetch-next-address` and `save-machine-to-db` steps.

dns.go
enoperm: Are there still concerns about this thread?

machine.go
```diff
@@ -64,6 +66,47 @@ func (machine Machine) isRegistered() bool {
 	return machine.Registered
 }
 
+type MachineAddresses []netaddr.IP
```
kradalby, line 69: This type should be called IPAddresses or something else reusable in case it could be useful other places.

poll.go
```diff
@@ -388,10 +388,14 @@ func (h *Headscale) PollNetMapStream(
 			updateRequestsReceivedOnChannel.WithLabelValues(machine.Name, machine.Namespace.Name).
 				Inc()
 			if h.isOutdated(machine) {
+				var lastUpdate time.Time
```
kradalby, line 391: oh, good catch...

integration_test.go
```diff
@@ -164,16 +164,16 @@ func (s *IntegrationTestSuite) tailscaleContainer(
 		Name:     hostname,
 		Networks: []*dockertest.Network{&s.network},
 		Cmd: []string{
-			"tailscaled",
-			"--tun=userspace-networking",
-			"--socks5-server=localhost:1055",
+			"tailscaled", "--tun=tsdev",
```
kradalby, line 167: ah this is good

oidc.go
```diff
@@ -126,6 +126,7 @@ var oidcCallbackTemplate = template.Must(
 	</html>`),
 )
 
+// TODO: Why is the entire machine registration logic duplicated here?
```
kradalby, line 129: I think there is fair amount of things that could be tried up about it yes, could you create an issue about it so we can try to track it?

poll.go
```diff
@@ -152,14 +154,33 @@ func (h *Headscale) PollNetMapHandler(ctx *gin.Context) {
 		Str("id", ctx.Param("id")).
 		Str("machine", machine.Name).
 		Msg("Loading or creating update channel")
-	updateChan := make(chan struct{})
 
-	pollDataChan := make(chan []byte)
+	// TODO: could probably remove all that duplication once generics land.
```
kradalby, line 158: \o/ generics!

utils.go
```diff
@@ -133,61 +133,74 @@ func encode(
...
-	ips := make([]netaddr.IP, len(addresses))
-	for index, addr := range addresses {
-		if addr != "" {
-			ip, err := netaddr.ParseIP(addr)
-			if err != nil {
-				return nil, fmt.Errorf("failed to parse ip from database: %w", err)
-			}
-
-			ips[index] = ip
+	// FIXME: This really deserves a better data model,
```
kradalby, line 190: Can you create an issue to track this?

kradalby (commented):
Thanks a lot @enoperm, this is stellar work. I have posted some comments with some questions/clarifications, could you have a look at those?

Other than that I am happy to get this in when we get that resolved.

cmd/headscale/cli/utils.go
```diff
@@ -221,10 +221,20 @@ func getHeadscaleConfig() headscale.Config {
...
+	prefixes := make([]netaddr.IPPrefix, 0, len(configuredPrefixes))
```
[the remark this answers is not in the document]
enoperm: So, the effective value of `ip_prefixes` would be `set(ip_prefix ~ ip_prefixes)`? As long as `ip_prefixes` defaults to an empty list, that sounds like a sane, backwards compatible way of handling it, but how does it play along with `ip_prefixes` having a non-zero default value? Wouldn't it mean people who forget to update their configs will unexpectedly see newly joined machines receiving addresses from both their configured pool and the current default IPv4 pool?

Maybe, from a user perspective, it would make more sense to emit a deprecation warning on using `ip_prefix`, and enforce that one or the other *must* be set - but not both at once.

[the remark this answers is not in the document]
enoperm: Alternatively, do not user `viper` to set the default value, perform the merge+dedup work, and add defaults values if the resulting set is empty - I think that would do the trick.

poll.go
enoperm: To be honest, I kind of forgot about this comment. That block of code had quite a bit more duplication when I wrote it, and I'm not in favor of using generics just to drop a mere three lines of code.

utils.go
enoperm: I suppose, though I haven't really thought much about it - I guess the idiomatic thing to do in a relational database would be to introduce a separate table for IP addresses. With the right schema and queries, I *think* it should also help reduce the amount of code that would be needed to be covered by a mutex around address allocation - If we can have allocated, but not yet associated addresses in the table, the `getUsedIPs` step in it can treat those as used, so we can replace `mutex(get-next-address -> save-machine)` with `mutex(allocate-address) -> save-machine || deallocate-unused-address-on-failure`.

machine.go
enoperm: I have only introduced this type to implement the [Scanner](https://godocs.io/database/sql#Scanner) and [Valuer](https://godocs.io/database/sql#NullBool.Value) interfaces, and consider it a bit of a quick and dirty solution to storing multiple addresses. As can be seen in the other comment about the data model, I consider it more of a temporary state until a more optimal solution comes along("let not perfect be the enemy of good enough" ...), so I am not sure I'd like to see it being introduced elsewhere in the code. :)

poll.go
enoperm: It might have been better off in a separate MR though? I think I patched a number of these while working on getting integration tests to not crash - the IPv6 branch was a lot more likely trigger them.

utils.go
enoperm: Say, does it even make sense to open an issue about something, that is not yet even exists in on `main`? I mean, issues are not limited to branches/tags, and as such I *think* they should be about the mainline. Someone reading an issue about this comment on storing multiple ip addresses per machine would, if they do not already know better, may get the impression such a thing already exists on the main branch as is part of a feature, which is not the case.

I feel like might take on more work by saying this, but let's either open the issue once it actually concerns behaviour already part of the application, or do the rework in this branch, in which case there is no issue to open in the first place, just a conversation in this PR.

oidc.go
enoperm: Done, #294

cmd/headscale/cli/utils.go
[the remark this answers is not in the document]
kradalby: I think that makes sense, I think for a clear message, both emitting a deprecated warning in the logs, but also having it in the `config-example.yaml` make sense, I think there is a surprising amount of users dont reading the logs unless they have to.

I wonder if Viper has a way of marking and handling deprecation?

machine.go
kradalby: Fair, I can live with that. I think I am hunted a bit with the fact that we use a relational DB "already" as I would like to just have a state that could be serialised, but thats a bit late now :P. 

Lets called this resolved and work out better solutions in the future.

poll.go
kradalby: Yes, but I am keen to get them in, so we can live with that for now :)

poll.go
kradalby: Yes, this isnt the only place were Generics will help us :P

machine.go
enoperm: Well, you can always serialize to SQL. sqlite3 is a really nice RDBMS too, and can store everything in a single file. Is there a problem with relational DBs you ran into specifically?

utils.go
kradalby: I think the GitHub naming is a problem here, I look at "Issues" at "Tickets" and I just want a work item we can track this against. So as I consider merging this before this FIXME is resolved, I would like to have it tracked so it is less likely to be forgotten.

I want to get this PR in, and run it as a 0.13.0-beta release so we can leverage some of this exciting changes and stability improvements.

cmd/headscale/cli/utils.go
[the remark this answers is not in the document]
enoperm: I have implemented this config merging, though I have not tested it manually, yet.

cmd/headscale/cli/utils.go
enoperm: Example config updated.

machine.go
enoperm: Note of breaking change added to changelog.

utils.go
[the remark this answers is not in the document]
enoperm: Extracted as `GetIPPrefixEndpoints`. I am not sure this is the best possible name for it, but it is the best I could come up with so far.

utils.go
enoperm: Opened #295.

cmd/headscale/cli/utils.go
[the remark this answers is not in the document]
enoperm: Okay, I have tested it, apparently converting the prefix back into a string does not normalize the address by itself, but converting it to a range drops the host bits, so converting *that* back into a prefix, and then a string, allows one to build a canonical representation of them. This way `100.64.0.0/10` and `100.64.0.1/10` maps to the same key, and deduplication happens.

machine.go
kradalby: Nono, dont get me wrong, I really like sqlite, I just wonder if we could have done with a bit more flexibility in the beginning and then designed an appropriate RDBMS schema later. But we will get there.

cmd/headscale/cli/utils.go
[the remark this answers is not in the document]
kradalby: Ok thats good, I think that we can live with this until we actually remove the option and can remove that code again.

kradalby:
I am happy with this PR and propose the following:

We merge a couple of outstanding _fix_ PRs, plus @enoperm other PR (#282) and release 0.12.4 as a fix release.

Then we merge this PR and start releasing 0.13.0-beta.

@juanfont What do you think? I have a look over the PR and see if you agree with this approach.

enoperm:
My other PR was #278, though. :)

cmd/headscale/cli/utils.go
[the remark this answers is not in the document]
enoperm: Uff, I forgot a debug println in the last commit.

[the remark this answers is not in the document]
enoperm: Pushed a fixup.

juanfont:
> I am happy with this PR and propose the following:
> 
> We merge a couple of outstanding _fix_ PRs, plus @enoperm other PR (#278) and release 0.12.4 as a fix release.
> 
> Then we merge this PR and start releasing 0.13.0-beta.
> 
> @juanfont What do you think? I have a look over the PR and see if you agree with this approach.

I fully agree. Let's merge the other PRs, release 0.12.4 and a prerelease of 0.13.0 including this amazing piece of engineering by @enoperm.

kradalby (approved):

PR #209, opened 2021-10-31
Enable prerelease support on goreleaser

juanfont:
So we can have stuff like x.y.z-beta1 tagged as a pre-released in GitHub by goreleaser.


kradalby:
This does not set it to full auto? Like on every merge? We have to actively write beta and then it marks it?

I personally prefer just having the artefact on the build job, then I know exactly what I'm getting.

Maybe we can have goreleaser do that?
We currently only have linux amd64 for that purpose:

https://github.com/juanfont/headscale/actions/runs/1404649843

Issue #219, opened 2021-11-12
Stricter linting


kradalby:
We currently have linting from `golangci-lint` standard package and some `buf` linting for proto.

We should add a configuration and consistent ci for linting that covers:

For go:
- catch bugs
- Best practices 
- Style

For yaml, markdown and more:
- formatting with `prettier`
- yamllint 

For proto:
- buf linting
- Clangformatting

All the linters and formatters should be guarded by CI and be trivial for the user to run via make.

kradalby:
I am to get this in after #212 is merged, if we do this with too much stuff open, we will have a lot of merge conflicts.

PR #223, opened 2021-11-13
Setup lint and format of Go, Proto and yaml/markdown

kradalby:
This PR sets up linting for all our different file types in the CI and adds a format command to make to try to improve consistency.

I have disabled a bunch of linters, some intentionally.

However there is a list of "best practice" linters that I think we should add, but we have a solid backlog to get them passing.

This PR is initial work for #219

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -45,7 +45,6 @@ linters:
     - goconst
     - exhaustive
     - varnamelen
-    - nilnil
 
     # We might want to enable this, but it might be a lot of work
     - cyclop
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -30,7 +30,6 @@ linters:
 
     # We should strive to enable these:
     - wrapcheck
-    - goerr113
     - dupl
     - makezero
 
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -30,7 +30,6 @@ linters:
     - stylecheck
     - wrapcheck
     - paralleltest
-    - noctx
     - nlreturn
     - ifshort
     - gomnd
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,6 @@
 ---
 run:
-  timeout: 5m
+  timeout: 10m
 
 issues:
   skip-dirs:
@@ -23,6 +23,7 @@ linters:
     - exhaustivestruct
     - tagliatelle
     - godox
+    - ireturn
 
     # We should strive to enable these:
     - testpackage
@@ -47,6 +48,8 @@ linters:
     - goconst
     - unconvert
     - exhaustive
+    - varnamelen
+    - nilnil
 
     # We might want to enable this, but it might be a lot of work
     - cyclop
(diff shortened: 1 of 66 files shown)


acls.go
```diff
@@ -108,7 +108,9 @@ func (h *Headscale) generateACLPolicySrcIP(u string) ([]string, error) {
 	return h.expandAlias(u)
 }
 
-func (h *Headscale) generateACLPolicyDestPorts(d string) ([]tailcfg.NetPortRange, error) {
+func (h *Headscale) generateACLPolicyDestPorts(
+	d string,
+) ([]tailcfg.NetPortRange, error) {
 	tokens := strings.Split(d, ":")
 	if len(tokens) < 2 || len(tokens) > 3 {
 		return nil, errorInvalidPortFormat
```
juanfont, line 115: Is this the standard gofmt config?

juanfont: I mean in general, not just this line...

Makefile
```diff
@@ -19,7 +27,12 @@ coverprofile_html:
 	go tool cover -html=coverage.out
 
 lint:
-	golangci-lint run --fix
+	golangci-lint run --fix --timeout 10m
+
+fmt:
+	prettier --check '**/**.{ts,js,md,yaml,yml,sass,css,scss,html}'
+	golines --max-len=88 --base-formatter=gofumpt -w $(GO_SOURCES)
```
juanfont, line 34: Why 88?

juanfont (dismissed):

acls.go
kradalby: No, its gofumpt,which I think is standard in vscode, and then it is run through golines to break long lines.

Makefile
kradalby: It is based on [python black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) (and _[Raymond Hettinger](https://youtu.be/wf-BqAjZb8M?t=260)_), which some might hate.

The idea is to make the code more naturally fit on narrow screens, so you can have multiple files open without bad editor wrapping.

kradalby: I'll adda note in the reader about code style

kradalby: Done

acls.go
kradalby: Added note in readme.

juanfont (dismissed):
All good!

kradalby:
This is ready for now. This should give us a lot of helpers to keep consistent good practice.

kradalby:
PTAL @juanfont, I think we need to have the old lint requirements removed

.golangci.yaml
```diff
@@ -49,3 +46,15 @@ linters:
     - cyclop
     - nestif
     - wsl # might be incompatible with gofumpt
+
+linters-settings:
+  varnamelen:
```
juanfont, line 44: :( this will hurt

juanfont (approved):
Wow. Just, wow.

.golangci.yaml
kradalby: I have fixed The part that hurts

PR #229, opened 2021-11-19
Upgrade Tailscale version to 1.18.x

kradalby:
See individual commits.

There is a new key library, and it forces some refactoring, I might have missed something, but this should have integration tests passing for the new and the previous version when its done.

This PR contains a **breaking change**.

The old private key format is no longer supported, and users will have to delete their private key. Headscale will now automatically generate it on startup if it does not exist.


juanfont (approved):
This is fantastic :)

PR #236, opened 2021-11-24
fix(derp-example): change regionid in node

restanrm:
Using a wrong `regionid` value lead to non working DERP custom server. No checks are performed for this kind of thing making it difficult to find and fix.


kradalby:
My bad.

As for checks, we allow the users to manually add all the upstream Tailscale servers, and not use the automatic fetcher, so having a check for this would just fail valid ones either. Or you would have to implement a allowlist.

kradalby (approved):

restanrm:
For the tests I meant something like a linter that just check redundancy between the id in the map and the regionID. It should be kind of a pain to implement and doesn't provide much. 

Thanks for merging the PR !

PR #237, opened 2021-11-26
Fix default preauthkey expiration

cure:
When creating a preauthkey, the default expiration was passed through as a nil value, instead of the default value (24h). This resulted in the preauthkey being created with expiration key '0001-01-01 00:00:00', which meant the key would not work, because it was already expired.

This commit applies the default expiration time (24h) when a preauthkey is created without a specific expiration. It also updates an integration test to make sure this bug does not reoccur.

### DIFF
--- a/cmd/headscale/cli/preauthkeys.go
+++ b/cmd/headscale/cli/preauthkeys.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	DefaultPreAuthKeyExpiry = 24 * time.Hour
+	DefaultPreAuthKeyExpiry = 1 * time.Hour
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


kradalby:
This one's on me for not commenting better! And also not removing the default text.

It was added intentionally, however that was only to preserve the current behaviour of the cli at the time, which was infinite keys if there was no expiration set. IIRC.

Anyways, I'm very happy for this to go in, and while we are at it, how about 1h or 30m default expiration? 24h seems a bit extreme as a default?

kradalby:
Ah, and as you pointed out, the expiry check in the other end was probably incorrect. Thanks for catching this.

kradalby (dismissed):

cure:
Sure! One hour default sounds good, I've rebased and pushed the corresponding change.

kradalby (approved):

PR #247, opened 2021-12-01
Add private_key_path to example config

negbie:


kradalby (approved):
Apologies, I must have not reverted the example config when I reverted removing it

kradalby:
Thanks :)

PR #248, opened 2021-12-01
Ignoe derp.yaml, don't panic in Serve()

negbie:
If there is no specifc reason to panic I think it's better to just return the error.


negbie:
@kradalby Do you know why the integration-test for the second pr fails? It adds only more logging.

kradalby:
They are a bit flaky, I think they are a bit more resource hungry than actions wants to so they might get in an awkward state.

Reruns usually helps...

negbie:
Makes sense. Thanks for the info!

app.go
```diff
@@ -418,12 +418,12 @@ func (h *Headscale) Serve() error {
 
 	err = h.ensureUnixSocketIsAbsent()
 	if err != nil {
-		panic(err)
+		return err
```
kradalby, line 421: I think this is sensible,

But could you wrap these errors to give the user more context, when we are first at improving :)

So:
```suggestion
		return fmt.Errorf("description: %w", err)
```

app.go
```diff
@@ -418,12 +418,12 @@ func (h *Headscale) Serve() error {
...
 	err = h.ensureUnixSocketIsAbsent()
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("unable to remove old socket file: %w", err)
 	}
 
 	socketListener, err := net.Listen("unix", h.cfg.UnixSocket)
 	if err != nil {
-		panic(err)
+		return err
```
kradalby, line 426:
```suggestion
		return fmt.Errorf("failed to set up gRPC socket: %w", err)
```

app.go
```diff
@@ -441,7 +441,7 @@ func (h *Headscale) Serve() error {
 
 	networkListener, err := net.Listen("tcp", h.cfg.Addr)
 	if err != nil {
-		panic(err)
+		return err
```
kradalby, line 444:
```suggestion
		return fmt.Errorf("failed to bind to TCP address: %w", err)
```

kradalby (approved):

PR #290, opened 2022-01-28
Add generate private-key command

### LINKED ISSUE #286: Add command line option to generate the private key
If a user has some sort of secret management, having to start tailscale to generate the private key file is a bit frustrating, we should add an option to do this:

```bash
headscale generate-privatekey
```

It could either spit out to `stdout` or to a file.

kradalby:
This commit adds a command to generate a private key for headscale.

Mostly useful for systems were you drive the deployment from another
machine and use a secret management system.

Resolves #286


cmd/headscale/cli/generate.go
```diff
@@ -0,0 +1,41 @@
...
+	"tailscale.com/types/key"
+)
+
+func init() {
+	rootCmd.AddCommand(generateCmd)
+	generateCmd.AddCommand(generatePrivateKeyCmd)
+}
+
+var generateCmd = &cobra.Command{
+	Use:   "generate",
```
juanfont, line 16: [disclaimer from non-native speaker]

Would it be better just `create` ?

juanfont (approved):

cmd/headscale/cli/generate.go
kradalby: I feel that create makes something in the database, while this generate is something that don't change the headscale instance or database

juanfont (approved):

Issue #254, opened 2021-12-12
client need certificate?


woday-dev:
**Issue description**
When TLS key and cert is enabled in headscale server, tailscale client cannot connect, as server said client need to provide certificate.
After check with code, seems like bellow code is responsible, is this behavior expected?
Seems make no sense to require tailscale client provide any certificate.
```
app.go
ClientAuth:   tls.RequireAnyClientCert,
```

negbie:
https://www.cloudflare.com/de-de/learning/access-management/what-is-mutual-tls/

woday-dev:
> https://www.cloudflare.com/de-de/learning/access-management/what-is-mutual-tls/

Thanks for explanation,  it's mutual-tls and it's safer, that's not the issue.
Yet seems no option to let tailscale client provide certificate, thus handshake cannot be made.

ficapy:
I have the same problem, using HTTP can be deployed successfully, but using HTTPS, the client can't connect

arch4ngel:
I'm a bit perplexed by this as well.

I understand it's possible to issue client certificates once one has a CA, but I'm unsure how to supply that certificate to Tailscale client's for authentication.

I'm quite new to Tailscale and I've done some Google due diligence but I've come up empty handed.

Tailscale itself has a `cert` subcommand but it looks to be concerned with procuring certs from the Tailscale network.

I'd be grateful for a tip on this.

arch4ngel:
Quite certain I'm not doing it right, but I implemented an Nginx reverse HTTP proxy using [this guidance](https://app.smallstep.com/hello-mtls/doc/client/nginx-proxy) at the perimeter of my network.

It gets the job done, though joining nodes is obviously limited to my home network.

PR #297, opened 2022-01-29
Configurable mtls

arch4ngel:
- [x] read the [CONTRIBUTING guidelines](README.md#user-content-contributing
- [x] added unit tests
- [] added integration tests
- [] updated CHANGELOG.md

# The Problem

Currently, mTLS is set to require *any* certificate from clients when TLS is enabled. Verification of the certificate is not performed by Headscale, and *as far as I can tell*, there isn't a way to configure the Tailscale client to accept TLS certificate for authentication.

# Implemented Solution

This merge makes mTLS configurable by adding a configuration parameter to `config.yaml`. It's disabled via the `disabled` value by default and can be configured to require any certificate (`relaxed`) or require and verify (`enforced`) a client certificate.

# Related Issues

- #254 calls this out as a question but not really a bug.


cmd/headscale/cli/utils.go
```diff
@@ -40,6 +40,7 @@ func LoadConfig(path string) error {
 
 	viper.SetDefault("tls_letsencrypt_cache_dir", "/var/www/.cache")
 	viper.SetDefault("tls_letsencrypt_challenge_type", "HTTP-01")
+	viper.SetDefault("tls_client_auth_mode", "disabled")
```
kradalby, line 43: This default is different than the current setting, how come?

app.go
```diff
@@ -644,12 +645,34 @@ func (h *Headscale) getTLSSettings() (*tls.Config, error) {
...
+			// Client cert is required and verified.
+			clientAuthMode = tls.RequireAndVerifyClientCert
+		} else {
+			return nil, errors.New(
+				"Invalid tls_clientAuthMode provided: " +
+					h.cfg.TLSClientAuthMode)
+		}
+
+		log.Info().Msg(fmt.Sprintf(
+			"Client authentication (mTLS) is \"%s\". See the docs to learn about configuring this setting.",
```
kradalby, line 666: Can you link the docs here?

docs/tls.md
```diff
@@ -29,3 +29,22 @@ headscale can also be configured to expose its web service via TLS. To configure
...
 ```
+
+### Configuring Mutual TLS Authentication (mTLS)
+
+mTLS is a method by which an HTTPS server authenticates clients, e.g. Tailscale,
+using TLS certificates. The capability can be configured by by applying one of
+the following values to the `tls_client_auth_mode` setting in the configuration
+file.
+
+| Value | Behavior |
```
kradalby, line 40: This file needs to be ran through `prettier`

app.go
```diff
@@ -644,12 +645,34 @@ func (h *Headscale) getTLSSettings() (*tls.Config, error) {
 		if !strings.HasPrefix(h.cfg.ServerURL, "https://") {
 			log.Warn().Msg("Listening with TLS but ServerURL does not start with https://")
 		}
+
+		var clientAuthMode tls.ClientAuthType
+		if h.cfg.TLSClientAuthMode == "disabled" {
```
kradalby, line 650: This should be a `switch` statement, and the alternatives should be Constants.

kradalby (changes requested):
Could you add a test to verify that the values work as intended and a failure case to verify failure?

And add the entry to `config-example.yaml` with some docs (and linking to the tls doc)

cmd/headscale/cli/utils.go
arch4ngel: Please correct me if I'm wrong, but I think Tailscale doesn't support client certificates right now.

The current default, [RequireAnyClientCert](https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/crypto/tls/common.go;l=302), prevents connections over TLS unless any cert is provided by the client.

If I'm not overlooking something, then this means that Tailscale can't interact with Headscale when it's configured to use TLS.

app.go
arch4ngel: Happy to make the change, but should we be concerned about link rot?

app.go
arch4ngel: Good call. I'm new to Go, so thank you for this critique. Much less typing in a switch statement.

cmd/headscale/cli/utils.go
kradalby: > If I'm not overlooking something, then this means that Tailscale can't interact with Headscale when it's configured to use TLS.

Hmm, I am not sure what this means? I am running my `headscale` with TLS?

app.go
kradalby: Fair enough, we can skip that.

cmd/headscale/cli/utils.go
arch4ngel: How curious! All of my browsers throw a bad ssl error unless a client certificate is supplied, then it works as expected. 

I'm using the config parameters to point at cert values for letsencrypt files I already have on hand.

I believe issue #254 reflects the same outcome. 

Do we have a misconfiguration somewhere or something?

kradalby: Hmm, happy to make it configurable so users with this problem can at least test them, however, there are many successful users (me included) and I think think there is a reason to change the default for everyone.

My current setup has tls set in nginx, but a week ago, I used the letsencrypt built in to headscale.

So yep, let's get it in, but, we should not change the default.

arch4ngel: Agreed! Sorry for the confusion there.

Should have things buttoned up tonight/early tomorrow.

app.go
arch4ngel: In the latest commit, I moved this logic to a new function to make testing a bit simpler and implemented a switch statement.

docs/tls.md
arch4ngel: It has now been ran through prettier. My bad on the oversight.

cmd/headscale/cli/utils.go
arch4ngel: Alrighty, the latest commit preserves the original default.

app.go
```diff
@@ -88,8 +92,9 @@ type Config struct {
 	TLSLetsEncryptCacheDir      string
 	TLSLetsEncryptChallengeType string
 
-	TLSCertPath string
-	TLSKeyPath  string
+	TLSCertPath       string
+	TLSKeyPath        string
+	TLSClientAuthMode string
```
kradalby, line 97: 1. I would argue that you can just use the `tls.ClientAuthType` type here.

app.go
```diff
@@ -649,18 +654,48 @@ func (h *Headscale) getTLSSettings() (*tls.Config, error) {
 		if !strings.HasPrefix(h.cfg.ServerURL, "https://") {
 			log.Warn().Msg("Listening with TLS but ServerURL does not start with https://")
 		}
+
+		clientAuthMode, err := h.GetClientAuthMode()
+		if err != nil {
+			return nil, err
+		}
```
kradalby, line 661: 2. Then remove this ( or cut it for later )

app.go
```diff
@@ -649,18 +654,48 @@ func (h *Headscale) getTLSSettings() (*tls.Config, error) {
...
+
+		log.Info().Msg(fmt.Sprintf(
+			"Client authentication (mTLS) is \"%s\". See the docs to learn about configuring this setting.",
+			h.cfg.TLSClientAuthMode))
+
 		tlsConfig := &tls.Config{
-			ClientAuth:   tls.RequireAnyClientCert,
+			ClientAuth:   clientAuthMode,
```
kradalby, line 668: 3. Then do this:

```suggestion
			ClientAuth:   h.cfg.TLSClientAuthMode,
```

cmd/headscale/cli/utils.go
```diff
@@ -85,6 +86,12 @@ func LoadConfig(path string) error {
 		!strings.HasPrefix(viper.GetString("server_url"), "https://") {
 		errorText += "Fatal config error: server_url must start with https:// or http://\n"
 	}
+
+	clientAuthMode := viper.GetString("tls_client_auth_mode")
+	if clientAuthMode != "disabled" && clientAuthMode != "relaxed" && clientAuthMode != "enforced" {
+		errorText += "Invalid tls_client_auth_mode supplied. Accepted values: disabled, relaxed, enforced."
+	}
```
kradalby, line 93: 4. Then instead here call the `h.GetClientAuthMode` function and check the error instead of having that `if`.

cmd/headscale/cli/utils.go
```diff
@@ -301,8 +308,9 @@ func getHeadscaleConfig() headscale.Config {
 		),
 		TLSLetsEncryptChallengeType: viper.GetString("tls_letsencrypt_challenge_type"),
 
-		TLSCertPath: absPath(viper.GetString("tls_cert_path")),
-		TLSKeyPath:  absPath(viper.GetString("tls_key_path")),
+		TLSCertPath:       absPath(viper.GetString("tls_cert_path")),
+		TLSKeyPath:        absPath(viper.GetString("tls_key_path")),
+		TLSClientAuthMode: viper.GetString("tls_client_auth_mode"),
```
kradalby, line 313: 5. Then just set it here if there was no error in 4.:
```suggestion
		TLSClientAuthMode: clientAuthMode,
```

config-example.yaml
```diff
@@ -92,6 +92,13 @@ acme_email: ""
 # Domain name to request a TLS certificate for:
 tls_letsencrypt_hostname: ""
 
+# Client (Tailscale/Browser) authentication mode (mTLS)
+# Acceptable values:
+# - disabled: client authentication disabled
+# - relaxed: client certificate is required but not verified
+# - enforced: client certificate is required and verified
+tls_client_auth_mode: disabled
```
kradalby, line 100: This should be relaxed to reflect the default.

```suggestion
tls_client_auth_mode: relaxed
```

kradalby (changes requested):
Looking like its get there, I think this might be a way to simplify the flow a bit and remove some redundant checks.

docs/tls.md
```diff
@@ -29,3 +29,20 @@ headscale can also be configured to expose its web service via TLS. To configure
 tls_cert_path: ""
 tls_key_path: ""
 ```
+
+### Configuring Mutual TLS Authentication (mTLS)
+
+mTLS is a method by which an HTTPS server authenticates clients, e.g. Tailscale,
+using TLS certificates. The capability can be configured by by applying one of
```
stefan-caraiman, line 36: Duplicate wording? `by by applying one of`

kradalby:
@arch4ngel Looks good, last few things is just make the linters happy, and add a line to the changelog under TBD.

Good job :)

arch4ngel:
> @arch4ngel Looks good, last few things is just make the linters happy, and add a line to the changelog under TBD.
> 
> Good job :)

Thanks for your patience on this! Work consumed me for a few weeks.

kradalby:
Need another run of `prettier` and then I'll approve, good work, sorry for all the nits ;)

kradalby (approved):

PR #303, opened 2022-01-30
Migrate ip_address field to ip_addresses

kradalby:
The automatic migration did not pick up this change and it breaks
current setups as it only adds a new field which is empty.

This means that clients who dont request a new IP will not have any IP
at all.

### DIFF
--- a/db.go
+++ b/db.go
@@ -30,18 +30,24 @@ func (h *Headscale) initDB() error {
 	if h.dbType == Postgres {
 		db.Exec("create extension if not exists \"uuid-ossp\";")
 	}
+
+	_ = db.Migrator().RenameColumn(&Machine{}, "ip_address", "ip_addresses")
+
 	err = db.AutoMigrate(&Machine{})
 	if err != nil {
 		return err
 	}
+
 	err = db.AutoMigrate(&KV{})
 	if err != nil {
 		return err
 	}
+
 	err = db.AutoMigrate(&Namespace{})
 	if err != nil {
 		return err
 	}
+
 	err = db.AutoMigrate(&PreAuthKey{})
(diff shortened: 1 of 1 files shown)


kradalby:
@enoperm, could you help me test this? It seem to work for my sqlite setup, and I think this is needed as 0.13.0-beta2.

enoperm:
> @enoperm, could you help me test this? It seem to work for my sqlite setup, and I think this is needed as 0.13.0-beta2.

I too, have a sqlite setup, so if that works for you I expect no differences here. My current in-use database is old enough to have an `ip_address` field, but it is all empty as all current nodes in it have joined after the change to `ip_addresses`, so if anything, this would break *that* setup, I'll have to drop that column before I upgrade to this version if I do not want lose all my addresses. :)

I could try configuring an older 0.11 or 0.12 build with postgres and upgrade that, in case you were looking to confirm behaviour on such a setup.

At any rate, the code change itself looks good, assuming it does not fail when the column does not exists anymore (ie. subsequent startups), but I suppose that's why the return value is ignored.

enoperm:
One little observation, that *may* improve readability further down the road: in go you can use backtics for string literals, so one does not need to escape quotes in strings like SQL statements:
```go
"create extension if not exists \"uuid-ossp\";"
// vs
`create extension if not exists "uuid-ossp";`
```

kradalby:
> At any rate, the code change itself looks good, assuming it does not fail when the column does not exists anymore (ie. subsequent startups), but I suppose that's why the return value is ignored.

Yep, that is the intention, the docs just dont give a direction on what pattern should be used...

juanfont (approved):

PR #311, opened 2022-02-08
Issues with current ACL implementation and solution proposal

restanrm:
This pull request is not really meant to be merged, but to try and explain the issues with the current ACL's and propose an alternative. This was discussed with @kradalby on discord.

I hope that the issue is correctly explained.


docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+# ACLs
+
+A key component of tailscale is the notion of Tailnet. This notion is hidden but the implications that it have on how to use tailscale are not.
+
+For tailscale an [tailnet](https://tailscale.com/kb/1136/tailnet/) is the following:
+
+> For personal users, you are a tailnet of many devices and one person. Each device gets a private Tailscale IP address in the CGNAT range and every device can talk directly to every other device, wherever they are on the internet.
+>
+> For businesses and organizations, a tailnet is many devices and many users. It can be based on your Microsoft Active Directory, your Google Workspace, a GitHub organization, Okta tenancy, or other identity provider namespace. All of the devices and users in your tailnet can be seen by the tailnet administrators in the Tailscale admin console. There you can apply tailnet-wide configuration, such as ACLs that affect visibility of devices inside your tailnet, DNS settings, and more.
+
```
kradalby, line 11: I can see how this "error" was misunderstood and implemented in the beginning yes. Let us make it better.

docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+}
+```
+
+While this works, it requires a lot of manual editing on the configuration and to keep track of all devices IP address.
+
+## Proposition for a next implementation
+
+In order to ease the use of ACL's we need to split the tailnet and users notion.
+
+A solution could be to consider a headscale server (in it's entirety) as a tailnet.
```
kradalby, line 40: I think this sounds like a reasonable scope for `headscale`. It is not meant to be a "enterprise self-hosted" implementation, and limiting it to one tailnet (at least for now) seem like it can simplify things for us quite a bit.

docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+
+For personal users the default behavior could either allow all communications between all namespaces (like tailscale) or dissallow all communications between namespaces (current behavior).
+
+For businesses and organisations, viewing a headscale instance a single tailnet would allow users (namespace) to talk to each other with the ACLs. As described in tailscale's documentation [[1]], a server should be tagged and personnal devices should be tied to a user. Translated in headscale's terms each user can have multiple devices and all those devices should be in the same namespace. The servers should be tagged and used as such.
```
kradalby, line 44: It seems like _allow all_ as a default seem desirable, and then you add ACLs to limit it. That seem to be the thing that would align the most with Tailscales upstream behaviour.

docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+
+This implementation would render useless the sharing feature that is currently implemented since an ACL could do the same.
```
kradalby, line 46: I agree, and removing the code would simplify the code, reduce it by quite a bit.

Having it ACLs control this would also give us "one interface" instead of the current two: sharing + ACL

docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+
+What could be improved would be to peer different headscale installation and allow `sharing`. This would raises issues about compatible network IPs range.
```
kradalby, line 48: This sounds like it _could_ be a longterm goal, but out of scope for 90% of the base usecases.

docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+        {"action":"accept", "users":["boss"], "ports":["boss:*"]},
+        {"action":"accept", "users":["dev1"], "ports":["dev1:*"]},
+        {"action":"accept", "users":["dev2"], "ports":["dev2:*"]},
+        {"action":"accept", "users":["admin1"], "ports":["admin1:*"]},
+        {"action":"accept", "users":["intern1"], "ports":["intern1:*"]},
+    ]
+}
+```
+
+With this implementation, the sharing step is not necessary. Maintenance cost of the ACL file is lower and less tedious (no need to map hostname and IP's into it).
```
kradalby, line 288: I like this, it would also make the OIDC implementation a lot more sensible (one login -> one namespace).

kradalby (commented):
I think this makes sense. I think it has been like this since the usecases of the maintainers has currently been 100% personal usage and we just have not ran into this. But I can see how it is better.

My other comments should probably be read before this, as this is the last section:

However my main concern is the migration path.

I am contemplating two relativeliy different one:

1. Change the logic

By this I mean, implement the new proposed solution, allowing access between all namespaces and fixing the ACL system.
As a part of allowing access, remove the shared node code, table, tests etc.

It would be a breaking change, but we are not 1.0.0, so that "should be fine" :tm:.

2. "rename" namespace to user

This is a "even more" breaking change, while I think it would make sense to clean up the naming to avoid confusion and bring it closer to the upstream lingo. 

Basically this would:
- Implement the new suggestion as User (to reflect what it actually is)
- Remove share node feature
- Remove the namespace (cli, table, code, etc).

This would make the user/group keywords in ACLs make more sense aswell.

Before doing either, we should make sure we "get it right", I can't see any concerns right now, but would like to get some more eyes on it.

@juanfont @madjam002 @ohdearaugustin @cure 

Edit:
I also want to write the ACLs in YAML (as well as HUJSON), but this should be trivial.

madjam002:
I really like this and is basically what I've done with my Headscale instance. Removing sharing and just relying on ACLs to control access seems like a good solution.

One thing to note - I'm pretty sure I read somewhere on the Tailscale docs that peers are only sent to people who have permission to access them via ACLs (I can't find the page now where I read it). Perhaps this is something that Headscale could implement in the future, it doesn't really help security as a connection can't be made if an ACL doesn't allow it, but it just means that `tailscale status` / Windows tray menu would only show nodes that are allowed for that particular user/device via ACLs.

docs/acls.md
ohdearaugustin: I also agree on @kradalby for now this implementation would be sustainable.

docs/acls.md
ohdearaugustin: I will be in the favor for the current behavior, as it implies more security by default. This is probably a typical black/whitelist problem.
Being more align with the Tailscale upstream behaviour is a good point to consider.

docs/acls.md
ohdearaugustin: Absolutely agree on that.

docs/acls.md
ohdearaugustin: Also rather longterm, as probably more complicated to implement and as mentioned by @kradalby not a usecase for most users.

docs/acls.md
ohdearaugustin: No need hostname and IP mapping is great. Generally this feature should be preserved.

docs/acls.md
```diff
@@ -0,0 +1,288 @@
...
+headscale namespaces create intern1
+```
+
+We don't need to create namespaces for the servers because the servers will be
+tagged. When registering the servers we will need to add the flag
+`--advertised-tags=tag:<tag1>,tag:<tag2>`, and the user (namespace) that is
+registering the server should be allowed to do it. Since anyone can add tags to
+a server they can register, the check of the tags is done on headscale server
+and only valid tags are applied. A tag is valid if the namespace that is
+registering it is allowed to do it.
```
ohdearaugustin, line 212: Serverside validation of tag permissions sounds good.

ohdearaugustin:
This adaptation of the ACL would need breaking changes, but I also see no problem with that as the project is still in constant developed and we won't need a backwards compatibility.

Furthermore if we complete rework the ACL maybe we also should add  the ACL configuration to a API. So a more dynamical configuration is possible. As for now we would need to change the configuration file, which will need a restart/reload of the server. 

All in all I think that is a great proposal.

I also would love to write the ACL in Yaml. Not important just nice to have.

docs/acls.md
restanrm: I'm personally more in favor of using ACL's to filter users behavior and all open by default. Users would not have to worry that nodes cannot talk to each other when they start their Headscale instance. If this is a needed feature it's quite easy to implement with the ACLs. The biggest issue is to properly understand Tailscale's ACLs that are quite difficult to grasp in my opinion.

docs/acls.md
restanrm: > I really like this and is basically what I've done with my Headscale instance. Removing sharing and just relying on ACLs to control access seems like a good solution.
> 
> One thing to note - I'm pretty sure I read somewhere on the Tailscale docs that peers are only sent to people who have permission to access them via ACLs (I can't find the page now where I read it). Perhaps this is something that Headscale could implement in the future, it doesn't really help security as a connection can't be made if an ACL doesn't allow it, but it just means that `tailscale status` / Windows tray menu would only show nodes that are allowed for that particular user/device via ACLs.

Yes, if we consider an ACLs that allows A to talk to B but not for B to talk to A, in the `status` command of tailscale on node A we can see node B. But we cannot see node A from the `status` command executed on B before A send a packet. After it's visible but you cannot initiate connexion to A from B. Sorry if it's not easy to understand.

restanrm:
>

restanrm:
> This adaptation of the ACL would need breaking changes, but I also see no problem with that as the project is still in constant developed and we won't need a backwards compatibility.
> 
> Furthermore if we complete rework the ACL maybe we also should add the ACL configuration to a API. So a more dynamical configuration is possible. As for now we would need to change the configuration file, which will need a restart/reload of the server.
> 
> All in all I think that is a great proposal.
> 
> I also would love to write the ACL in Yaml. Not important just nice to have.

I consider the ACL configuration as an API a very good thing to do. Restart Headscale to reload ACLs is not quite good. 

I didn't think of YAML to write configuration but it should be nice and quite easy to implement !

restanrm:
> 



> I think this makes sense. I think it has been like this since the usecases of the maintainers has currently been 100% personal usage and we just have not ran into this. But I can see how it is better.
> 
> My other comments should probably be read before this, as this is the last section:
> 
> However my main concern is the migration path.
> 
> I am contemplating two relativeliy different one:
> 
>     1. Change the logic
> 
> 
> By this I mean, implement the new proposed solution, allowing access between all namespaces and fixing the ACL system. As a part of allowing access, remove the shared node code, table, tests etc.
> 
> It would be a breaking change, but we are not 1.0.0, so that "should be fine" tm.
> 
>     2. "rename" namespace to user
> 
> 
> This is a "even more" breaking change, while I think it would make sense to clean up the naming to avoid confusion and bring it closer to the upstream lingo.
> 
> Basically this would:
> 
>     * Implement the new suggestion as User (to reflect what it actually is)
> 
>     * Remove share node feature
> 
>     * Remove the namespace (cli, table, code, etc).
> 
> 
> This would make the user/group keywords in ACLs make more sense aswell.
> 
> Before doing either, we should make sure we "get it right", I can't see any concerns right now, but would like to get some more eyes on it.
> 
> @juanfont @madjam002 @ohdearaugustin @cure
> 
> Edit: I also want to write the ACLs in YAML (as well as HUJSON), but this should be trivial.

Thank you for you extensive review of this ! I agree with all points raised here. All those features are quite too much in my opinion for a single pull request. For migration path, we could either go slow and not change too much default behavior. Either by adding a flag to enable the new version of the ACLs or only enabling behavior if ACLs are enabled or completely break everything since it's a pre 1.0 release. 

Your second part of the migration path seems to me to be the natural next step of this rework. Removing the `namespace` cli seems a lot because not every users would be using OIDC. I think we should rename this cli but keeping it is a must have. For a personal instance OIDC is often too much a burden to setup.

kradalby:
I think that this document should be moved into `docs/proposals` or something, and maybe have a `001-` in front.
Proposal is a bit more versatile than Issues since it is easier to comment on the actual text.

> I consider the ACL configuration as an API a very good thing to do. Restart Headscale to reload ACLs is not quite good.

Then I think that we should make a new proposal for this, because it does require some design and concern:

- How do we maintain the current setup, allowing "config as code"
- were do we store/update changes and how do we treat it comparing to the already defined as code

> Removing the namespace cli seems a lot because not every users would be using OIDC. I think we should rename this cli but keeping it is a must have. For a personal instance OIDC is often too much a burden to setup.

I recon it would make sense to rename it in any case, for non-OIDC setups aswell, my argument for this is to make it more aligned with the upstream, allowing us to scrap extra documentation to explain what stuff means. 

And you could probably argue that when all machines can talk to all machines, it makes more sense to associate them with a user, rather than namespace. User is also used in the ACL IIRC?

This would probably also make sense to have in the proposal folder.


> All those features are quite too much in my opinion for a single pull request. For migration path, we could either go slow and not change too much default behavior.

I agree, lets do it gradually, and do it from 0.14.0, since 0.13.0 is quite a big release already.

> Either by adding a flag to enable the new version of the ACLs or only enabling behavior if ACLs are enabled or completely break everything since it's a pre 1.0 release.

I will personally vote for full change rather than "flag guarding" it, as it just means we need to maintain more logic.

restanrm:
Ok, I've moved the content into a proposals folder. Do you want me to try and write some other proposals for new `user` command and `acl` command or should we do this in another pull request ? 

As for this format, I think we will loose the comments. I could modify this document to try and integrate your comments, but I think we will still loose informations.

docs/acls.md
restanrm: I'll drop this sentence then. It's not in the scope of this document and it could be discussed when it's time to do it.

kradalby:
> Ok, I've moved the content into a proposals folder. Do you want me to try and write some other proposals for new `user` command and `acl` command or should we do this in another pull request ? 

Let us do that in separate PRs, one (big) problem at the time.

kradalby (approved):

PR #316, opened 2022-02-12
Remove kv-namespace-worker

kradalby:
This commit removes the namespace kv worker and related code, now that
we talk over gRPC to the server, and not directly to the DB, we should
not need this anymore.

Basically, I think this should be safe now, but not sure, let's see what the tests say. What do you think @juanfont 

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md


juanfont (approved):
This looks good!

kradalby:
Merging this as I believe it's safe, but we can keep in mind if weird things happen

Issue #318, opened 2022-02-13
Issues in kustomize example


leigh-johnson:
**Bug description**

:wave: Thank you for all of the hard work on headscale + thanks to the community contributors who created the kustomize example.

I'm running through the [kustomize example](https://github.com/juanfont/headscale/tree/main/docs/examples/kustomize) and noticed an error in the init script.

I understand this is a community-contributed configuration, so I'll use this issue to document my experience and submit patches as I go. =) Let me know who I should tag for review on these.

**To Reproduce**

### Issue 318.1: kustomize init.bash error: cp: cannot stat '../derp.yaml': No such file or directory

1. clone repo && cd headscale/docs/examples/kustomize
2. run init.bash
3. notice: example derp.yml was moved or deleted. 
```
$./init.bash 
++ dirname ./init.bash
+ cd .
+ umask 022
+ mkdir -p base/site/
+ '[' '!' -e base/site/public.env ']'
+ cat
+ '[' '!' -e base/site/derp.yaml ']'
+ cp ../derp.yaml base/site/derp.yaml
cp: cannot stat '../derp.yaml': No such file or directory
```

**Context info**
```
$ git rev-parse HEAD
c794f32f586c1e900ead2c1b5a24bf76d2875aea
```

Issue #228, opened 2021-11-18
Implement Helmchart


ohdearaugustin:
Create a helm chart for headscale with following features:

- headscale with sqlite
- headsacle with postgres
- headscale with cockroachdb

I started already with implementing it on base of the kustomization files includes here.
Could be found https://github.com/ohdearaugustin/charts/tree/topic/headscale

Open for further discussion.

kradalby:
I am for this, I think I also want to have the `k8s/` folder removed (or updated and structured) as part of this. It seems largely unmaintained and generates support requests.

gabe565:
I have built a Helm chart for Headscale if anybody is interested! It has autogenerated documentation and uses the bjw-s common library (formerly the k8s-at-home common library), so it is very configurable. I have been using it for a few weeks without any issues!

See the docs on [Artifact Hub](https://artifacthub.io/packages/helm/gabe565/headscale) or in [gabe565/charts](https://github.com/gabe565/charts/tree/main/charts/headscale).

PR #319, opened 2022-02-13
Fix errror in kustomize init.bash

leigh-johnson:
- [] added unit tests (N/A)
- [] added integration tests (N/A)
- [] updated documentation if needed (N/A)
- [] updated CHANGELOG.md (N/A)

* Fixes issue 318.1 reported in #318
* shellcheck linter (shellcheck defaults)
* Added check for `wg` and print version for future debugging/issue reporting quality of life :raised_hands:

### DIFF
--- a/docs/examples/kustomize/init.bash
+++ b/docs/examples/kustomize/init.bash
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -eux
-cd $(dirname $0)
+cd "$(dirname "$0")"
 
 umask 022
 mkdir -p base/site/
--- a/docs/examples/kustomize/init.bash
+++ b/docs/examples/kustomize/init.bash
@@ -30,6 +30,9 @@ regions:
 EOF
 )
 
+wg_version=$(wg --version || (echo "wg command not found. Please install wireguard and try again" && exit 1))
+echo "Using wireguard version: $wg_version"
+
 umask 077
 mkdir -p base/secrets/
 [ ! -e base/secrets/private-key ] && (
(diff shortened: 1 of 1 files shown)


ohdearaugustin:
Is the configuration working with only random data for the derp?

leigh-johnson:
> Is the configuration working with only random data for the derp?

No, the demo values would need to be replaced with custom DERP servers. https://tailscale.com/kb/1118/custom-derp-servers/

Not sure what the maintenance goals are for this demo, but happy to lend a hand if this is helpful. Otherwise, no worries about closing this PR. After running through the demo, I created my own deployment templates to support a different Ingress and external Postgres host.

ohdearaugustin:
As by now the kustuomize templates are not really maintained by anyone. And the general idea was to move to a helm chart. See #228. Just didn't managed to have time for it. We also were thinking about removing this configuration.

I would not prefer to add configuration that is not "working" too.

Maybe you could add some comment to the other issue for what the helm chart should be supporting.

leigh-johnson:
> As by now the kustuomize templates are not really maintained by anyone. And the general idea was to move to a helm chart. See #228. Just didn't managed to have time for it. We also were thinking about removing this configuration.
> 
> I would not prefer to add configuration that is not "working" too.
> 
> Maybe you could add some comment to the other issue for what the helm chart should be supporting.

Sounds good, thanks for the link to that issue about helm! 

What do you think of adding a more explicit deprecation notice here? The recent-ish commit plus note about being community-maintained is what made me contribute this patch, which just gets the end-to-end demo working without errors.
https://github.com/juanfont/headscale/tree/main/docs/examples/kustomize

PR #320, opened 2022-02-14
Improvements on the ACLs and bug fixing

restanrm:
- [x] added unit tests
- [] added integration tests
- [x] updated CHANGELOG.md

This PR is a first implementation of what has been discussed on #311 It should be reviewed commit by commit since last commits are linting modifications and changes quite much some part of the code that I didn't touch.

All subject discussed in PR #311 are not implemented here. I think all those modifications should be splitted in multiple PRs.

If this PR is too big, some commits could be moved outside of this PR it's related but just fixes some bugs on the ACLs parsing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 **TBD (TBD):**
 
 **BREAKING**:
-- ACLs have been rewritten and the behavior is different from before. It's now more aligned to tailscale's view of the feature. Namespaces are viewed as users and can communicate with each others. Tags should now work correctly and adding a host to Headscale should now reload the rules. The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features.
 
+- ACLs have been rewritten and the behavior is different from before. It's now more aligned to tailscale's view of the feature. Namespaces are viewed as users and can communicate with each others. Tags should now work correctly and adding a host to Headscale should now reload the rules. The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features.
 
 **0.13.0 (2022-xx-xx):**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,24 @@
 
 **TBD (TBD):**
 
+**0.14.0 (2022-xx-xx):**
+
+**UPCOMING BREAKING**:
+From the **next** version (`0.15.0`), all machines will be able to communicate regardless of 
+if they are in the same namespace. This means that the behaviour currently limited to ACLs 
+will become default. From version `0.15.0`, all limitation of communications must be done 
+with ACLs.
+
+This is a part of aligning `headscale`'s behaviour with Tailscale's upstream behaviour.
+
 **BREAKING**:
 
-- ACLs have been rewritten and the behavior is different from before. It's now more aligned to tailscale's view of the feature. Namespaces are viewed as users and can communicate with each others. Tags should now work correctly and adding a host to Headscale should now reload the rules. The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features.
+- ACLs have been rewritten to align with the bevaviour Tailscale Control Panel provides. **NOTE:** This is only active if you use ACLs
+  - Namespaces are now treated as Users
+  - All machines can communicate with all machines by default
+  - Tags should now work correctly and adding a host to Headscale should now reload the rules. 
+  - The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features
+
 
 **0.13.0 (2022-xx-xx):**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 **TBD (TBD):**
 
+**BREAKING**:
+- ACLs have been rewritten and the behavior is different from before. It's now more aligned to tailscale's view of the feature. Namespaces are viewed as users and can communicate with each others. Tags should now work correctly and adding a host to Headscale should now reload the rules. The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features.
+
+
 **0.13.0 (2022-xx-xx):**
 
 **Features**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@
 **0.14.0 (2022-xx-xx):**
 
 **UPCOMING BREAKING**:
-From the **next** version (`0.15.0`), all machines will be able to communicate regardless of 
-if they are in the same namespace. This means that the behaviour currently limited to ACLs 
-will become default. From version `0.15.0`, all limitation of communications must be done 
+From the **next** version (`0.15.0`), all machines will be able to communicate regardless of
+if they are in the same namespace. This means that the behaviour currently limited to ACLs
+will become default. From version `0.15.0`, all limitation of communications must be done
 with ACLs.
 
 This is a part of aligning `headscale`'s behaviour with Tailscale's upstream behaviour.
@@ -17,7 +17,7 @@ This is a part of aligning `headscale`'s behaviour with Tailscale's upstream beh
 - ACLs have been rewritten to align with the bevaviour Tailscale Control Panel provides. **NOTE:** This is only active if you use ACLs
   - Namespaces are now treated as Users
   - All machines can communicate with all machines by default
-  - Tags should now work correctly and adding a host to Headscale should now reload the rules. 
+  - Tags should now work correctly and adding a host to Headscale should now reload the rules.
   - The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features
 
 **0.13.0 (2022-02-18):**
--- a/acls.go
+++ b/acls.go
@@ -20,7 +20,6 @@ const (
 	errInvalidUserSection = Error("invalid user section")
 	errInvalidGroup       = Error("invalid group")
 	errInvalidTag         = Error("invalid tag")
-	errInvalidNamespace   = Error("invalid namespace")
 	errInvalidPortFormat  = Error("invalid port format")
 )
 
@@ -69,6 +68,7 @@ func (h *Headscale) LoadACLPolicy(path string) error {
 	}
 
 	h.aclPolicy = &policy
+
 	return h.UpdateACLRules()
 }
 
@@ -79,6 +79,7 @@ func (h *Headscale) UpdateACLRules() error {
 	}
 	log.Trace().Interface("ACL", rules).Msg("ACL rules generated")
 	h.aclRules = rules
+
 	return nil
 }
 
@@ -182,7 +183,7 @@ func (h *Headscale) generateACLPolicyDestPorts(
 // - a namespace
 // - a group
 // - a tag
-// and transform these in IPAddresses
+// and transform these in IPAddresses.
 func expandAlias(machines []Machine, aclPolicy ACLPolicy, alias string) ([]string, error) {
 	ips := []string{}
 	if alias == "*" {
@@ -200,6 +201,7 @@ func expandAlias(machines []Machine, aclPolicy ACLPolicy, alias string) ([]strin
 				ips = append(ips, node.IPAddresses.ToStringSlice()...)
 			}
 		}
+
 		return ips, nil
 	}
 
@@ -225,6 +227,7 @@ func expandAlias(machines []Machine, aclPolicy ACLPolicy, alias string) ([]strin
 				}
 			}
 		}
+
 		return ips, nil
 	}
 
@@ -276,6 +279,7 @@ func excludeCorrectlyTaggedNodes(aclPolicy ACLPolicy, nodes []Machine, namespace
 	for _, machine := range nodes {
 		if len(machine.HostInfo) == 0 {
 			out = append(out, machine)
+
 			continue
 		}
 		hi, err := machine.GetHostInfo()
@@ -286,6 +290,7 @@ func excludeCorrectlyTaggedNodes(aclPolicy ACLPolicy, nodes []Machine, namespace
 		for _, t := range hi.RequestTags {
 			if containsString(tags, t) {
 				found = true
+
 				break
 			}
 		}
@@ -293,6 +298,7 @@ func excludeCorrectlyTaggedNodes(aclPolicy ACLPolicy, nodes []Machine, namespace
 			out = append(out, machine)
 		}
 	}
+
 	return out, nil
 }
 
@@ -346,42 +352,45 @@ func listMachinesInNamespace(machines []Machine, namespace string) []Machine {
 			out = append(out, machine)
 		}
 	}
+
 	return out
 }
 
 // expandTagOwners will return a list of namespace. An owner can be either a namespace or a group
-// a group cannot be composed of groups
+// a group cannot be composed of groups.
 func expandTagOwners(aclPolicy ACLPolicy, tag string) ([]string, error) {
 	var owners []string
 	ows, ok := aclPolicy.TagOwners[tag]
 	if !ok {
 		return []string{}, fmt.Errorf("%w. %v isn't owned by a TagOwner. Please add one first. https://tailscale.com/kb/1018/acls/#tag-owners", errInvalidTag, tag)
 	}
-	for _, ow := range ows {
-		if strings.HasPrefix(ow, "group:") {
-			gs, err := expandGroup(aclPolicy, ow)
+	for _, owner := range ows {
+		if strings.HasPrefix(owner, "group:") {
+			gs, err := expandGroup(aclPolicy, owner)
 			if err != nil {
 				return []string{}, err
 			}
 			owners = append(owners, gs...)
 		} else {
-			owners = append(owners, ow)
+			owners = append(owners, owner)
 		}
 	}
+
 	return owners, nil
 }
 
 // expandGroup will return the list of namespace inside the group
-// after some validation
+// after some validation.
 func expandGroup(aclPolicy ACLPolicy, group string) ([]string, error) {
-	gs, ok := aclPolicy.Groups[group]
+	groups, ok := aclPolicy.Groups[group]
 	if !ok {
 		return []string{}, fmt.Errorf("group %v isn't registered. %w", group, errInvalidGroup)
 	}
-	for _, g := range gs {
+	for _, g := range groups {
 		if strings.HasPrefix(g, "group:") {
 			return []string{}, fmt.Errorf("%w. A group cannot be composed of groups. https://tailscale.com/kb/1018/acls/#groups", errInvalidGroup)
 		}
 	}
-	return gs, nil
+
+	return groups, nil
 }
--- a/acls.go
+++ b/acls.go
@@ -129,7 +129,11 @@ func (h *Headscale) generateACLRules() ([]tailcfg.FilterRule, error) {
 	return rules, nil
 }
 
-func (h *Headscale) generateACLPolicySrcIP(machines []Machine, aclPolicy ACLPolicy, u string) ([]string, error) {
+func (h *Headscale) generateACLPolicySrcIP(
+	machines []Machine,
+	aclPolicy ACLPolicy,
+	u string,
+) ([]string, error) {
 	return expandAlias(machines, aclPolicy, u)
 }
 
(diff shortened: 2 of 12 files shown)


kradalby:
Can you have a look at the conflicts :)? all the other stuff is in now, and @juanfont will have a look at the proposal.

restanrm:
Sorry I didn't saw your comment I'll look at it tomorrow. I'll need to work on integration tests also.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 **TBD (TBD):**
 
+**BREAKING**:
+
+- ACLs have been rewritten and the behavior is different from before. It's now more aligned to tailscale's view of the feature. Namespaces are viewed as users and can communicate with each others. Tags should now work correctly and adding a host to Headscale should now reload the rules. The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features.
```
kradalby, line 7:
```suggestion
- ACLs have been rewritten to align with the bevaviour Tailscale Control Panel provides. **NOTE:** This is only active if you use ACLs
  - Namespaces are now treated as Users
  - All machines can communicate with all machines by default
  - Tags should now work correctly and adding a host to Headscale should now reload the rules. 
  - The documentation have a [fictional example](docs/acls.md) that should cover some use cases of the ACLs features

```

Not sure if this is better, I tried to break it up a bit.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 **TBD (TBD):**
 
+**BREAKING**:
```
kradalby, line 5:
```suggestion
**0.14.0 (2022-xx-xx):**

**BREAKING**:
```

utils.go
```diff
@@ -212,6 +212,16 @@ func (h *Headscale) getUsedIPs() ([]netaddr.IP, error) {
 	return ips, nil
 }
 
+func containsString(ss []string, s string) bool {
```
kradalby, line 215: Man, I am looking forward to Generics...

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+		log.Error().Err(err).Msg("Error accessing db")
+
+		return Machines{}, err
+	}
+	mMachines := make(map[uint64]Machine)
+
+	// Aclfilter peers here. We are itering through machines in all namespaces and search through the computed aclRules
+	// for match between rule SrcIPs and DstPorts. If the rule is a match we allow the machine to be viewable.
+
+	// FIXME: On official control plane if a rule allow user A to talk to user B but NO rule allows user B to talk to
```
kradalby, line 164: This should be a github issue

kradalby (commented):
Some starters, have not started on the code yet

acls.go
```diff
@@ -306,3 +356,64 @@ func (h *Headscale) expandPorts(portsStr string) (*[]tailcfg.PortRange, error) {
 
 	return &ports, nil
 }
+
+func listMachinesInNamespace(machines []Machine, namespace string) []Machine {
```
kradalby, line 360: Maybe this could be called `filterMachinesByNamespaces`?

acls_test.go
```diff
@@ -52,6 +59,244 @@ func (s *Suite) TestBasicRule(c *check.C) {
...
+	c.Assert(err, check.IsNil)
+
+	_, err = app.GetMachine("foo", "testmachine")
+	c.Assert(err, check.NotNil)
+	hostInfo := []byte(
+		"{\"OS\":\"centos\",\"Hostname\":\"foo\",\"RequestTags\":[\"tag:foo\"]}",
+	)
+	machine := Machine{
+		ID:             1,
+		MachineKey:     "foo",
```
kradalby, line 203:
```suggestion
		MachineKey:     "12345",
```

The machine key and the tag should not have the same value as that could confuse someone (me :P), but more seriously ensure that those two variables dont overlap internally in the code.

acls_test.go
```diff
@@ -52,6 +59,244 @@ func (s *Suite) TestBasicRule(c *check.C) {
...
+	namespace, err := app.CreateNamespace("foo")
+	c.Assert(err, check.IsNil)
+
+	pak, err := app.CreatePreAuthKey(namespace.Name, false, false, nil)
+	c.Assert(err, check.IsNil)
+
+	_, err = app.GetMachine("foo", "testmachine")
+	c.Assert(err, check.NotNil)
+	hostInfo := []byte(
+		"{\"OS\":\"centos\",\"Hostname\":\"foo\",\"RequestTags\":[\"tag:foo\"]}",
```
kradalby, line 199:
```suggestion
		"{\"OS\":\"centos\",\"Hostname\":\"testmachine\",\"RequestTags\":[\"tag:foo\"]}",
```

The hostname and the machine name should align in all the tests I believe.

acls_test.go
```diff
@@ -52,6 +59,244 @@ func (s *Suite) TestBasicRule(c *check.C) {
...
+		RegisterMethod: RegisterMethodAuthKey,
+		AuthKeyID:      uint(pak.ID),
+		HostInfo:       datatypes.JSON(hostInfo),
+	}
+	app.db.Save(&machine)
+
+	app.aclPolicy = &ACLPolicy{
+		TagOwners: TagOwners{"tag:test": []string{"foo"}},
+		ACLs: []ACL{
+			{Action: "accept", Users: []string{"foo"}, Ports: []string{"*:*"}},
```
kradalby, line 219: Is the "foo" in `Users` based on the tag or the machine name?

acls_test.go
```diff
@@ -52,6 +59,244 @@ func (s *Suite) TestBasicRule(c *check.C) {
...
+	}
+	err := app.UpdateACLRules()
+	c.Assert(errors.Is(err, errInvalidTag), check.Equals, true)
+}
+
+// this test should validate that we can expand a group in a TagOWner section and
+// match properly the IP's of the related hosts. The owner is valid and the tag is also valid.
+// the tag is matched in the Users section.
+func (s *Suite) TestValidExpandTagOwnersInUsers(c *check.C) {
+	namespace, err := app.CreateNamespace("foo")
```
kradalby, line 102:
```suggestion
	namespace, err := app.CreateNamespace("validate-tagowner-expands-in-user")
```

Test values should be descriptive and independent, this makes it a lot easier to follow the tests as you read them without having to go up and down to check if `foo` was a hostname or `bar` was a tag, group, or user. 

If they have description based on the test their a part of, e.g. `tagValidateTagOwnerExpandsInUser` for tag, it will also guarantee that the tag is independent and will not interfere with other parts of the tests (even if the test framework is suppose to sort this.

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+
+// getFilteredByACLPeerss should return the list of peers authorized to be accessed from machine.
+func (h *Headscale) getFilteredByACLPeers(machine *Machine) (Machines, error) {
+	log.Trace().
+		Caller().
+		Str("machine", machine.Name).
+		Msg("Finding peers filtered by ACLs")
+
+	machines := Machines{}
+	if err := h.db.Preload("Namespace").Where("machine_key <> ? AND registered",
```
kradalby, line 153: Why does this not call `ListAllMachines` ?

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+	// how to talk to node A and then add the peering resource.
+
+	for _, mchn := range machines {
+		for _, rule := range h.aclRules {
+			var dst []string
+			for _, d := range rule.DstPorts {
+				dst = append(dst, d.IP)
+			}
+			if (containsAddresses(rule.SrcIPs, machine.IPAddresses) && (containsAddresses(dst, mchn.IPAddresses) || containsString(dst, "*"))) ||
+				(containsAddresses(rule.SrcIPs, mchn.IPAddresses) && containsAddresses(dst, machine.IPAddresses)) {
```
kradalby, line 185: Could this conditional be factored out to a function? that way we can give it a name that is easier to comprehend.

So something with a signature like `(rule Rule, source Machine, destination Machine) bool {}`

kradalby: I suppose one function with that signature that is called twice, with the `||` in the middle maybe 🤔

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+	//   We can successfully establish a communication from A to B. When it's done, if we run the `tailscale status` command
+	//   on node B again we can now see node A. It's not possible to establish a communication from node B to node A.
+	// On this implementation of the feature
+	//   on any `tailscale status` command on node A we can see node B. The `tailscale status` command on node B DOES show A.
+	//
+	// I couldn't find a way to not clutter the output of `tailscale status` with all nodes that we could be talking to.
+	// In order to do this we would need to be able to identify that node A want to talk to node B but that Node B doesn't know
+	// how to talk to node A and then add the peering resource.
+
+	for _, mchn := range machines {
```
kradalby, line 178:
```suggestion
	for _, peer := range machines {
```

This could be called peer if im not mistaken?

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+		Msg("Finding peers filtered by ACLs")
+
+	machines := Machines{}
+	if err := h.db.Preload("Namespace").Where("machine_key <> ? AND registered",
+		machine.MachineKey).Find(&machines).Error; err != nil {
+		log.Error().Err(err).Msg("Error accessing db")
+
+		return Machines{}, err
+	}
+	mMachines := make(map[uint64]Machine)
```
kradalby, line 159: peers?

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+				dst = append(dst, d.IP)
+			}
+			if (containsAddresses(rule.SrcIPs, machine.IPAddresses) && (containsAddresses(dst, mchn.IPAddresses) || containsString(dst, "*"))) ||
+				(containsAddresses(rule.SrcIPs, mchn.IPAddresses) && containsAddresses(dst, machine.IPAddresses)) {
+				mMachines[mchn.ID] = mchn
+			}
+		}
+	}
+
+	authorizedMachines := make([]Machine, 0, len(mMachines))
```
kradalby, line 191: if we also use `authorizedPeers` here, we signalise that this is a list of adjacent machines, not included the input one.

kradalby (commented):
I think this is looking very promising!

I have left comments on the "input data" for tests for the ACLs, its quite hard to follow when using foo, bar, etc. Would be great if you could change that to descriptive, unique values and I will review that section again.

Otherwise, the expand* tests looks quite sensible and well covering, easy to extend!

There is a couple of other naming nits, but mostly good, thanks.

machine.go
restanrm: Nice suggestion. I had some reserve for this since the previous computations of dst list would need to be done twice, and I wasn't sure about the check with the destination containing '*'. It seems ok though to combine them.

machine.go
restanrm: I'm sorry, I don't understand what you mean here.

utils.go
restanrm: So am I :smile:

machine.go
kradalby: I was thinking it makes more sense to name the list peers instead of machines, as i think it contains only "peers of the passed machine"

But I might be mistaken

restanrm: ok I'll rename that too

machine.go
```diff
@@ -119,6 +119,95 @@ func (machine Machine) isExpired() bool {
...
+	// On this implementation of the feature
+	//   on any `tailscale status` command on node A we can see node B. The `tailscale status` command on node B DOES show A.
+	//
+	// I couldn't find a way to not clutter the output of `tailscale status` with all nodes that we could be talking to.
+	// In order to do this we would need to be able to identify that node A want to talk to node B but that Node B doesn't know
+	// how to talk to node A and then add the peering resource.
+
+	for _, peer := range machines {
+		for _, rule := range h.aclRules {
+			if matchSourceAndDestinationWithRule(rule, *machine, peer) || matchSourceAndDestinationWithRule(rule, peer, *machine) {
```
restanrm, line 188: This modification broke the tests I'm looking into it

restanrm: So, I had to revert the modifications on this. I added some comment for the reasons in the commit description. We need to find a better approach. The first try wasn't good enough.

acls_test.go
restanrm: The "foo" in `Users` was a namespace. I renamed this, I hope it's easier to read now.

acls_test.go
restanrm: I changed for more explicit names, `username` for namespaces, `teams` for groups and groups of servers for tags. It's better than before but not up to what you describe here.

kradalby: Better thanks, when using the same values in adjacent tests, we don't make them independent of each other, but we can let this go for now and I can do a follow up PR and you have help me review it.

machine.go
kradalby, line 153:
```suggestion
	# TODO: Factor this out to a `ListPeers(m Machine)` function listing all except the passed Machine
	if err := h.db.Preload("Namespace").Where("machine_key <> ? AND registered",
```

Let us do this for now, I have some ideas for cleanups to this file when we have merged, so let's not block it based on that.

machine.go
```diff
@@ -119,6 +119,92 @@ func (machine Machine) isExpired() bool {
...
+	// In order to do this we would need to be able to identify that node A want to talk to node B but that Node B doesn't know
+	// how to talk to node A and then add the peering resource.
+
+	for _, peer := range machines {
+		for _, rule := range h.aclRules {
+			var dst []string
+			for _, d := range rule.DstPorts {
+				dst = append(dst, d.IP)
+			}
+			if (containsAddresses(rule.SrcIPs, machine.IPAddresses) && (containsAddresses(dst, peer.IPAddresses) || containsString(dst, "*"))) || (
```
kradalby, line 183:
```suggestion
			# TODO: Make this a bit easier to read by making a wrapper function.
			if (containsAddresses(rule.SrcIPs, machine.IPAddresses) && (containsAddresses(dst, peer.IPAddresses) || containsString(dst, "*"))) || (
```

acls_test.go
```diff
@@ -52,6 +59,244 @@ func (s *Suite) TestBasicRule(c *check.C) {
 	c.Assert(rules, check.NotNil)
 }
```
kradalby, line 61:
```suggestion

# TODO(kradalby): Make tests values safe, independent and descriptive.
```

Lets add this so I remember, I think its easier for me to show, rather than trying to explain, because I am not that good at formulating these arguments.

kradalby (commented):
Just a couple more nits and I think we can take this for a spin.

kradalby:
Could you run prettier lint?

And I see the integration test fail, but I can't really see why, and not sure if any of the changes should really affect it. 

Could you have a dive?

Other than that I think I'm happy to get it in :)

restanrm:
> Could you run prettier lint?
> 
> And I see the integration test fail, but I can't really see why, and not sure if any of the changes should really affect it.
> 
> Could you have a dive?
> 
> Other than that I think I'm happy to get it in :)

Linting is done. For the integrations tests, I see some mentions of ACL parsing in the logs and calls seems to go through `getFilteredByACLPeers`. I couldn't find where the ACL are read from. It can be a good explanation for the errors of the integrations tests. It's also possible that my code that check presence of ACLs is wrong. I'll try to dive into it tonight.

kradalby:
It looks like there is nil pointer as a result to a call to the `/map` endpoint:

```
2022/02/21 17:17:03 [Recovery] 2022/02/21 - 17:17:03 panic recovered:
POST /machine/3efd0dd32dc939483bad4d7bf4a5ce2af31a1a0a294c7a4ece72a16918c06c17/map HTTP/1.1
Host: headscale:8080
Accept-Encoding: gzip
Content-Length: 641
User-Agent: Go-http-client/1.1


runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/runtime/panic.go:221 (0x44d686)
/usr/local/go/src/runtime/signal_unix.go:735 (0x44d656)
/go/src/headscale/acls.go:94 (0xcee90a)
/go/src/headscale/acls.go:76 (0xcee7c4)
/go/src/headscale/poll.go:98 (0xd0ce39)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/context.go:168 (0xb6ceb4)
/go/pkg/mod/github.com/zsais/go-gin-prometheus@v0.1.0/middleware.go:364 (0xb6ce95)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/context.go:168 (0x946dc1)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/recovery.go:99 (0x946dac)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/context.go:168 (0x946026)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/logger.go:241 (0x946009)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/context.go:168 (0x945570)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/gin.go:555 (0x9451d8)
/go/pkg/mod/github.com/gin-gonic/gin@v1.7.7/gin.go:511 (0x944d11)
/usr/local/go/src/net/http/server.go:2879 (0x75457a)
/usr/local/go/src/net/http/server.go:1930 (0x74fc27)
/usr/local/go/src/runtime/asm_amd64.s:1581 (0x469240)
```

Ill see if I can come up with a suggestion before I have to go out again.

poll.go
```diff
@@ -85,12 +85,25 @@ func (h *Headscale) PollNetMapHandler(ctx *gin.Context) {
...
+	if err != nil {
+		return
+	}
 	machine.Name = req.Hostinfo.Hostname
 	machine.HostInfo = datatypes.JSON(hostinfo)
 	machine.DiscoKey = DiscoPublicKeyStripPrefix(req.DiscoKey)
 	now := time.Now().UTC()
 
+	// update ACLRules with peer informations (to update server tags if necessary)
+	err = h.UpdateACLRules()
```
kradalby, line 98: Either this should be wrapped with `if h.aclPolicy != nil` or (see other comment)

acls.go
```diff
@@ -69,30 +68,37 @@ func (h *Headscale) LoadACLPolicy(path string) error {
...
 
 func (h *Headscale) generateACLRules() ([]tailcfg.FilterRule, error) {
 	rules := []tailcfg.FilterRule{}
 
+	machines, err := h.ListAllMachines()
+	if err != nil {
+		return nil, err
+	}
+
 	for index, acl := range h.aclPolicy.ACLs {
```
kradalby, line 94: This could be `nil`, if ACLs are not active, and the new `h.UpdateACLRules` is called, which it now is in each poll handler.

We should probably wrap it on the other side.

poll.go
```diff
@@ -85,12 +85,25 @@ func (h *Headscale) PollNetMapHandler(ctx *gin.Context) {
...
+	if err != nil {
+		log.Error().
+			Caller().
+			Str("func", "handleAuthKey").
+			Str("machine", machine.Name).
+			Err(err)
+	}
```
kradalby, line 105:
```suggestion
	if h.aclPolicy != nil {
		err = h.UpdateACLRules()
		if err != nil {
			log.Error().
				Caller().
				Str("func", "handleAuthKey").
				Str("machine", machine.Name).
				Err(err)
		}
	}

```

This should do it.

kradalby: But open to better solutions.

kradalby:
As a side note, we should probably have integration tests for ACL, but that can be done in the future, want to clean them up and remove all the shared stuff first

poll.go
restanrm: I've opted for both adding error to the callee and avoiding error case on caller side. The callee still need to report and error since it can generate errors if no ACLPolicies are defined.

As for the placement of this piece of code, I'm also open to suggestion. I've put it here but don't know well enough the code to find better place. We need to update the ACL when the HostInfo of the machines are modified, and when hosts are registered.

restanrm:
> As a side note, we should probably have integration tests for ACL, but that can be done in the future, want to clean them up and remove all the shared stuff first

Yes I wanted to do it but here, but the PR is big enough. We should create a issue for this subject as well.

acls_test.go
restanrm: With pleasure !

acls_test.go
kradalby: oh wops

kradalby (approved):

kradalby:
Good job, lets do it, I think I will release this as a beta straight away.

kradalby:
Would you mind creating the issues mentioned in this PR @restanrm ?

restanrm:
> Would you mind creating the issues mentioned in this PR @restanrm ?

Yes, I'll create it now

PR #349, opened 2022-02-22
Remove dependency on CGO

kradalby:
This commit changes the SQLite dependency to one that does not depend on
CGO. It uses a C-to-Go translated sqlite library that is Pure go.

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md


juanfont:
Oh! Didn't know about this module. Cool!

cure:
Oh! Awesome! I didn't know about this library either!

kradalby:
There is one drawback, I had to set concurrent connections for SQLite to one, I am not sure if that actually has a practical difference as I suspect the whole thing is loaded to memory,

But can't confidently merge it until I have poked around a bit more.

I'll let you know what I'll find  and click ready.

kradalby:
Current state of this, I think I can look a bit more tomorrow:

1. I had a SQLITE_BUSY error based on Gorm opening a connection pool and this lib did not support multiple connection out of the box, so I set it to one single connection. 
 
Unanswered question: Does that actually matter in our case?
I would think probably not, I suspect it actually does not affect SQLite performance as much as for example postgres as the DB is likely in memory, we dont use it that heavily and if you need a high performance DB (or the scale were this is a problem) you should probably use postgres anyways? 

2. The pure Go library will be _somewhat_ slower

but for the same reasons as above, I dont think it matters over the other practical gains... Benchmarks: https://github.com/glebarez/go-sqlite/tree/master/benchmark

I'll try to investigate the single connection a bit, might be something I did, or it might not matter.

@juanfont  @cure thoughts on the benefits vs drawbacks?

cure:
I don't think that in principle, a single concurrent connection is going to be a problem for our use case. Any delays because of this are unlikely to be significant for small-ish sites. And if it becomes a problem, the site can always switch to PostgreSQL.

Maybe we could somehow measure connection contention and expose it as a metric? But that should probably be a separate story, and it may not be worth the effort.

I'm not concerned about the library being official or not. If it becomes unmaintained later, we can deal with that then. 

Getting rid of CGO is really nice and will speed up builds etc!

juanfont (approved):

PR #350, opened 2022-02-22
Feat OIDC login as namespace

### LINKED ISSUE #327: Replace `oidc.domain_map` with email to namespace
Depends on the release of #320 and #311.

When we have removed the barriers between namespaces (soon to be called users), we should replace the configuration option `oidc.domain_map` with logic that just maps the username/email from OIDC straight as the namespace.

This will simplify the setup quite a bit and was the originally suggested setup, but @kradalby did not understand `namespaces` == `users` at the time and made a different suggestion.

restanrm:
- [x] added unit tests
- [] added integration tests
- [x] updated CHANGELOG.md

should close #327


namespaces.go
```diff
@@ -233,3 +269,24 @@ func (n *Namespace) toProto() *v1.Namespace {
...
+// NormalizeNamespaceName will replace forbidden chars in namespace
+// it can also return an error if the namespace doesn't respect RFC 952 and 1123
+func NormalizeNamespaceName(name string) (string, error) {
+	name = strings.ToLower(name)
+	name = strings.ReplaceAll(name, "@", ".")
+	name = strings.ReplaceAll(name, "'", "")
+	name = normalizeNamespaceRegex.ReplaceAllString(name, "-")
+
+	for _, elt := range strings.Split(name, ".") {
+		if len(elt) > 63 {
```
kradalby: This is good, I think I would have forgotten this.

It makes me thing tho, do we need an option to only use the username or the name before `@`? I suspect that with emails, we can easily get too long names here?

reallylongfirstname.middlename.longlastname@reallylongdomainname.com

This will throw an error, but not sure if we will display it to users, and users might not have a clue whats going on.

namespaces.go
```diff
@@ -30,7 +36,12 @@ type Namespace struct {
 // CreateNamespace creates a new Namespace. Returns error if could not be created
 // or another namespace already exists.
 func (h *Headscale) CreateNamespace(name string) (*Namespace, error) {
+	var err error
 	namespace := Namespace{}
+	name, err = NormalizeNamespaceName(name)
```
kradalby, line 46: I do not think we should normalize what users pass to functions, it should be catched and validated _before_ it reaches this function. This is can cause unexpected behaviour for the dev, and it should not need to be done before methods like delete.

We should validate input to the `headscale namespace create` command, not change it, but rather return an error.

And we should normalize it in the OIDC logic, where we _know_ its not valid.

oidc.go
```diff
@@ -282,109 +281,90 @@ func (h *Headscale) OIDCCallback(ctx *gin.Context) {
...
-				log.Error().
-					Caller().
-					Err(err).
-					Str("namespace", namespaceName).
-					Msg("could not find or create namespace")
-				ctx.String(
-					http.StatusInternalServerError,
-					"could not find or create namespace",
-				)
+	namespaceName, err := NormalizeNamespaceName(claims.Email)
```
kradalby, line 284: Like this :)

kradalby (changes requested):
Some comments, did you update the changelog? 🤔

namespaces.go
restanrm: Yes this is also something that I thought about without reaching a result. I could add a check at some place where we return the hostname if it's too long. In the RFC 1123 we can read the following:

> Host software MUST handle host names of up to 63 characters and
> SHOULD handle host names of up to 255 characters.

If we throw an error like this we should add some comments on resolution or keep the `domain_map` config around to solve this kind of issues.

namespaces.go
restanrm: That's was my initial thought, I'll go back there !

namespaces.go
kradalby: hmm, the combinations of options here is a bit of a can of worms:

- Add too long ones into domain-only namespace
- strip domain and use just username
- what if username is still too long

I suggest keeping it a bit simple to begin with, and actually, I wonder if people dont _really_ need to have the domain from the email in the namespace (as they are all likely the same).

Maybe we should have a `oidc.strip_email_domain` which is true by default, so by default, username is used, and if people intend to have different email addresses, you turn that off and get the whole thing.

So by default:
alice@enterprise.com -> alice
bob@enterprise.com -> bob

But you can turn that off:
alice@enterprise.com -> alice.enterprice-com (or what it will become)
bob@collaboratingcompany.com -> bob.collaboratingcompany-com

> Host software MUST handle host names of up to 63 characters and
> SHOULD handle host names of up to 255 characters.

As for 63 vs 255, I suggest we saw that 255 is an error and 63 is a warning.

namespaces.go
restanrm: So, I didn't implemented in the client but server-side. Because if we use another mean to talk to the API, the check would not be implemented.

restanrm:
With this came to my mind that `hostname` validation is not done. @kradalby should we normalize the `hostname` received from the nodes ?

CHANGELOG.md
```diff
@@ -28,13 +28,15 @@ This is a part of aligning `headscale`'s behaviour with Tailscale's upstream beh
 - Add API Key support
   - Enable remote control of `headscale` via CLI [docs](docs/remote-cli.md)
   - Enable HTTP API (beta, subject to change)
+- OIDC login will map user to an associated namespace.
```
kradalby, line 31: When we have landed on how this work, we should elaborate here, and in the example config

namespaces.go
```diff
@@ -41,12 +41,11 @@ type Namespace struct {
 // CreateNamespace creates a new Namespace. Returns error if could not be created
 // or another namespace already exists.
 func (h *Headscale) CreateNamespace(name string) (*Namespace, error) {
-	var err error
-	namespace := Namespace{}
-	name, err = NormalizeNamespaceName(name)
+	err := CheckNamespaceName(name)
```
kradalby, line 44: I agree on the server side, and I think it should be checked in this function, and rename. But not checked in Get and Destroy. Get and Destroy should return a "namespace not found" not invalid.

kradalby (commented):
> With this came to my mind that `hostname` validation is not done. @kradalby should we normalize the `hostname` received from the nodes ?

Yes, but lets make an issue, so this can be tracked separately.

namespaces.go
restanrm: Good point, I'll make that change

CHANGELOG.md
restanrm: Should I remove changelog modifications ?

namespaces.go
restanrm: The potential issue that I can see with the `strip_email_domain` as default is that if some users want to login with different email provider it's possible to have an issue. It can even become a security issue (impersonating admin account). I don't know OIDC well enough to see if it can be an issue. But I guess with a tool like `dex` (login aggregator), it seems possible.

restanrm: My reference to RFC 1123 seems incomplete. RFC 2181 says the following:

> The DNS itself places only one restriction on the particular labels
   that can be used to identify resource records.  That one restriction
   relates to the length of the label and the full name.  The length of
   any one label is limited to between 1 and 63 octets.  A full domain
   name is limited to 255 octets (including the separators).

It seems to be the last RFC describing host names. The wikipedia page have the same information: https://en.wikipedia.org/wiki/Domain_Name_System

machine.go
```diff
@@ -25,6 +25,11 @@ const (
 	errMachineAlreadyRegistered   = Error("machine already registered")
 	errMachineRouteIsNotAvailable = Error("route is not available on machine")
 	errMachineAddressesInvalid    = Error("failed to parse machine addresses")
+	errHostnameTooLong            = Error("Hostname too long")
+)
+
+const (
+	maxHostnameLength = 255
```
kradalby, line 32: Actually, I just thought, is it the _entire_ url that has the max? Or the segments between . ?

I just had a memory about some kubernetes namespace stuff.

This is mostly to raise this, I'm happy about it as is for now.

kradalby (commented):
I'll have a suggestion for the Changelog when I get to the airport, but all looks good now :)

machine.go
restanrm: The Fully Qualified Domain Name has a maximum length of 255 and the labels (segments in between) have a maximum length of 63

kradalby: Ah yes of course

CHANGELOG.md
```diff
@@ -32,13 +32,15 @@ This is a part of aligning `headscale`'s behaviour with Tailscale's upstream beh
...
 
 **Changes**:
 
 - `ip_prefix` is now superseded by `ip_prefixes` in the configuration [#208](https://github.com/juanfont/headscale/pull/208)
 - Upgrade `tailscale` (1.20.4) and other dependencies to latest [#314](https://github.com/juanfont/headscale/pull/314)
 - fix swapped machine<->namespace labels in `/metrics` [#312](https://github.com/juanfont/headscale/pull/312)
 - remove key-value based update mechanism for namespace changes [#316](https://github.com/juanfont/headscale/pull/316)
+- removal of the `oidc.domain_map` configuration parameter
```
kradalby, line 43:
```suggestion
```

CHANGELOG.md
```diff
@@ -32,13 +32,15 @@ This is a part of aligning `headscale`'s behaviour with Tailscale's upstream beh
 - Add API Key support
   - Enable remote control of `headscale` via CLI [docs](docs/remote-cli.md)
   - Enable HTTP API (beta, subject to change)
+- OIDC login will map user to an associated namespace.
```
kradalby, line 35:
```suggestion
- OpenID Connect users will be mapped per namespaces
  - Each user will get its own namespace, created if it does not exist
  - `oidc.domain_map` option has been removed
  - `strip_email_domain` option has been added (see [config-example.yaml](./config_example.yaml))
```

kradalby (commented):
Lets do this and I'll approve :)

kradalby (approved):

restanrm:
Integrations tests seems to have a random component. One update on changelog was ok, the other was not…

kradalby:
> Integrations tests seems to have a random component. One update on changelog was ok, the other was not…

Sadly it's not random, but race conditions/sync issues in headscale. 

I am continuously working on them, and have an draft with a couple of improvements. But for now, just rerunning it is the way 😖

kradalby:
When #357  is in, can you resolve the conflicts? I think there will be some more, so you can wait for that one.

restanrm:
> When #357 is in, can you resolve the conflicts? I think there will be some more, so you can wait for that one.

Ok, I'll wait for #357

kradalby:
> Ok, I'll wait for #357

Merged now.

kradalby (approved):

PR #490, opened 2022-03-08
Ignore expired machine records - closes #489

### LINKED ISSUE #489: Tailscale client fails to re-authenticate using an auth key after logout
**Bug description**

If a Tailscale client that uses an auth key to authenticate logs out and attempts to re-authenticate using a new key the client hangs and headscale server panics.

**To Reproduce**

1) Use a new pre-auth key to authenticate a new Tailscale client with headscale
2) Client should login successfully.
3) Log the client out via the Tailscale client.
4) Create a new pre-auth key and attempt to re-authenticate via the Tailscale client.

**Context info**

- Version of headscale used: main/latest
- Version of tailscale client: 1.22.0
- OS (e.g. Linux, Mac, Cygwin, WSL, etc.) and version: macOS Big Sur 
- Log output

Headscale Log

```
2022-03-08T17:08:45Z DBG Machine registration has expired. Sending a authurl to register machine=test-machine
2022-03-08T17:08:45Z DBG Processing auth key for test-machine func=handleAuthKey machine=test-machine
2022-03-08T17:08:45Z DBG Authentication key was valid, proceeding to acquire IP addresses func=handleAuthKey machine=test-machine
2022-03-08T17:08:45Z INF Successfully authenticated via AuthKey func=handleAuthKey ips="fd7a:115c:a1e0::2, 100.64.0.2" machine=test-machine
[GIN] 2022/03/08 - 12:08:45 | 200 |    3.884024ms |   192.168.1.154 | POST     "/machine/221d92c201f65e402666072e3c223d73e1329603a0f8337350d6c9c4cc70fd60"
2022-03-08T17:08:45Z DBG Client map request processed handler=PollNetMap id=221d92c201f65e402666072e3c223d73e1329603a0f8337350d6c9c4cc70fd60 machine=test-machine omitPeers=false readOnly=false stream=true
2022-03-08T17:08:45Z INF Client is ready to access the tailnet handler=PollNetMap machine=test-machine
2022-03-08T17:08:45Z INF Sending initial map handler=PollNetMap machine=test-machine
2022-03-08T17:08:45Z INF Notifying peers handler=PollNetMap machine=test-machine
2022-03-08T17:08:45Z DBG There has been updates since the last successful update to test-machine handler=PollNetMapStream last_state_change=1646759325 last_successful_update=1646759307 machine=test-machine
2022-03-08T17:08:45Z DBG Machine registration has expired. Sending a authurl to register machine=test-machine
2022-03-08T17:08:45Z DBG Processing auth key for test-machine func=handleAuthKey machine=test-machine
2022-03-08T17:08:45Z ERR api.go:545 > Failed authentication via AuthKey error="AuthKey has already been used" func=handleAuthKey machine=test-machine
2022-03-08T17:08:45Z ERR api.go:568 > Failed authentication via AuthKey func=handleAuthKey machine=test-machine

2022/03/08 12:08:47 [Recovery] 2022/03/08 - 12:08:47 panic recovered:
POST /machine/221d92c201f65e402666072e3c223d73e1329603a0f8337350d6c9c4cc70fd60 HTTP/1.1
Host: 192.168.1.154:8080
Content-Length: 1296
User-Agent: Go-http-client/1.1


runtime error: invalid memory address or nil pointer dereference
```

mevansam:
This pull request fixes #489 with the following changes:

1. added function to return a valid unexpired machine given a machine key ignoring expired records for the same machine key
2. changed machine by key lookup to use the above function
3. added logic to also skip expired machines when determining unused IPs.


poll.go
```diff
@@ -60,7 +60,7 @@ func (h *Headscale) PollNetMapHandler(ctx *gin.Context) {
 		return
 	}
 
-	machine, err := h.GetMachineByMachineKey(machineKey)
+	machine, err := h.GetValidMachineByMachineKey(machineKey)
```
kradalby, line 63: I think that instead of filtering out the machine, we should check `.isExpired` here and return an error that describes the issue.

If we filter it out, we will return a "Record not found" which is not correct, the record is found, but it has expired. 

If think an if after all the error checking ,and then returning if it has expired makes the most sense.

utils.go
```diff
@@ -194,7 +198,9 @@ func (h *Headscale) getUsedIPs() (*netaddr.IPSet, error) {
 	// but this was quick to get running and it should be enough
 	// to begin experimenting with a dual stack tailnet.
 	var addressesSlices []string
-	h.db.Model(&Machine{}).Pluck("ip_addresses", &addressesSlices)
+	h.db.Model(&Machine{}).
```
kradalby, line 201: I dont think we safely can assume that we can hand out IPs that belong to expired nodes as of now, I think that requires a bit more thinking. 

I think this needs to be reverted, and a separate issue needs to be raised for discussion.

kradalby, line 201: I dont think we safely can assume that we can hand out IPs that belong to expired nodes as of now, I think that requires a bit more thinking. 

I think this needs to be reverted, and a separate issue needs to be raised for discussion.

kradalby (commented):
Good find, thanks, I think we need to approach it a bit differently, please see the comments. 

Feel free to discuss.

poll.go
mevansam: The issue is that logout and re-auth results in 2 records for the same machine. One expired and the other not. GetMachineByKey() picks the first machine record from the query result which happens to be the expired one. Another approach would be to refresh the expiration of the expired record when re-authenticating rather than creating a new one. This was how it was handled before the code was refactored.

utils.go
mevansam: I agree. This is probably not ideal. I added this as if we keep around expired records for the same machine then we will eventually run out of IPs. If we instead fix the handleAuthKey() to refresh the machine recorded if expired instead of inserting a new one this is not needed.

mevansam:
Let me know your thoughts on updating handleAuthKey to check if the machine being authenticated has an expired record and simply reset the record instead of adding a new one.

mevansam:
Perhaps a better solution may be to add refresh logic similar to line 224-254 in oidc.go to the handleAuthKey() function.

kradalby:
Yes, you have definitively identified a regression, it was never suppose to register a new record for expired machines, so let's fix so the reauth sets a new expiry

PR #495, opened 2022-03-10
Regression bug fix when re-authenticating machine with auth-key

mevansam:
This pull request fixes a regression bug where re-authenticating an expired machine (as a result of logout) resulted in a duplicate machine record. The fix simply refreshes the expired machine record to make it current.

This request updates https://github.com/juanfont/headscale/pull/490 based on review.


kradalby (approved):
Thanks, good catch

PR #498, opened 2022-03-13
Machine rename support

bravechamp:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Changes:

- Machines can have nicknames
- If nickname is set, it overrides Machine Name and new name will be used instead
- Added nickname on `nodes list` output
- New command `nodes rename` sets and clears nickname


restanrm (commented):
I think you are missing some lock mechanism to forbid update of the name from the machine like the `enable autogenerate machine name` from https://tailscale.com/kb/1098/machine-names/.
If you don't bring something like that, it would overwritten quickly I think.

bravechamp:
Thanks for the feedback. But, that's the point of the nickname field. It's a new field and can only be set by cli (or api). When set, it overrides machine name. When cleared, machine name is used. So, `enable autogenerate machine name` checkbox functionality is there.

bravechamp:
Well, `db.AutoMigrate(&Machine{})` already does that, right? During my tests, column appeared as soon as I run the executable.

cmd/headscale/cli/nodes.go
```diff
@@ -40,6 +40,13 @@ func init() {
 	}
 	nodeCmd.AddCommand(expireNodeCmd)
 
+	renameNodeCmd.Flags().Uint64P("identifier", "i", 0, "Node identifier (ID)")
```
kradalby, line 43: general thought: Is this a rename, or is it a setname or something?

cmd/headscale/cli/nodes.go
```diff
@@ -207,6 +214,54 @@ var expireNodeCmd = &cobra.Command{
 	},
 }
 
+var renameNodeCmd = &cobra.Command{
```
kradalby, line 217: These command will need [integration tests](https://github.com/juanfont/headscale/blob/main/integration_cli_test.go#L528)

machine.go
```diff
@@ -367,6 +368,29 @@ func (h *Headscale) ExpireMachine(machine *Machine) {
 	h.db.Save(machine)
 }
 
+// RenameMachine takes a Machine struct and sets the expire field to now.
```
kradalby, line 371: This comment seems wrong ;)

machine.go
```diff
@@ -40,6 +40,7 @@ type Machine struct {
 	DiscoKey    string
 	IPAddresses MachineAddresses
 	Name        string
+	Nickname    string
```
kradalby, line 43: I am a bit conflicted about the naming here; `Nickname` doesnt just sound completely right, in a way, what is now called `Name` should become `HostName` and then `Nickname` should be `Name` or maybe `GivenName`

machine.go
```diff
@@ -367,6 +368,29 @@ func (h *Headscale) ExpireMachine(machine *Machine) {
 	h.db.Save(machine)
 }
 
+// RenameMachine takes a Machine struct and sets the expire field to now.
+func (h *Headscale) RenameMachine(machine *Machine, newName string) {
+	newNickname, err := NormalizeToFQDNRules(
+		newName,
+		h.cfg.OIDC.StripEmaildomain,
+	)
```
kradalby, line 376: Since this is a user set name, we should not modify the value, we should throw a "name is not valid error" instead of setting something the user didnt expect.

machine.go
```diff
@@ -554,11 +578,15 @@ func (machine Machine) toNode(
 		keyExpiry = time.Time{}
 	}
 
+	name := machine.Name
+	if len(machine.Nickname) > 0 {
+		name = machine.Nickname
+	}
```
kradalby, line 584: I think we need to think about this a bit differently, if we implement this feature, we will have two name fields:

- `Name`
- `Nickname`

`Name` should be the _exact_ hostname that is passed by the Tailscale client, and should _not_ be used in the DNS config.

`Nickname` should be set as the Normalized name of `Name` on registration (e.g. it will always have a value) and then it _can_ be changed by the user later.

`Nickname` should therefore always be used for the DNS Config.

cmd/headscale/cli/nodes.go
bravechamp: I just reused `headscale namespaces rename` command to keep things familiar. I'm open for suggestions. Right now you can `headscale nodes -i 1 rename newmachinename` to set the `Nickname` field or `headscale nodes -i 1 rename` to clear it.

machine.go
bravechamp: I agree `Nickname` is not a good name. I didn't want to change the name of an existing field. Shall I change the `Name` to `Hostname` and keep the db column name as `name`?

machine.go
bravechamp: Oops. Result of pure laziness. Fixing...

cmd/headscale/cli/nodes.go
bravechamp: Will do

machine.go
bravechamp: OK, sounds good. I'll do that

machine.go
bravechamp: Then, we have to store the state of "[Auto-generate from OS hostname](https://tailscale.com/kb/1098/machine-names/#renaming-a-machine)" checkbox in database, right? Otherwise, each client update will overwrite this field.

cmd/headscale/cli/nodes.go
kradalby: Ye, its mostly fine, I didnt have a _really_ good name

machine.go
kradalby: I am slightly concerned that the migration software actually understands this tho... I would like to have that sorted. I personally am not interested in having naming things we are not happy with accumulate. 

I suspect that using something like `RenameColumn` (https://gorm.io/docs/migration.html) would just end in an "infinite" loop, renaming it every startup

I think we can do:
```suggestion
	GivenName    string
```

and `HostName`, and add a comment above both with how they are set, and what they mean.

machine.go
kradalby: Yes, so the flow should be:

1. Machine sends a register request with hostname: TailScaleNode-123
2. We set the `HostName` (currently `Name`) field to TailScaleNode-123
3. We Normalize TailScaleNode-123 and set the result to `GivenName` (currently `Nickname`)

Then if a user wants to _rename_ the host, we let them do that and use the validation part of the Normalization infra.

kradalby: It does look like the `GivenName` should be unique?

restanrm: Yes for the DNS to work properly, normalization function should check if GivenName is already registered in DB and add something to it. This step should behave proerly automatically without the use of an admin or the user to change anything. If the user is not satisfied with the name he could change it manually on the tailscale node.

kradalby:
Closing this in favour of #560

PR #499, opened 2022-03-15
Make STUN run by default when embedded DERP is enabled

juanfont:
This commit also allows to set an external STUN server, while running the embedded DERP server (without embedded STUN)

### DIFF
--- a/app.go
+++ b/app.go
@@ -48,6 +48,14 @@ import (
 )
 
 const (
+	errSTUNAddressNotSet                   = Error("STUN address not set")
+	errUnsupportedDatabase                 = Error("unsupported DB")
+	errUnsupportedLetsEncryptChallengeType = Error(
+		"unknown value for Lets Encrypt challenge type",
+	)
+)
+
+const (
 	AuthPrefix         = "Bearer "
 	Postgres           = "postgres"
 	Sqlite             = "sqlite3"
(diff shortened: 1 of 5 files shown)


app.go
```diff
@@ -62,6 +62,7 @@ const (
 	errUnsupportedLetsEncryptChallengeType = Error(
 		"unknown value for Lets Encrypt challenge type",
 	)
+	errSTUNAddressNotSet = Error("STUN address not set")
```
kradalby, line 65: Define this at the top as a `const`

app.go
```diff
@@ -500,6 +501,10 @@ func (h *Headscale) Serve() error {
 	h.DERPMap = GetDERPMap(h.cfg.DERP)
 
 	if h.cfg.DERP.ServerEnabled {
+		if h.cfg.DERP.STUNAddr == "" { // When embedded DERP is enabled we always need a STUN server address, embedded or external
```
kradalby, line 504:
```suggestion
		// When embedded DERP is enabled we always need a STUN server address, embedded or external
		if h.cfg.DERP.STUNAddr == "" {
```

config-example.yaml
```diff
@@ -69,10 +69,13 @@ derp:
     region_code: "headscale"
     region_name: "Headscale Embedded DERP"
 
-    # If enabled, also listens in UDP at the configured address for STUN connections to help on NAT traversal
+    # Enabled by default when embedded DERP is enabled. Listens in UDP at the configured address for STUN connections
+    # to help on NAT traversal.
+    # If DERP is enabled, but STUN is disabled you still need to input an external STUN server in the listen_addr field.
```
kradalby, line 74: hmm, I am not sure about that name/config option being reused for the server if it is disabled, that sounds confusing :/...

app.go
juanfont: ? Is a const at the top

config-example.yaml
kradalby: how about `derp.override_stun_server`

kradalby (approved):

kradalby (approved):

PR #521, opened 2022-03-22
Docker docs enhancements

Niek:
While configuring a Docker setup I noticed that the docs could use some enhancements.

### DIFF
--- a/docs/running-headscale-container.md
+++ b/docs/running-headscale-container.md
@@ -45,6 +45,7 @@ touch ./config/config.yaml
 ```
 
 Modify the config file to your preferences before launching Docker container. Here are some settings that you likely want:
+
 ```yaml
 server_url: http://your-host-name:8080 # Change to your hostname or host IP
 metrics_listen_addr: 0.0.0.0:9090 # Listen to 0.0.0.0 so it's accessible outside the container
--- a/docs/running-headscale-container.md
+++ b/docs/running-headscale-container.md
@@ -100,7 +100,8 @@ curl http://127.0.0.1:9090/metrics
 6. Create a namespace ([tailnet](https://tailscale.com/kb/1136/tailnet/)):
 
 ```shell
-docker exec headscale headscale namespaces create myfirstnamespace
+docker exec headscale \
+  headscale namespaces create myfirstnamespace
 ```
 
 ### Register a machine (normal login)
--- a/docs/running-headscale-container.md
+++ b/docs/running-headscale-container.md
@@ -49,9 +49,12 @@ Here are some settings that you likely want:
 
 ```yaml
 server_url: http://your-host-name:8080 # Change to your hostname or host IP
-metrics_listen_addr: 0.0.0.0:9090 # Listen to 0.0.0.0 so it's accessible outside the container
-private_key_path: /etc/headscale/private.key # The default /var/lib/headscale path is not writable in the container
-db_path: /etc/headscale/db.sqlite # The default /var/lib/headscale path is not writable  in the container
+# Listen to 0.0.0.0 so it's accessible outside the container
+metrics_listen_addr: 0.0.0.0:9090
+# The default /var/lib/headscale path is not writable in the container
+private_key_path: /etc/headscale/private.key
+# The default /var/lib/headscale path is not writable  in the container
+db_path: /etc/headscale/db.sqlite
 ```
 
 4. Start the headscale server while working in the host headscale directory:
--- a/docs/running-headscale-container.md
+++ b/docs/running-headscale-container.md
@@ -65,12 +65,14 @@ docker run \
   --detach \
   --rm \
   --volume $(pwd)/config:/etc/headscale/ \
-  --publish 0.0.0.0:8080:8080 \
+  --publish 127.0.0.1:8080:8080 \
   --publish 127.0.0.1:9090:9090 \
   headscale/headscale:<VERSION> \
   headscale serve
 
 ```
+
+Note: use `0.0.0.0:8080:8080` instead of `127.0.0.1:8080:8080` if you want to expose the container externally.
 
 This command will mount `config/` under `/etc/headscale`, forward port 8080 out of the container so the
 `headscale` instance becomes available and then detach so headscale runs in the background.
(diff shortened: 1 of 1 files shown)


juanfont (dismissed):
Looks good!

ohdearaugustin:
Looks good.

docs/running-headscale-container.md
```diff
@@ -101,7 +107,7 @@ tailscale up --login-server YOUR_HEADSCALE_URL
 To register a machine when running `headscale` in a container, take the headscale command and pass it to the container:
 
 ```shell
-docker exec headscale -- \
```
kradalby, line 104: Why do you remove these?

docs/running-headscale-container.md
```diff
@@ -44,7 +44,13 @@ curl https://raw.githubusercontent.com/juanfont/headscale/main/config-example.ya
 touch ./config/config.yaml
 ```
 
-Modify the config file to your preferences before launching Docker container.
+Modify the config file to your preferences before launching Docker container. Here are some settings that you likely want:
+```yaml
+server_url: http://your-host-name:8080 # Change to your hostname or host IP
+metrics_listen_addr: 0.0.0.0:9090 # Listen to 0.0.0.0 so it's accessible outside the container
+private_key_path: /etc/headscale/private.key # The default /var/lib/headscale path is not writable in the container
+db_path: /etc/headscale/db.sqlite # The default /var/lib/headscale path is not writable  in the container
```
kradalby, line 52: Should we not rather use a docker volume here?

/etc sounds like the wrong place

docs/running-headscale-container.md
```diff
@@ -44,7 +44,13 @@ curl https://raw.githubusercontent.com/juanfont/headscale/main/config-example.ya
 touch ./config/config.yaml
 ```
 
-Modify the config file to your preferences before launching Docker container.
+Modify the config file to your preferences before launching Docker container. Here are some settings that you likely want:
+```yaml
+server_url: http://your-host-name:8080 # Change to your hostname or host IP
```
kradalby, line 49: Can you put the comments above so the line does not get that long ?

kradalby (changes requested):

docs/running-headscale-container.md
Niek: `/etc/headscale` is the docker volume - or do you want a separate volume for the key and db? That sounds a bit overkill to me.

docs/running-headscale-container.md
Niek: Fix in [73f1c06](https://github.com/juanfont/headscale/pull/521/commits/73f1c06f65e1b010da2bf37044d100f2f2af52c5)

docs/running-headscale-container.md
Niek: Because (at least in docker 20.x) this doesn't work:
```bash
$ docker exec headscale -- headscale version
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "--": executable file not found in $PATH: unknown
$ docker exec headscale headscale version
dev
```

docs/running-headscale-container.md
ohdearaugustin: I also in practice would rather go with the approach of only one volume.

docs/running-headscale-container.md
ohdearaugustin: those "--" also seem a bit fishy to me.

docs/running-headscale-container.md
```diff
@@ -54,7 +62,7 @@ docker run \
   --detach \
   --rm \
   --volume $(pwd)/config:/etc/headscale/ \
-  --publish 127.0.0.1:8080:8080 \
+  --publish 0.0.0.0:8080:8080 \
```
ohdearaugustin, line 65: This was intentionally left on 127.0.0.1. So people won't publish this service by default and have to think before the do. So rather on the approach keep it save until you really want to publish it.

docs/running-headscale-container.md
kradalby: I want to separate variable data (`/var`, so database and key that changes over time) from the configuration (`/etc`)

So one way is to mount a data folder into `/var/lib/headscale` or do create a docker volume and mount it to the same folder.

docs/running-headscale-container.md
kradalby: Double dashes is a _really_ good convention in shell programs, [good elaboration here](https://www.baeldung.com/linux/double-dash-in-shell-commands).

But you are right, I was certain I had tested it with `docker` but it does not seem to work. Sad, why does docker not conform to standards.

ohdearaugustin: Really nice explanation, yeah docker is "special" in many different ways..

docs/running-headscale-container.md
ohdearaugustin: Maybe just mention it in the documentation. That either way is possible.

docs/running-headscale-container.md
Niek: Fixed in [3272feb](https://github.com/juanfont/headscale/pull/521/commits/3272febfb3e42081d14b620b0863f3c6defa7440)

docs/running-headscale-container.md
Niek: @kradalby The current docs mention creating the `db.sqlite` in the config directory/volume. The only thing I fixed was the incorrect paths in the config file.
If you want to make a separate `data` directory mapped to `/var/lib/headscale` you also need to change the first steps of the doc.

docs/running-headscale-container.md
```diff
@@ -87,7 +100,7 @@ curl http://127.0.0.1:9090/metrics
 6. Create a namespace ([tailnet](https://tailscale.com/kb/1136/tailnet/)):
 
 ```shell
-docker exec headscale -- headscale namespaces create myfirstnamespace
+docker exec headscale headscale namespaces create myfirstnamespace
```
kradalby, line 103:
```suggestion
docker exec headscale \
  headscale namespaces create myfirstnamespace
```

Lets do this so the double headscale dont look so confusing

kradalby (approved):

PR #482, opened 2022-03-07
Add Nix reproducible build system

kradalby:
This pr adds a flake.nix build file, it can be used for three
things:

Build `headscale` from local or straight from git:

```shell
nix build
or
nix build github:juanfont/headscale
```

Run and Build `headscale` from local or straight from git:

```shell
nix run
or
nix run github:juanfont/headscale
```

Set up a development environment including all our tools,
- linters
- protobuf tooling
- compilers

```shell
nix develop
```

Docker image creation:

```
nix build .#headscale-docker
docker load < result
```

PR #557, opened 2022-04-21
remove necessary buf installation

mpldr:
This commit adds buf as a regular dependency and go running it instead
of requiring installing buf to one's own GOBIN.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

This also has the added advantage that every user is always using the same version.


kradalby:
I was slightly mixed on this, but changed my mind when I realise this is what I set up for the actual gRPC plugins.

Unless working on API stuff, it isnt needed for building, but might as well have it available.

I am happy to add this now, but I will consider it obsoleted by #482, which defines the whole build env in a stricter manner.

kradalby (approved):

Issue #561, opened 2022-04-25
OpenID Connect: scopes, custom parameters and domain filter


apognu:
I had a quick look at the config and code for the OpenID Connect integration, and right now, as far as I can tell, the _scopes_ are hardcoded. It should be possible to customize them to fit one's deployment.

Also, some OpenID Connect implementation (namely, Google Auth) allows any Google Account to connect even if the application was created within a Google Workspace namespace. Only allowing account in specific domains would make sense here for enterprise deployments of Headscale (or even individuals with their own domain).

Lastly, we could allow custom parameters to be added to the request in the Authorize Endpoint of the flow. Some providers provide functionnality behind those _custom_ query parameters, for instance:
 * `hd` on Google Auth or `domain_hint` on Azure AD
 * `tenant` on Azure AD

I could work on a PR if those features are of interest.

apognu:
FYI, I have something ready if this is wanted.

apognu:
Submitted a PR with what was described here, with user allowlisting added on top of domain allowlisting (which can be useful with Google Auth without Workspace, for instance).

apognu:
This was merged in #565.

PR #565, opened 2022-04-30
Allow more configuration over the OIDC flow

apognu:
Adds knobs to configure three aspects of the OpenID Connect flow:

 * Custom scopes to override the default "openid profile email".
 * Custom parameters to be added to the Authorize Endpoint request.
 * Domain allowlisting for authenticated principals.
 * User allowlisting for authenticated principals.

- [x] raised a GitHub issue or discussed it on the projects chat beforehand (#561)
- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/app.go
+++ b/app.go
@@ -119,6 +119,10 @@ type OIDCConfig struct {
 	Issuer           string
 	ClientID         string
 	ClientSecret     string
+	Scope            []string
+	ExtraParams      map[string]string
+	AllowedDomains   []string
+	AllowedUsers     []string
 	StripEmaildomain bool
 }
 
--- a/oidc.go
+++ b/oidc.go
@@ -53,7 +53,7 @@ func (h *Headscale) initOIDC() error {
 				"%s/oidc/callback",
 				strings.TrimSuffix(h.cfg.ServerURL, "/"),
 			),
-			Scopes: []string{oidc.ScopeOpenID, "profile", "email"},
+			Scopes: h.cfg.OIDC.Scope,
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


apognu:
Linting didn't pass because other files (not changed by this MR) were improperly formatted and I didn't want to touch them since they were out of scope for these specific changes. I can rebase once upstream is fixed.

oidc.go
```diff
@@ -45,6 +45,11 @@ func (h *Headscale) initOIDC() error {
 			return err
 		}
 
+		scope := h.cfg.OIDC.Scope
+		if len(scope) == 0 {
+			scope = []string{oidc.ScopeOpenID, "profile", "email"}
+		}
```
kradalby, line 51: Maybe it would be nicer to set this as a default instead? That way we have all "default" logic in the config part and not special logic polluting the code.

oidc.go
```diff
@@ -187,6 +199,46 @@ func (h *Headscale) OIDCCallback(ctx *gin.Context) {
...
+	if len(h.cfg.OIDC.AllowedDomains) > 0 {
+		domainFound := false
+
+		for _, domain := range h.cfg.OIDC.AllowedDomains {
+			if strings.HasSuffix(claims.Email, fmt.Sprintf("@%s", domain)) {
+				domainFound = true
+
+				break
+			}
+		}
```
kradalby, line 212: I feel like this can be rewritten to something more generic that strips the claim to the hostname and use a generic function for String in list? (This would also apply to the user bellow).

I am certain we have a couple of those spread around in the code, at least one.

oidc.go
```diff
@@ -187,6 +199,46 @@ func (h *Headscale) OIDCCallback(ctx *gin.Context) {
...
+
+			return
+		}
+	}
+
+	// If AllowedUsers is provided, check that the authenticated princial is part of that list.
+	if len(h.cfg.OIDC.AllowedUsers) > 0 {
+		userFound := false
+
+		for _, user := range h.cfg.OIDC.AllowedUsers {
```
kradalby, line 226: Same comment as for Domain

app.go
```diff
@@ -119,6 +119,10 @@ type OIDCConfig struct {
 	Issuer           string
 	ClientID         string
 	ClientSecret     string
+	Scope            []string
+	ExtraParams      map[string]string
+	AllowedDomains   []string
+	AllowedUsers     []string
```
kradalby, line 125: This question is rooted in my lack of knowledge of naming conventions for OIDC, is this the appropriate naming? (I dont doubt that it is, but maybe a link to the spec would be nice)

kradalby (commented):
Looks very reasonable, couple of questions :)

kradalby (approved):

PR #566, opened 2022-04-30
Do not setLastStateChangeToNow every expireEphemeralNodesWorker call

juanfont:
`setLastStateChangeToNow` was being called everytime we run `expireEphemeralNodesWorker`, even if there are no nodes to expire.

- [x] updated CHANGELOG.md


kradalby (approved):

kradalby:
Oh lol yes, that explains a lot 🙈

PR #583, opened 2022-05-12
add command argument --show-routes to node list.

deonthomasgy:
It's very hard to find routes using `headscale route`, there's also nowhere to list all nodes and routes.

`headscale node list --show-routes` 

image

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


kradalby:
Hi @deonthomasgy 

I think these prams looks quite sensible, but I think I would like a bit more generic interface. I have not had time to think about it yet, but maybe so you can turn on more at once 🤔

Not sure how it would look like, but something that can turn on and off columns:


```
--columns tags,routes,name,ip 
```

deonthomasgy:
> Hi @deonthomasgy
> 
> I think these prams looks quite sensible, but I think I would like a bit more generic interface. I have not had time to think about it yet, but maybe so you can turn on more at once thinking
> 
> Not sure how it would look like, but something that can turn on and off columns:
> 
> ```
> --columns tags,routes,name,ip 
> ```

hmm... I see where you are going, so maybe the default is what headscale prints right now and you can add
`--with-columns=tags,routes` ... so in future it makes it easier to add more. what do you think?

kradalby:
Yes, something like that, I agree that we should just make it modular, and then the default comma separated list is all the current columns.

That should make it easy to add more in the future, and we can make "presets" line "wide" which would display a larger preset.

Maybe using the same flags  as kubectl makes sense to keep things consistent with widely used tools.

restanrm (commented):
Maybe you should split this in 2 PR's, on to add the show route command and one to improve the options of the list command

juanfont:
Going to close this for the time being. Really needs to be split + fix the conflicts.

Issue #173, opened 2021-10-16
Add reload and test config commands


0xcd0e:
Like the `nginx -t` command, it would be nice to verify the config (config.json, derp map, acl) prior to loading it.
Also, like `nginx -s reload`, it would be nice to reload the configuration during runtime without having to restart the service.

juanfont:
I see the point of this, but I reckon we should develop first an internal API to communicate the CLI with the running service.

Also, bear in mind that given how Tailscale works, restarting the service does not break the connectivity of the nodes.

juanfont:
You mean the current AMAZING state of the project 🤣

kradalby:
We should handle signals, so we can do what "traditional" unix apps does and reload the configfile, acls etc on `HUP` and do graceful shutdown on `INT`, `KILL`, etc.

kradalby:
Adding this support for ACLs in #601, in addition to a bit more graceful signal handling for shutdown.

Config reload will have to sit for a bit more, I think we need to do a few rounds of cleanups in the config handling first, to decouple it so this is possible.

PR #601, opened 2022-05-31
Allow ACLs to be reloaded with SIGHUP

kradalby:
This PR partially resolves https://github.com/juanfont/headscale/issues/173, allowing the users to send a reload signal (SIGHUP) to headscale to reload the ACL policies.

In addition, sending a SIGTERM, SIGINT etc, will now attempt to shutdown headscale gracefully.

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Add option to enable/disable logtail (Tailscale's logging infrastructure) [#596](https://github.com/juanfont/headscale/pull/596)
   - This change disables the logs by default
 - Use [Prometheus]'s duration parser, supporting days (`d`), weeks (`w`) and years (`y`) [#598](https://github.com/juanfont/headscale/pull/598)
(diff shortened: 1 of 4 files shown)


juanfont (approved):

GoodiesHQ:
Given that this is the case, is there any possible issue with setting a cronjob for running `docker kill --signal HUP headscale` every 1-5 minutes or so? I have tested with running it manually and it seems to be very quick (granted, I only have a very small ACL file, so maybe the time taken scales with complexity). Using the Headscale-UI program to assign tags to nodes is great, but it doesn't seem to take effect until it is reloaded. Since there is currently no API support for ACLs, you can't add users to groups with it, either.

kradalby:
@GoodiesHQ I would assume that is fine, how performant/problematic is hard to predict, but there shouldnt be anything wrong with it.

PR #602, opened 2022-06-01
Fix forced Tags with legitimate tagOwners

iSchluff:
Basically always check whether a tag is forced instead of just on tagOwners failure. Also replace loops with contains.

- [x] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/acls.go
+++ b/acls.go
@@ -250,16 +250,17 @@ func expandAlias(
 	}
 
 	if strings.HasPrefix(alias, "tag:") {
+		// check for forced tags
+		for _, machine := range machines {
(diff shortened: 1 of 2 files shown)


iSchluff:
Just noticed that without this PR forcedTags can't be used at all because to have a valid acl containing tags they need to be listed under tagOwners.

Maybe with forcedTags that requirement could be relaxed?

restanrm (commented):
Good catch, thank you !

kradalby (approved):

Issue #617, opened 2022-06-08
Support new ACL format

In february tailscale added `src` and `dst` field for ease of use on `users` and `ports` respectively


restanrm:
We should support both and present example with `src` and `dst` as it look like to be the default in the future

Issue #572, opened 2022-05-03
ACLs don't work with host aliases or wildcard sources


madjam002:
**Bug description**

When using host aliases as a source/destination in ACL policies, or wildcards as the destination, ACL policies don't work and nodes will be unable to communicate with eachother.

My best guess is that this is because the peers aren't being calculated correctly when sending the peer list to each Tailscale client, as source nodes don't appear in the peer list when querying `tailscale status` on destination nodes (perhaps an issue in `getFilteredByACLPeers`).

**To Reproduce**

Given two nodes, `node1` and `node2` and the following ACL policy (not actually tested but is theoretical), and let's say `node1` is authenticated with a user which is part of `mygroup` (excluded in the example below for brevity):

```
{
  "hosts": {
    "node1": "100.64.0.1",
    "node2": "100.64.0.2"
  },
  "acls": [{
    "action": "accept",
    "src": ["group:mygroup"],
    "dst": ["node2:22"]
  }]
}
```

Communication will not be possible from node1 -> node2 over port 22. `tailscale status` on `node2` won't show `node1` as a peer.

Changing the destination to the IP address itself:
```
    "dst": ["100.64.0.2:22"]
```
and communication will now be possible. `tailscale status` on `node2` will show `node1` as a peer. The issue exists in reverse as well, so if a host alias is used in the `src`, the destination node won't have any valid source nodes as a peer, whereas using the IP address directly works fine.

The same issue exists if the destination is set to `*:22` as a wildcard, whereas e.g `somenamespace:22` will work fine.

On a separate note, given `node1` and `node3`, where `node3` is advertised as an exit node, I've noticed that `node3` does not appear in the peer list of `node1` even if an ACL permits routes over the public internet, unless an ACL rule exists explicitly allowing access to a port of `node3`, e.g `{"action":"accept","src":["100.64.0.1"],"dst":["100.64.0.2:1111"]}` (some bogus port we don't actually use). I think this is a separate issue although it is similar in nature, and I'm not sure if this is expected behaviour compared to how Tailscale's official control plane operates. (actually this subnet router issue looks like it is covered by issue #502)

**Context info**

- Tested with headscale commit 3fbfc5a6492d1d60600508efce1ca8d8d1eb3331
- Tailscale v1.22.2
- OS Linux NixOS 22.05
- Kernel version 5.15.34

PR #618, opened 2022-06-08
Update ACL syntax and add support for protocol filtering

juanfont:
Implements #617.

Tailscale has changed the format of their ACLs to use a more firewall-y terms ("users" & "ports" -> "src" & "dst"). They have also started using all-lowercase tags. 

Also, protocol filtering support :)

This PR implements these changes.

### DIFF
--- a/acls.go
+++ b/acls.go
@@ -23,6 +23,7 @@ const (
 	errInvalidGroup      = Error("invalid group")
 	errInvalidTag        = Error("invalid tag")
 	errInvalidPortFormat = Error("invalid port format")
+	errWildcardIsNeeded  = Error("wildcard as port is required for the procotol")
 )
 
 const (
@@ -134,9 +135,17 @@ func (h *Headscale) generateACLRules() ([]tailcfg.FilterRule, error) {
 			srcIPs = append(srcIPs, srcs...)
 		}
 
+		protocols, needsWildcard, err := parseProtocol(acl.Protocol)
+		if err != nil {
+			log.Error().
+				Msgf("Error parsing ACL %d. protocol unknown %s", index, acl.Protocol)
+
+			return nil, err
+		}
+
 		destPorts := []tailcfg.NetPortRange{}
 		for innerIndex, dest := range acl.Destinations {
-			dests, err := h.generateACLPolicyDest(machines, *h.aclPolicy, dest)
+			dests, err := h.generateACLPolicyDest(machines, *h.aclPolicy, dest, needsWildcard)
 			if err != nil {
 				log.Error().
 					Msgf("Error parsing ACL %d, Destination %d", index, innerIndex)
@@ -149,6 +158,7 @@ func (h *Headscale) generateACLRules() ([]tailcfg.FilterRule, error) {
 		rules = append(rules, tailcfg.FilterRule{
 			SrcIPs:   srcIPs,
 			DstPorts: destPorts,
+			IPProto:  protocols,
 		})
 	}
 
@@ -167,6 +177,7 @@ func (h *Headscale) generateACLPolicyDest(
 	machines []Machine,
 	aclPolicy ACLPolicy,
 	dest string,
+	needsWildcard bool,
 ) ([]tailcfg.NetPortRange, error) {
 	tokens := strings.Split(dest, ":")
 	if len(tokens) < expectedTokenItems || len(tokens) > 3 {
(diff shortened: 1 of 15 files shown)


juanfont:
Also addresses #572

gnosisfarm:
As Hack said, looks great on TCP connections. Looks like UDP is broken with any ACL applied. UDP traffic does not seem return.
I tried a few different rules with no protocol mentioned and with explicit proto = udp. I am not able to pass a DNS request from.   Does UDP work for you? 

Packet capture on the DNS server with an ACL policy in place shows the UDP packet not getting a proper response. 

08:12:35.471876 IP (tos 0x0, ttl 64, id 6918, offset 0, flags [none], proto UDP (17), length 64)
    100.64.0.1.41054 > 100.64.0.4.53: 33344+ A? dev.corp.example.com. (36)
08:12:35.471916 IP (tos 0xc0, ttl 64, id 15612, offset 0, flags [none], proto ICMP (1), length 92)
    100.64.0.4 > 100.64.0.1: ICMP 100.64.0.4 udp port 53 unreachable, length 72

acls_test.go
```diff
@@ -143,7 +143,7 @@ func (s *Suite) TestValidExpandTagOwnersInUsers(c *check.C) {
 
 // this test should validate that we can expand a group in a TagOWner section and
 // match properly the IP's of the related hosts. The owner is valid and the tag is also valid.
-// the tag is matched in the Ports section.
+// the tag is matched in the Destinations section.
 func (s *Suite) TestValidExpandTagOwnersInPorts(c *check.C) {
```
restanrm, line 147: Shouldn't you rename this as well ?

acls_test.go
```diff
@@ -628,7 +642,8 @@ func Test_expandTagOwners(t *testing.T) {
 
 func Test_expandPorts(t *testing.T) {
```
restanrm, line 643: Maybe it should be renamed too

restanrm (approved):
It looks good !

acls.go
```diff
@@ -23,6 +23,7 @@ const (
 	errInvalidGroup      = Error("invalid group")
 	errInvalidTag        = Error("invalid tag")
 	errInvalidPortFormat = Error("invalid port format")
+	errWildcardIsNeeded  = Error("wildcard as port is required for the procotol")
```
kradalby: spelling error

acls.go
```diff
@@ -214,6 +225,54 @@ func (h *Headscale) generateACLPolicyDest(
...
+// protocols that will be allowed, following the IANA IP protocol number
+// https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
+//
+// If the ACL proto field is empty, it allows ICMPv4, ICMPv6, TCP, and UDP,
+// as per Tailscale behaviour (see tailcfg.FilterRule).
+//
+// Also returns a boolean indicating if the protocol
+// requires all the destinations to use wildcard as port number (only TCP,
+// UDP and SCTP support specifying ports).
+func parseProtocol(protocol string) ([]int, bool, error) {
```
kradalby, line 238: All of these numbers will trigger "magicnumber" when the linter is up and running, and in any case, it would be better to  probably have constants for these

acls_test.go
```diff
@@ -321,6 +321,20 @@ func (s *Suite) TestPortRange(c *check.C) {
...
+func (s *Suite) TestProtocolParsing(c *check.C) {
+	err := app.LoadACLPolicy("./tests/acls/acl_policy_basic_protocols.hujson")
+	c.Assert(err, check.IsNil)
+
+	rules, err := app.generateACLRules()
+	c.Assert(err, check.IsNil)
+	c.Assert(rules, check.NotNil)
+
+	c.Assert(rules, check.HasLen, 3)
+	c.Assert(rules[0].IPProto[0], check.Equals, 6)  // tcp
```
kradalby, line 333: This is why we want constants.

acls.go
```diff
@@ -214,6 +225,54 @@ func (h *Headscale) generateACLPolicyDestPorts(
...
+		return []int{132}, false, nil
+	case "icmp":
+		return []int{1, 58}, true, nil
+
+	default:
+		protocolNumber, err := strconv.Atoi(protocol)
+		if err != nil {
+			return nil, false, err
+		}
+		needsWildcard := protocolNumber != 6 && protocolNumber != 17 && protocolNumber != 132 // nolint
```
kradalby, line 270:
```suggestion
		needsWildcard := protocolNumber != 6 && protocolNumber != 17 && protocolNumber != 132
```

That wont do ;)

acls_test.go
juanfont: No, that one actually expands ports :)

kradalby (approved):

PR #622, opened 2022-06-09
namespaces are tailnets

majst01:
Stumbled over this statement in the Readme, but from my understanding namespaces are the equivalent to tailnets. If i am wrong simply close this PR.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,7 @@ organisation.
 ## Design goal
 
 `headscale` aims to implement a self-hosted, open source alternative to the Tailscale
-control server. `headscale` has a narrower scope and an instance of `headscale`
-implements a _single_ Tailnet, which is typically what a single organisation, or
-home/personal setup would use.
+control server.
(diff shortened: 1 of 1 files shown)


kradalby:
Hi, they used to be half way tailnets. They are not anymore as we opened them up so we were closer to the "correct" implementation.

majst01:
> Hi, they used to be half way tailnets. They are not anymore as we opened them up so we were closer to the "correct" implementation.

OK, what is missing to reach that ?

kradalby:
> OK, what is missing to reach that ?

For headscale to support multiple tailnets? Quite a lot of re-engineering. 

The original issue was discussed here: https://github.com/juanfont/headscale/pull/311

We might add it back in the future, but it is not on any immediate roadmap and would require a lot of thinking.

PR #643, opened 2022-06-14
Fix occasional panic on registration

iSchluff:
Introduced by random node suffix: GenerateRandomStringDNSSafe will panic occasionally if the random base64 string contains too many - and _ due to the replacement. Fix by looping.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -325,11 +325,17 @@ func GenerateRandomStringURLSafe(n int) (string, error) {
 // number generator fails to function correctly, in which
 // case the caller should not continue.
 func GenerateRandomStringDNSSafe(n int) (string, error) {
-	str, err := GenerateRandomStringURLSafe(n)
+	var str string
+	var err error
+	for len(str) < n {
+		str, err = GenerateRandomStringURLSafe(n)
+		if err != nil {
+			return "", err
+		}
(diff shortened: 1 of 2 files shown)


utils.go
```diff
@@ -325,11 +325,20 @@ func GenerateRandomStringURLSafe(n int) (string, error) {
 // number generator fails to function correctly, in which
 // case the caller should not continue.
 func GenerateRandomStringDNSSafe(n int) (string, error) {
-	str, err := GenerateRandomStringURLSafe(n)
-
-	str = strings.ToLower(strings.ReplaceAll(strings.ReplaceAll(str, "_", ""), "-", ""))
+	var str string
+	var err error
+	for {
```
e-zk, line 330: This infinite for loop with the break on the if statement on line 336 here can be changed to a while-style for loop instead:

```go
for len(str) < n {
    str, err = GenerateRandomStringURLSafe(n)
    if err != nil {
        return "", err
    }
    str = strings.ToLower(strings.ReplaceAll(strings.ReplaceAll(str, "_", ""), "-", ""))
}
```

iSchluff: yep, updated

kradalby (approved):

Issue #526, opened 2022-03-26
Proposal: Implement TS2021 (Tailscale control protocol v2)


juanfont:
Tailscale clients communicate with the control server using Tailscale's control protocol. This is what basically Headscale implements.

It is based on a HTTP API with a bit of Long Polling, and a grain of NaCl (https://nacl.cr.yp.to/) to encrypt the JSON payloads. Since 2019 the protocol has remained mostly stable - with just some extra fields being added to support new functionality like MagicDNS or Taildrop. 

In our side the core of the implementation is located at `api.go` (where the registration methods are located) and `poll.go` (where lies the method that the clients use to receive updates).

----

A couple of weeks ago Tailscale team let us know (!) that they are working to implement the version 2 of the control protocol, codename TS2021. 

They have also been kind enough to a) share some internal documentation on the implementation (!!), and b) release code in https://github.com/tailscale/tailscale that helps us _A LOT_ with the implementation (!!!!!).

They did this in order not to break Headscale. Very very very big kudos to them!

## About TS2021

TS2021 is a Noise-based protocol (https://noiseprotocol.org/noise.html), using the IK pattern (https://noiseprotocol.org/noise.html#interactive-handshake-patterns-fundamental). It is the same cryptographic framework as the one used for Signal or Whatsapp.

We will not have to deal with Noise too much. For us, the very first step is a POST call to `/ts2021` and an upgrade + hijack of the TCP connection. Then the code I mentioned above quicks in, to create the Noise session. Once this is established, the API is reachable to the clients using what it looks like a H2C server (essentially just the good old v1 API, but without NaCl encryption for the payloads).

From what we can see, as of late March 2022 they have not yet fully migrated all the API methods to use TS2021. So we will have to follow them up gradually.

## Our steps

1. Prepare our API machinery (always wanted to use this word) to be able to deal with clients using TS2019 (no idea if they call it this way) and TS2021. This includes a minor change in the `/key` method, and removing NaCl for TS2021.

2. Implement the `/ts2021` handler (its quite similar to what we do for the embedded DERP server)

3. Plug a H2C server to the Noise connection under `/ts2021` to expose our current API.

4. Keep track of their `CurrentCapabilityVersion`, gradually enabling new API calls under TS2021

## Current status

I have a prototype _mostly_ working. I will clean it a little bit an prepare a draft PR for scrutiny.

danderson:
[...]

PR #645, opened 2022-06-15
Implement TS2021 

juanfont:
This PR adds full support for TS2021, the new control protocol of Tailscale. For reference, see https://github.com/juanfont/headscale/issues/526.

This is the second implementation I do of this, this time without using Gin - just plain old net/http. Gin was terminating the Noise connection or not flushing it.

T2021 prefers to run over http, to avoid double encryption. Currently, any port is supported for http. However, for https only tcp/443 works. https://github.com/tailscale/tailscale/pull/4323 is open in upstream to change that.

Tests are not working yet. Working on that. But It Should Work(tm)


juanfont:
I am going to close this one, and make a third attempt :)

PR #656, opened 2022-06-20
Drop Gin as web framework for TS2019 API

juanfont:
Until now, and since the very beginning of Headscale, we have used Gin (https://github.com/gin-gonic/gin) as the web framework to serve the original Tailscale API (TS2019, as opposed to Tailscale API v2 aka. TS2021).

It served (no pun intended) us well. And for many other use cases Gin is great. But while implementing TS2021 I found some weird situations where Gin was flushing or closing connections that where meant to be kept. 

This PR moves all the TS2019 endpoints to plain `net/http` handler functions. It also switches Gin's router to `gorilla/mux`.


app.go
```diff
@@ -326,48 +327,56 @@ func (h *Headscale) grpcAuthenticationInterceptor(ctx context.Context,
...
+		w http.ResponseWriter,
+		r *http.Request,
+	) {
+		log.Trace().
+			Caller().
+			Str("client_address", r.RemoteAddr).
+			Msg("HTTP authentication invoked")
 
-	authHeader := ctx.GetHeader("authorization")
+		authHeader := r.Header.Get("X-Session-Token")
```
juanfont, line 340: This is fixed on a later commit

kradalby:
Looks good;

- Gin is still in `go.mod`, is there a reason why it cant be removed completely? (`go mod tidy` ?)
- When changing `go.mod`, flake.nix must be updated with the new hash of modules (`nix build`)
- It looks like golangci-lint is working again ;)

kradalby (dismissed):

kradalby (approved):

06kellyjac:
`Drop Gin as web framework in Headscale` in the changelog links to the wrong PR instead of this one :)

Issue #652, opened 2022-06-17
Segfault when running ```headscale nodes list --output json``` when backend is not running.


theonemcdonald:
When running `headscale nodes list --output json` against a server that is offline, a segfault is encountered. `headscale nodes list` fails gracefully with `Could not connect....`. This segfault is problematic for frontends that consume json output exclusively.

```
$ headscale nodes list --output json
panic: runtime error: invalid memory address or nil pointer dereference
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xb1feab]

goroutine 1 [running]:
google.golang.org/grpc.(*ClientConn).Close(0x0)
        google.golang.org/grpc@v1.46.0/clientconn.go:998 +0x4b
panic({0x1124f80, 0x1c52bc0})
        runtime/panic.go:838 +0x207
google.golang.org/grpc.(*ClientConn).Invoke(0x80203cf18?, {0x1464490?, 0xc00054ec60?}, {0x1292a76?, 0x0?}, {0x118afc0?, 0xc00054c040?}, {0x118b080?, 0xc00054c180?}, {0x0, ...})
        google.golang.org/grpc@v1.46.0/call.go:32 +0x5e
github.com/juanfont/headscale/gen/go/headscale/v1.(*headscaleServiceClient).ListMachines(0xc00007a780, {0x1464490, 0xc00054ec60}, 0xc0002f4420?, {0x0, 0x0, 0x0})
        github.com/juanfont/headscale/gen/go/headscale/v1/headscale_grpc.pb.go:195 +0xce
github.com/juanfont/headscale/cmd/headscale/cli.glob..func16(0x1c666e0?, {0x125c23c?, 0x2?, 0x2?})
        github.com/juanfont/headscale/cmd/headscale/cli/nodes.go:168 +0x2d0
github.com/spf13/cobra.(*Command).execute(0x1c666e0, {0xc000550980, 0x2, 0x2})
        github.com/spf13/cobra@v1.4.0/command.go:860 +0x663
github.com/spf13/cobra.(*Command).ExecuteC(0x1c68260)
        github.com/spf13/cobra@v1.4.0/command.go:974 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.4.0/command.go:902
github.com/juanfont/headscale/cmd/headscale/cli.Execute()
[...]

PR #658, opened 2022-06-22
Quick fix to segfault on CLI when Headscale is not running

juanfont:
When we pass `--output json` or similar we suppress all the logs -which also deactivates the call to `os.Exit` when doing Fatal().

This is a quick fix for #652.

### DIFF
--- a/cmd/headscale/cli/utils.go
+++ b/cmd/headscale/cli/utils.go
@@ -117,7 +117,7 @@ func getHeadscaleCLIClient() (context.Context, v1.HeadscaleServiceClient, *grpc.
 	conn, err := grpc.DialContext(ctx, address, grpcOptions...)
 	if err != nil {
 		log.Fatal().Caller().Err(err).Msgf("Could not connect: %v", err)
-		os.Exit(-1) // we get here if logging is supressed (i.e., json output)
+		os.Exit(-1) // we get here if logging is suppressed (i.e., json output)
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/headscale/cli/utils.go
```diff
@@ -55,6 +55,7 @@ func getHeadscaleCLIClient() (context.Context, v1.HeadscaleServiceClient, *grpc.
 			Err(err).
 			Caller().
 			Msgf("Failed to load configuration")
+		os.Exit(-1) // we get here if logging is supressed (i.e., json output)
```
kradalby, line 58:
```suggestion
		os.Exit(-1) // we get here if logging is suppressed (i.e., json output)
```

cmd/headscale/cli/utils.go
```diff
@@ -116,6 +117,7 @@ func getHeadscaleCLIClient() (context.Context, v1.HeadscaleServiceClient, *grpc.
 	conn, err := grpc.DialContext(ctx, address, grpcOptions...)
 	if err != nil {
 		log.Fatal().Caller().Err(err).Msgf("Could not connect: %v", err)
+		os.Exit(-1) // we get here if logging is supressed (i.e., json output)
```
kradalby, line 120:
```suggestion
		os.Exit(-1) // we get here if logging is suppressed (i.e., json output)
```

kradalby (approved):

kradalby (approved):

Issue #662, opened 2022-06-24
Allow to specify initial API key as a flag for serve command


GrigoriyMikhalkin:
**Feature request**

It's probably somewhat controversial feature request. But it would be really useful if it would be possible to specify API key as a flag for `headscale serve` command.

Example of a case when it would be useful: there's is a service that serves as an intermediary between headscale server and tailscale clients(distributes auth keys). Predefined API key would be useful to simplify deployment of the whole setup. As without it, the only solution i see is a bit complex setup with SSH client built inside a service.

Is there some security concerns why it shouldn't be done or maybe some other known solutions for this case?

kradalby:
Hi

I dont think allowing arbitrary keys to be set is ideal, it seems to be going against common practices of how API keys are generated.

If this is for provisioning, why can you not use the provisioning tool to create the API key on the server, then read it back?

I know this can be done with Ansible at least?

kradalby:
Yes, Ansible can execute commands on the server and read stdout, so it should have no problems with excuting the creating of a apikey and reading it into a variable, [have a look at this stack overflow](https://stackoverflow.com/a/36060102).

juanfont:
Going to close this for the time being.

dimaqq:
I too would like provision "fixed" keys.

That would make orchestrating our stack so much easier, when `tailscale` / `headscale` is a very small part of the stack.

Ideally, I'd like to be able to launch headscale in the cloud, some edge device and tailscale-enabled containers all in one go.

Today, I have to make sure that headscale starts up first, use script to read the preauth keys out, pass those as env vars to edge and containers, and only then I can start the edge device and the tailscale-enabled containers. I find that confusing and brittle.

PR #666, opened 2022-06-26
add flags for headscale serve to provide initial api key

GrigoriyMikhalkin:
Just played around to check one of the possible solutions to https://github.com/juanfont/headscale/issues/662.

Here i introduced 3 new flags for `headscale serve` command: `api-key-prefix`, `api-key-pass` and `api-key-expiration`. Together they allow to save pre-generated API key into DB to use it later for authentication.


juanfont:
> Another, probably better, idea is to introduce new command: `headscale apikeys save --prefix<> --password<> --expiration<>`. That would be more flexible.

I agree. I think the place to create predefined api keys is in the `apikeys` subcommand. `serve` is... well... to serve the server.

kradalby:
I dont think we should implement this, comment in the issue.

juanfont:
Going to close this one, as we need to rethink a bit how to do it.

PR #668, opened 2022-06-30
graceful shutdown fix

GrigoriyMikhalkin:
There's few problems with a graceful shutdown that i tried to fix with this PR.

1. `shutdownChan` channel that was used to finish `PollNetMapStream` didn't work if no `PollNetMapStream` was running at the time of shutdown. Also, it could stop only single `PollNetMapStream` routine, as single value was passed to channel instead of closing it.
2. There is a race condition with function that handles termination signals, as it isn't included in an error group.

### DIFF
--- a/app.go
+++ b/app.go
@@ -95,6 +95,7 @@ type Headscale struct {
 	ipAllocationMutex sync.Mutex
 
 	shutdownChan chan struct{}
+	wg           sync.WaitGroup
 }
 
 // Look up the TLS constant relative to user-supplied TLS client
@@ -153,6 +154,7 @@ func NewHeadscale(cfg *Config) (*Headscale, error) {
 		privateKey:        privKey,
 		aclRules:          tailcfg.FilterAllowAll, // default allowall
 		registrationCache: registrationCache,
+		wg:                sync.WaitGroup{},
 	}
 
 	err = app.initDB()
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -96,6 +96,7 @@ type Headscale struct {
 	ipAllocationMutex sync.Mutex
 
 	shutdownChan chan struct{}
+	wg           sync.WaitGroup
```
kradalby, line 99: Could we give this a more descriptive name? what is the group waiting for?

app.go
```diff
@@ -677,7 +681,7 @@ func (h *Headscale) Serve() error {
 		syscall.SIGTERM,
 		syscall.SIGQUIT,
 		syscall.SIGHUP)
-	go func(c chan os.Signal) {
+	sig_func := func(c chan os.Signal) {
```
kradalby, line 684: sigFunc

kradalby (commented):
Looks reasonable, just some small nits

GrigoriyMikhalkin:
@kradalby I additionally decomposed `OIDCCallback` method. Otherwise `gocyclo` complaint about it's complexity.

app.go
GrigoriyMikhalkin: Updated the name. This wait group is for `PollNetMap` streams.

kradalby:
> @kradalby I additionally decomposed `OIDCCallback` method. Otherwise `gocyclo` complaint about it's complexity. 

Can you separate this into a different PR?

GrigoriyMikhalkin:
> > @kradalby I additionally decomposed `OIDCCallback` method. Otherwise `gocyclo` complaint about it's complexity.
> 
> Can you separate this into a different PR?

Done

juanfont (commented):
Only one minor thing to be fixed.

app.go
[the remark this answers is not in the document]
GrigoriyMikhalkin: That's already fixed in recent commits) https://github.com/juanfont/headscale/pull/668/files#diff-6c4b6ed7dc8834cef100f50dae61c30ffe7775a3f3f6f5a557517cb740c44a2dR675

juanfont (approved):

juanfont:
LGTM!

PR #670, opened 2022-07-06
ping db in health check

iSchluff:
Add basic db check to health endpoint

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/api.go
+++ b/api.go
@@ -30,6 +30,44 @@ const (
 	)
 )
 
+func (h *Headscale) HealthHandler(
+	writer http.ResponseWriter,
+	req *http.Request,
+) {
+	respond := func(err error) {
+		writer.Header().Set("Content-Type", "application/health+json; charset=utf-8")
+
+		res := struct {
+			Status string `json:"status"`
+		}{
+			Status: "pass",
+		}
+
+		if err != nil {
+			writer.WriteHeader(http.StatusInternalServerError)
+			log.Error().Caller().Err(err).Msg("health check failed")
+			res.Status = "fail"
+		}
+
+		buf, err := json.Marshal(res)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


api.go
```diff
@@ -30,6 +30,42 @@ const (
...
+		buf, err := json.Marshal(res)
+		if err != nil {
+			log.Error().Caller().Err(err).Msg("marshal failed")
+		}
+		writer.Write(buf)
+	}
+
+	err := h.pingDB()
+	if err != nil {
+		respond(err)
```
kradalby, line 62: We should not respond the error to the user, it is consider bad practice by OWASP;

https://cheatsheetseries.owasp.org/cheatsheets/Error_Handling_Cheat_Sheet.html

api.go
```diff
@@ -30,6 +30,42 @@ const (
...
 
+func (h *Headscale) HealthHandler(
+	writer http.ResponseWriter,
+	req *http.Request,
+) {
+	respond := func(err error) {
+		writer.Header().Set("Content-Type", "application/json; charset=utf-8")
+
+		res := struct {
+			Healthy string `json:"healthy"`
```
kradalby, line 41: I am happy with this, just wondering, is this a "defined" standard format for health?

I want:
1. Something is easily consumed by standard health checks, like Docker's, SystemD and Kubernetes
2. The closest to a standard as possible

kradalby (commented):
Overall quite happy with this, seems reasonable, posted a couple of comments and would you mind fixing the lint errors?

api.go
iSchluff: ok will change that

juanfont (approved):
Looks good! Thanks!

PR #685, opened 2022-07-21
Decompose OIDCCallback method

GrigoriyMikhalkin:
`gocyclo` complains about `OIDCCallback` method complexity. This PR decomposes this method into smaller ones to simplify it.

### DIFF
--- a/oidc.go
+++ b/oidc.go
@@ -136,6 +136,82 @@ func (h *Headscale) OIDCCallback(
 	writer http.ResponseWriter,
 	req *http.Request,
 ) {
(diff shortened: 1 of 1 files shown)


juanfont:
Hi @GrigoriyMikhalkin. Thanks for this PR. Happy to merge it.

One small thing. Could you please return the errors from the aux functions (instead of a bool for success)? That it will be able in the future to write some (very needed) tests for OIDC.

GrigoriyMikhalkin:
@juanfont Refactored aux functions to return errors.

juanfont (approved):
This is great!

Issue #690, opened 2022-07-25
Make server errors exported


GrigoriyMikhalkin:
There's a bunch of errors that `headscale` server can return. For example [`"Namespace already exists"`](https://github.com/juanfont/headscale/blob/main/namespaces.go#L19). Would it be desirable to export them?

Specifically i would find it useful as i use `headscale` as a lib to create a built-in client in a service. And there i want to check returned errors.

juanfont:
@GrigoriyMikhalkin happy to merge a PR doing this :)

Btw, out of pure curiosity, what is the use case?

GrigoriyMikhalkin:
@juanfont [Here](https://github.com/metal-stack/metal-api/pull/308/files#diff-e3be14987652bb48dcefbab95f677cfd5db025653af59c036ffdad1bb5d1f898R252) i'm trying to catch the `Namespace already exists` error.

It actually could be solved in another way(additional request to get all namespaces and try to find the one that's expected), but it just doesn't feel as an elegant solution.

I was thinking about using `GetNamespace` method, but it also returns an error which is not exported.

juanfont:
Happy to get all the errors exported. Can you prepare a PR?

GrigoriyMikhalkin:
@juanfont Yep, sure!

juanfont:
Fixed :)

PR #698, opened 2022-07-29
Export API errors

GrigoriyMikhalkin:
Solves https://github.com/juanfont/headscale/issues/690


utils.go
```diff
@@ -27,8 +27,8 @@ import (
 )
 
 const (
-	errCannotDecryptReponse = Error("cannot decrypt response")
-	errCouldNotAllocateIP   = Error("could not find any suitable IP")
+	ErrCannotDecryptReponse = Error("cannot decrypt response")
```
mjlshen, line 30: If these will be exported should we also fix this typo to `ErrCannotDecryptResponse`?

GrigoriyMikhalkin: Done

juanfont (approved):
LGTM! Thanks!

PR #700, opened 2022-07-31
Temporary fix integration tests with dedicated Dockerfile

juanfont:
Currently our integration tests are broken due to the changes in the Dockerfile to add multi-arch support. 

This is due to the lack of support in the dockertest library for the new BuildKit library from docker/moby. I have set a PR to dockertest to fix that https://github.com/ory/dockertest/pull/374

In the meantime, this PR adds a temporary extra Dockerfile in the old format, so we can keep working.


juanfont:
I am closing this. We will handle with a go mod replace to my fork.

PR #701, opened 2022-07-31
Fix integration test issues due to multi-arch

juanfont:
Currently our integration tests are broken due to the changes in the Dockerfile to add multi-arch support.

This is due to the lack of support in the dockertest library for the new BuildKit library from docker/moby. I have set a PR to dockertest to fix that https://github.com/ory/dockertest/pull/374

This PR changes the integration tests to use the patched version of dockertest, and replaces the library in go.mod to point to my fork. Hopefully this last part will be removed in a few days.


kradalby (dismissed):

juanfont:
Going to close this, as we need to investigate what is making the continuous integration tests fail.

Issue #582, opened 2022-05-12
Add ACL management via API


samson4649:
**Feature request**
Add management of the headscale ACLs via the API. This will allow for ACLs to be updated and current ACLS to be fetched from the headscale instance during runtime. 

|Method|Purpose|Comments|
|---|---|---|
|GET|Get currently active ACL|Data returned as JSON object|
|PUT|Set new active ACL|JSON blob provided to overwrite existing|

Aiming to get this started with JSON input and output to start.z

As well as making this a programmable interface for automating ACLs, this will also make integration tests for PR #581 easier to achieve.

restanrm:
Hello ! I love this idea and also wanted to work on this. I tried to make a proposal doc to describe this feature but didn't release it. As discussed in #492 it's not easy. We want to keep config as code principle as much as possible, but if the ACLs are in DB it's not the case. Also, the API modifying files on system is not ideal and would not work on k8s…

samson4649:
I've implemented a working update mechanism however im waiting merge for pr #581 (issue #492)

```bash
~ # ./headscale acl list
An updated version of Headscale has been found (0.16.0-beta4 vs. your current v0.15.0-389-g5d4dcb6-dirty). Check it out https://github.com/juanfont/headscale/releases
2022-06-20T07:11:37Z DBG Setting timeout timeout=5000
2022-06-20T07:11:37Z DBG HEADSCALE_CLI_ADDRESS environment is not set, connecting to unix socket. socket=/var/run/headscale/headscale.sock
{
	"groups": {
		"group:admin": [
			"host-1"
		],
		"group:boss": [
			"subnet-1"
		]
	},
	"hosts": {
		"host-1": "10.64.0.2/32",
		"subnet-1": "10.64.0.0/10"
	},
	"acls": [
		{
			"action": "accept",
			"src": [
				"boss"
			],
			"dst": [
				"boss:*"
			]
		},
		{
			"action": "accept",
			"src": [
				"admin"
			],
			"dst": [
				"dev1:*"
			]
		}
	]
}
~ # ./headscale acl update --file /etc/headscale/acl_2.hujson 
An updated version of Headscale has been found (0.16.0-beta4 vs. your current v0.15.0-389-g5d4dcb6-dirty). Check it out https://github.com/juanfont/headscale/releases
2022-06-20T07:11:43Z DBG Setting timeout timeout=5000
[...]

PR #581, opened 2022-05-12
added cli options to output ACLs

samson4649:
- [x] read the [CONTRIBUTING guidelines](README.md#user-content-contributing)
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated documentation if needed
- [x] updated CHANGELOG.md

Drafted to Fix #492

Issue #492, opened 2022-03-10
Print out loaded ACLs


12nick12:
Print out loaded ACLs. This would be awesome to make sure all ACLs are loaded.

kradalby:
We could have a command for inspecting the current loaded acls

restanrm:
I started to draw some lines around a new API to retrieve, modify and check ACLs based on the official tailscale API.

kradalby:
> I started to draw some lines around a new API to retrieve, modify and check ACLs based on the official tailscale API.

While I appreciate the need for API support, I really want this to be "config as code" first, I dont have any good idea on how to solve that yet.

I suppose a "study" of potential solutions would be sensible before going too far down one route.

restanrm:
Ok, I'll keep that in mind and try to list different kind of solutions.

PR #743, opened 2022-08-15
docs(proposals): add proposal for the ACL API

restanrm:
This Pull Request is a proposal for the ACL API that we could add to Headscale in order to improve usability of Headscale.

This is related to some issues:
- #492
- #582

Also the PR #581 is implementing part of the GET endpoint.


juanfont (approved):
This is very good. I like that it follows Tailscale API.

docs/proposals/002-acls-cli.md
```diff
@@ -0,0 +1,182 @@
...
+to allow ACL's manipulation from the API and not the filesystem. Manipulation
+of the ACL's via API would allow to change the ACL's via the command line, test
+the ACL's file before loading it. These actions would also be possible via a
+Web user interface.
+
+As a side effect, the file would not be the primary source of truth. We could
+still maintain this behavior with a switch in the configuration stating that we
+want to load the ACL's from a configured file if it exists. Also it's still
+possible to apply configuration as code like terraform is doing with static
+configuration loaded to destination with the API.
```
kradalby, line 14: I am trying to think if we could have something like "if nothing in the DB, then load from file", but that might be confusing to users as they might think they are updating it by updating on disk and restarting.

If we were to do a behaviour which is based on "keeping" the old behaviour, then I think it should use the API under the hood so we dont have to maintain too much duplicate code.

I would not be too oppose to removing the file behaviour, we just need to tell people a couple of versions in advance and write specific docs on how to load it now. That said, as an advocate for "config as code" and immutability, I do like the current setup where you do it in files.

docs/proposals/002-acls-cli.md
```diff
@@ -0,0 +1,182 @@
...
+  -o, --output string   Output format. Empty for human-readable, 'json', 'json-line' or 'yaml'
+
+Use "headscale acl [command] --help" for more information about a command.
+```
+
+### acl get subcommand
+
+```console
+headscale acl get --help
+Get will retrieve the ACLs currently loaded in Headscale. Use the `-o json`
```
kradalby, line 103: And YAML

juanfont: ♥️

docs/proposals/002-acls-cli.md
juanfont: Can we keep the file as the actual storage for the ACLs, rather than in the DB?

Even if consumed by the API, users would still have a good old file to do changes too.

restanrm: In this part I meant that the switch in the configuration is meant to overwrite the ACL's each time we reboot or reload the configuration. This is meant to keep the current behavior as using the file as a source of truth. This flag could even forbid changes to the rules stored in the database. I think it's interesting because not everyone wants to be use an API to manage the ACL's (I don't think you are the only one @kradalby :smile:). 

@juanfont I don't think that we can keep the actual storage for the ACL's. As an example for Headscale instances running in Kubernetes, it would require to much logic specific to Kubernetes to update the ConfigMap associated.

docs/proposals/002-acls-cli.md
restanrm: of course :smile:

docs/proposals/002-acls-cli.md
loprima-l: I think that we must add an option to chose between the config file and the database as, even if there's not an official support yet, a lot of peoples want to use Headscale with a GUI, and having the ACLs stored in a database will be more suitable for this purpose, but the "old techies"  would like to have file, maybe we can add a condition like if there is a file we use the file (and having an API route to know if ACLs are managed in a file), else storing the config in the database.

And maybe having an error on editing the ACLs with the API as writing a file doesn't seam a good practice to me while dealing with an API.

What do you think ?

loprima-l:
@juanfont Can you merge the PL please ? Or close it because I think the PL is not active anymore

PR #765, opened 2022-08-24
bugfix: Added base key validity period of 60 minutes.

### LINKED ISSUE #764: Headscale fails to activate clients with postgresql backend
**Bug description**

More specifically, 

In case of sqlite, 

In case of postgres, 

**To Reproduce**

1. Install postgresql 14

**Context info**

ishanjain28:
- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

Fixes #764 

PR Description,

Currently headscale initializes machine key's expiry time with `&time.Time{}`. This is formatted to, `0001-01-01T00:00:00Z`(in RFC3339 format). 

This means, The keys expiry time will always be in the past and clients will be stuck in a loop renewing the key. 

This PR addresses this problem by adding a minimum validity of 60 minutes to the generated keys.


ishanjain28:
This issue has not been fixed completely. 

Android clients are working fine and are getting the correct expiry time but Linux clients are still getting, 

> expiry                 | 0001-01-01 05:53:28+05:53:28

ishanjain28:
Fixed.

1. OIDC renewal code mentios that the function accepts current time but then passes, `&time.Time{}`. This has been changed to pass, `time.Now().Add(DefaultKeyExpireTime)`. 
2. In protocol_common.go, In the refresh operation it was not renewing the key. `expire` field in database was being overwritten to `0001-01-01T00:00:00Z`.

protocol_common.go
```diff
@@ -26,6 +26,10 @@ const (
 	//
 	// See also https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go
 	NoiseCapabilityVersion = 39
+
+	// DefaultKeyExpireTime determines the valid period of a generated key
```
juanfont, line 30: Comment should end in a period

protocol_common.go
```diff
@@ -26,6 +26,10 @@ const (
...
+	// Keys of new machines are valid for this period
```
juanfont, line 31: Comment should end in a period

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 0.17.1 (2022-XX-XX)
```
juanfont, line 3:
```suggestion
```

juanfont: Can you please add this under 0.17.0 ?

protocol_common.go
```diff
@@ -26,6 +26,10 @@ const (
...
+	DefaultKeyExpireTime = 60 * time.Minute
```
juanfont, line 32: Any rationale behind 60 minutes?

juanfont:
Digging into this, upstream only seems to say about key expiration this (in `RegisterRequest`)

	// Expiry optionally specifies the requested key expiry.
	// The server policy may override.
	// As a special case, if Expiry is in the past and NodeKey is
	// the node's current key, the key is expired.


Perhaps could be DefaultKeyExpireTime could be something configurable, with a default time of 60 minutes...

kradalby:

PR #767, opened 2022-08-25
Add support for PreAuthKey tags/automatic tagging

tsujamin:
- [X] added unit tests
- [X] added integration tests
- [x] updated CHANGELOG.md

These changes add the ability to automatically tag nodes when activated with a preauthkey as is available in SaaS

CreatePreAuthKey and like APIs were expanded to optionally accept a list of tags and the CLI arguments were updated. The database schema was also updated with a new table to store tags associated with preauth keys.

Keys associated to a preauthkey are applied to a new or existing Machine's ForcedKeys when registered/refreshed


tsujamin:
just to note that there's a small test break I'll need to push a change to depending on which order this and #763 get merged (if either is accepted

tsujamin:
fixed broken test from merge

restanrm (commented):
could you add a unit test to check that a node registered with such PreAuthKey have the proper Tags associated ?

tsujamin:
np will have a crack tomorrow - just noticed my local linting is broken too so will address both

restanrm:
> np will have a crack tomorrow - just noticed my local linting is broken too so will address both

Thanks ! Maybe you'll want to wait for an owner review first :smile:

restanrm:
> > np will have a crack tomorrow - just noticed my local linting is broken too so will address both
> 
> Thanks ! Maybe you'll want to wait for an owner review first smile

Sorry this comment was for another PR. No need for approval to add tests they are always welcome :smile:

tsujamin:
haha all good, fwiw just about to push a fix for the linting issues but not the unit test as the point where tags is applied is in `protocol_common.go` as part of machine registration/refresh as this most closely follows SaaS behaviour.

Let me know if an equivalent integration test is desired - just a bit of extra effort to stand up a new test case in there 🥲

protocol_common.go
```diff
@@ -355,6 +356,25 @@ func (h *Headscale) handleAuthKeyCommon(
 
 			return
 		}
+
+		aclTags := pak.toProto().AclTags
+		if len(aclTags) > 0 {
+			// This conditional preserves the existing behaviour, although SaaS would reset the tags on auth-key login
+			err = h.SetTags(machine, aclTags)
```
kradalby, line 360: If this _overrides_ the tags, then technically if the list is empty, we should still set it? as in, we should set it to no tags?

protocol_common.go
```diff
@@ -355,6 +356,25 @@ func (h *Headscale) handleAuthKeyCommon(
...
+		}
+
+		if err != nil {
```
kradalby, line 363: This error check should be inside the `if len(aclTags)` closure so we don't risk any weird scoping behaviour.

cmd/headscale/cli/preauthkeys.go
```diff
@@ -96,6 +108,15 @@ var listPreAuthKeys = &cobra.Command{
...
 			}
 
+			var aclTags string
+
+			if len(key.AclTags) > 0 {
+				for _, tag := range key.AclTags {
+					aclTags += "," + tag
+				}
+				aclTags = strings.TrimLeft(aclTags, ",")
+			}
```
kradalby, line 118: This `len` check is strictly not necessary, this should work:
```suggestion
			aclTags := ""

			for _, tag := range key.AclTags {
				aclTags += "," + tag
			}
			aclTags = strings.TrimLeft(aclTags, ",")
```

kradalby: Since the loop will not run if the list has 0 items, its safe to not check it.

grpcv1.go
```diff
@@ -106,11 +106,24 @@ func (api headscaleV1APIServer) CreatePreAuthKey(
...
+		for _, tag := range request.AclTags {
+			err := validateTag(tag)
+
+			if err != nil {
+				return &v1.CreatePreAuthKeyResponse{
+					PreAuthKey: nil,
+				}, status.Error(codes.InvalidArgument, err.Error())
+			}
+		}
+	}
```
kradalby, line 119:
```suggestion
	for _, tag := range request.AclTags {
		err := validateTag(tag)

		if err != nil {
			return &v1.CreatePreAuthKeyResponse{
				PreAuthKey: nil,
			}, status.Error(codes.InvalidArgument, err.Error())
		}
	}
```
We don't need to check for empty lists as `range` will just not have anything to iterate over.

preauth_keys.go
```diff
@@ -186,5 +234,11 @@ func (key *PreAuthKey) toProto() *v1.PreAuthKey {
 		protoKey.CreatedAt = timestamppb.New(*key.CreatedAt)
 	}
 
+	if len(key.ACLTags) > 0 {
+		for idx := range key.ACLTags {
+			protoKey.AclTags[idx] = key.ACLTags[idx].Tag
+		}
+	}
```
kradalby, line 241:
```suggestion
	for idx := range key.ACLTags {
		protoKey.AclTags[idx] = key.ACLTags[idx].Tag
	}
```

kradalby (commented):
Good work!, a few nits, one questions, otherwise good.

protocol_common.go
tsujamin: So this handles some divergent behaviour between SaaS and headscale. I believe it's possible in headscale for a machine to re-auth with a pre-auth key and keep its previously set tags, whereas upstream would reset the tags in my testing.

This logic tries to keep the best of both worlds, when a machine is re-authenticated with a pak:
  * if the pak has no associated tags, the machine keeps its previous tags (the current, non SaaS behaviour)
  * if the pak has tags, these tags replace any previously set on the machine (which would more closely align with a user's *intent*)

Removing the conditional would mean that, when a machine is re-authenticated with a pak, it would lose any previous tags. This is what SaaS does but I didn't want to force a change in headscales behaviour

tl;dr I feel this is messy but its somewhat behaviour preserving

cmd/headscale/cli/preauthkeys.go
tsujamin: :+1: ty this was me not understanding go at this point of the journey haha

protocol_common.go
kradalby: ok, make sense, we can live with that for now, and if we decide to change it, we will have to mark it as breaking.

tsujamin:
@kradalby thanks for the review - have merged in your recommendations

kradalby:
Seems like there is a lint error, and could you add changelog entry for this one as well?

kradalby (approved):

Issue #653, opened 2022-06-17
Toggle JSON logging


kazauwa:
**Feature request**

Allow to opt-in for JSON logs so they can be easily parsed.
I think the implementation is pretty straightforward, and I can get it done over the weekend.

kazauwa:
I suppose the implementation could be something like that:
https://github.com/juanfont/headscale/compare/main...kazauwa:feature/json-logs?expand=1

Setting `json_logs` to `true` swtiches to default zerolog formatting:
```
{"level":"info","path":"/var/lib/headscale/private.key","time":1655553834,"message":"No private key file at path, creating..."}
```

iSchluff:
this seems useful, would you mind creating a pull request?

kazauwa:
Done
Didn't want to create a PR before hearing from maintainers

iSchluff:
not a maintainer, just keen on this feature :)

PR #768, opened 2022-08-26
toggle json logging via config

kazauwa:
- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

Proposal pull request for #653


config-example.yaml
```diff
@@ -173,6 +173,7 @@ tls_cert_path: ""
 tls_key_path: ""
 
 log_level: info
+json_logs: false
```
kradalby, line 176: I think this should probably be something like:

```suggestion
log:
  format: json
```

Allowing us to expand in the future.

Level should be migrated to under that root, but that might be a future thing

kazauwa: I can migrate level as well, but this is a backward incompatible change, so we should be careful about it. Is there a guideline for these kinds of changes?

juanfont: We are still in a phase were we can afford performing not-that-breaking changes :)

kazauwa: Amazing!
I'm sorry for disappearing, was really busy these days. Hopefully will finish this MR this week.

juanfont (approved):

juanfont:
Thanks!!

PR #770, opened 2022-08-30
Fixes #766 Feature optional hostname suffix

### LINKED ISSUE #766: Optional random node suffixes.
The random node suffixes introduced in 0.16.0 mean machines onboarded with preauth keys end up on random MagicDNS hostnames. My use case is I generate preauth keys and unique hostnames for devices during provisioning programatically and I'd like for my software to be able to resolve the hostnames later within the namespace's subdomain.

I've got a branch [here](https://github.com/shanna/headscale/tree/feature-optional-hostname-suffix) that implements the change though I thought I better create a ticket first before submitting a PR since this may generate some discussion.

* I think suffixes should be off by default but YMMV. It does revert the new default prefixes introduced in 0.16.0 and I have no idea what Tailscale does by default. 🤷 
* The other ways this could be solved is by setting a node name in advance when you create a preauth key via the API or by honouring the `--hostname` tailscale cli flag? No idea how difficult these options would be to implement but I'm happy to have a crack if you'd rather not introduce another configuration variable.

shanna:
This PR fixes #766 by making hostname suffixes optional and off by default.

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 0.17.0 (2022-XX-XX)
 
+### BREAKING
+
+- Optional (off by default) random node suffixes. Reverts default Node DNS random suffixes introduced in 0.16.0.
+
+### Changes
+
 - Added support for Tailscale TS2021 protocol [#738](https://github.com/juanfont/headscale/pull/738)
 - Add ability to specify config location via env var `HEADSCALE_CONFIG` [#674](https://github.com/juanfont/headscale/issues/674)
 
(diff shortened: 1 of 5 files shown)


shanna:
I couldn't decide if I should extend tailscale DNSConfig so I'm not dumping DNS related configuration straight in the root configuration structure:

```
type Config struct {
  DNSConfig *DNSConfig
}

type DNSConfig struct {
  tailcfg.DNSConfig
  HostnameRandomSuffix bool
}
```

kradalby:
Commented on the issue

PR #780, opened 2022-09-03
Target Tailscale version to v1.30.0 (and replace inet.af/netaddr with net/netip)

juanfont:
This PR changes our target version to Tailscale 1.30, and switches Tailscale developers' inet.af/netaddr to the now mainline net/netip.


juanfont:
golangci-lint is broken due to the update to Go 1.19.

Will handle that in a new PR.

kradalby (approved):

PR #790, opened 2022-09-04
Add reverse proxy documentation

mike-lloyd03:
- Add reverse proxy documentation
- Update changelog

- [ ] added unit tests
- [ ] added integration tests
- [X] updated CHANGELOG.md

Per our discussion on Discord today, I'm adding a brief description for how to configure a reverse proxy in front of Headscale. The doc is rather concise so please let me know where more information is needed. This only covers nginx for now but more proxy servers can be added as necessary.


docs/reverse-proxy.md
```diff
@@ -0,0 +1,43 @@
...
+
+    location / {
+        proxy_pass http://<IP:PORT>;
+        proxy_set_header Host $server_name;
+        proxy_redirect http:// https://;
+        proxy_buffering off;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
+        add_header Strict-Transport-Security "max-age=15552000; includeSubDomains" always;
```
juanfont, line 40: This config won't work with Headscale 0.17.x & Tailscale > 1.30.0.

You are missing something like this for proxying WebSockets https://discord.com/channels/896711691637780480/896711692120129540/1015993290417504336

docs/reverse-proxy.md
```diff
@@ -0,0 +1,43 @@
+# Running behind a reverse proxy
+
+Running Headscale behind a reverse proxy is suitable for container-based deployments. This is especially useful on a server were port 443 is already being used for other web services.
+
+Headscale can be configured not to use TLS, leaving it to the reverse proxy to handle. Add the following configuration values to your headscale config file.
+
```
juanfont, line 6: Can you please point out that for Headscale v0.17.x+ having the reverse proxy configured to support WebSockets is mandatory? Otherwise Tailscale clients v1.30+ will not work.

WebSockets is also required when running the embedded DERP server.

In the embedded DERP case, even if you run a reverse proxy you will also have to open the specified STUN UDP port (see https://github.com/juanfont/headscale/blob/main/config-example.yaml#L85)

docs/reverse-proxy.md
mike-lloyd03: So I'm actually running 0.17.1-alpha right now and I have the `Upgrade` and `Connection` headers set but didn't include them in this PR because I didn't really know what they did haha.

Anyway, I did a bit more research and added them in. I'm not sure if the `map` variable is needed though. In my configuration, I just have the `Connection` header set to `upgrade`. I included it anyway. Let me know what you think.

docs/reverse-proxy.md
mike-lloyd03: I added a section saying that web socket configuration is mandatory going forward.

docs/reverse-proxy.md
```diff
@@ -0,0 +1,56 @@
+# Running behind a reverse proxy
+
+Running Headscale behind a reverse proxy is suitable for container-based deployments. This is especially useful on a server were port 443 is already being used for other web services.
+
+### Web Sockets
+The reverse proxy _must_ be configured to support websockets if you are running headscale 0.17.x+ and tailscale v1.30+.
```
juanfont, line 6:
```suggestion
The reverse proxy MUST be configured to support WebSockets, as it is needed for clients running Tailscale v1.30+. 

WebSockets support is required when using the headscale embedded DERP server. In this case, you will also need to expose the UDP port used for STUN (by default, udp/3478). Please check our [config-example.yaml](https://github.com/juanfont/headscale/blob/main/config-example.yaml).


```

docs/reverse-proxy.md
```diff
@@ -0,0 +1,56 @@
+# Running behind a reverse proxy
+
+Running Headscale behind a reverse proxy is suitable for container-based deployments. This is especially useful on a server were port 443 is already being used for other web services.
+
+### Web Sockets
```
juanfont, line 5:
```suggestion
### WebSockets
```

docs/reverse-proxy.md
```diff
@@ -0,0 +1,56 @@
+# Running behind a reverse proxy
+
+Running Headscale behind a reverse proxy is suitable for container-based deployments. This is especially useful on a server were port 443 is already being used for other web services.
```
juanfont, line 3:
```suggestion
Running headscale behind a reverse proxy is useful when running multiple applications on the same server, and you want to reuse the same external IP and port - usually tcp/443 for HTTPS.
```

docs/reverse-proxy.md
```diff
@@ -0,0 +1,56 @@
+# Running behind a reverse proxy
```
juanfont, line 1:
```suggestion
# Running headscale behind a reverse proxy
```

juanfont (commented):
Some suggestions...

mike-lloyd03:
Thanks for the suggestions.

The only issue I'm still having: Sometimes one of my Linux machines will show that it is `offline` every few minutes or so. After about a minute it returns to online. I'm not sure why this is only happening on only this machine and not on my Android device or the other two Linux machines I have configured. I'm doubting this has anything to do with the nginx config though.

mike-lloyd03:
Actually it looks like this is also happening on another Linux machine. The VPN connection is still up though so maybe it's not an issue.

juanfont:
Can you run prettier on the .md? The linter is complaining...

mike-lloyd03:
> Can you run prettier on the .md? The linter is complaining...

Done.

juanfont (approved):

PR #811, opened 2022-09-18
Fix subnet routers

kradalby:
This PR fixes (slightly hacky) an issue where subnet routers dont work as we dont pass the PrimaryRoutes parameter.


juanfont (approved):
LGTM!

Issue #661, opened 2022-06-23
Add SSH support to ACLs


GrigoriyMikhalkin:
**Feature request**

Tailscale introduced [Tailscale SSH](https://tailscale.com/kb/1193/tailscale-ssh/) feature, which allows to manage SSH connections as part of tailnet. In particular, user can control SSH access via ACLs. There's new field for that called `ssh` and here is an example of such config:
```
{
  "acls": [
    {
      "action": "accept",
      "src": ["*"],
      "dst": ["*:*"]
    }
  ],
  "ssh": [
    {
      "action": "accept",
      "src": ["autogroup:members"],
      "dst": ["autogroup:self"],
      "users": ["root", "autogroup:nonroot"]
    }
  ]
}
```

Would be great to see support for that in `headscale`. I would gladly try to help to implement this if this is ok.

juanfont:
Hi @GrigoriyMikhalkin, I am looking at it already - but want to merge some big stuff before.

But feel free to try :)

db48x:
I started on implementing this, but have not yet had a chance to test it. It is at least good enough to let `tailscale up --ssh` work as long as there is an explicit ACL, so presumably `tailscale ssh` will work as well. I’ll see if I can make time to test it this coming week, and perhaps continue the work if nobody beats me to it.

juanfont:
@617a7a I'll look into it as soon as we have the new protocol fully working

Thunderbottom:
@db48x tried the code from your fork, and I can't seem to get it to work even with ACLs explicitly specified, for example, my ACL configuration has:

```json
{
	"groups": {
		"group:example": [
			"example"  // Namespace
		]
	},
	"hosts": {
		"client": "100.64.0.2"
	},
	"acls": [
		{
			"action": "accept",
			"src": ["*"],
			"dst": ["*:*"]
		}
	],
	"ssh": [
		{
			"action": "accept",
			"src": ["autogroup:members"],
			"dst": ["client"],
			"users": ["autogroup:nonroot"]
		}

	]
}
```

and when I try to connect to the `client` host with `tailscale ssh`, I get `permission denied (tailscale)`. Is the current implementation lacking something? If it helps, the logs show me this:

```
2022-09-16T17:15:16+05:30 TRC ACL rules generated ACL=[{"DstPorts":[{"Bits":null,"IP":"*","Ports
":{"First":0,"Last":65535}}],"IPProto":[1,58,6,17],"SrcIPs":["*"]}]                             
[...]

PR #847, opened 2022-10-06
Add SSH ACL support

evenh:
Based on the fork of @db48x from #661. Doesn’t support the ‘autogroup’ ACL functionality.

This ACL works for us (formatted slightly for readability)
```json
{
  "acls":[{
      "action":"accept",
      "src":["group:employees"],
      "dst":["tag:proxy:*"]
    }
  ],
  "hosts":{},
  "groups":{
    "group:employees":[
      "john.doe",
      "jane.doe"
    ],
    "group:proxy":[
      "mycorp"
    ]
  },
  "tagOwners":{
    "tag:proxy":[
      "group:proxy"
    ]
  },
  "ssh":[
    {
      "action":"check",
      "src":["group:employees"],
      "dst":["tag:proxy"],
      "users":["some-allowlisted-user"],
      "checkPeriod":"8h"
    }
  ],
  "disableIPv4":false,
  "randomizeClientPort":false
}
```

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Fix OIDC registration issues [#960](https://github.com/juanfont/headscale/pull/960) and [#971](https://github.com/juanfont/headscale/pull/971)
 - Add support for specifying NextDNS DNS-over-HTTPS resolver [#940](https://github.com/juanfont/headscale/pull/940)
 - Make more sslmode available for postgresql connection [#927](https://github.com/juanfont/headscale/pull/927)
+- Add support for [SSH ACL](https://tailscale.com/kb/1018/acls/#tailscale-ssh) blocks [#847](https://github.com/juanfont/headscale/pull/847)
 
 ## 0.16.4 (2022-08-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,14 @@
 - Fix OIDC registration issues [#960](https://github.com/juanfont/headscale/pull/960) and [#971](https://github.com/juanfont/headscale/pull/971)
 - Add support for specifying NextDNS DNS-over-HTTPS resolver [#940](https://github.com/juanfont/headscale/pull/940)
 - Make more sslmode available for postgresql connection [#927](https://github.com/juanfont/headscale/pull/927)
-- Add support for [SSH ACL](https://tailscale.com/kb/1018/acls/#tailscale-ssh) blocks [#847](https://github.com/juanfont/headscale/pull/847)
+- Add experimental support for [SSH ACL](https://tailscale.com/kb/1018/acls/#tailscale-ssh) (see docs for limitations) [#847](https://github.com/juanfont/headscale/pull/847)
+  - Please note that this support should be considered _partially_ implemented
+  - SSH ACLs status:
+    - Support `accept` and `check` (SSH can be enabled and used for connecting and authentication)
+    - Rejecting connections **are not supported**, meaning that if you enable SSH, then assume that _all_ `ssh` connections **will be allowed**.
+    - If you decied to try this feature, please carefully managed permissions by blocking port `22` with regular ACLs or do _not_ set `--ssh` on your clients.
+    - We are currently improving our testing of the SSH ACLs, help us get an overview by testing and giving feedback.
+  - This feature should be considered dangerous and it is disabled by default. Enable by setting `HEADSCALE_EXPERIMENTAL_FEATURE_SSH=1`.
 
 ## 0.16.4 (2022-08-21)
 
--- a/acls.go
+++ b/acls.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/rs/zerolog/log"
 	"github.com/tailscale/hujson"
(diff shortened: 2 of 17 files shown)


machine.go
```diff
@@ -719,7 +719,11 @@ func (machine Machine) toNode(
 
 		KeepAlive:         true,
 		MachineAuthorized: !machine.isExpired(),
-		Capabilities:      []string{tailcfg.CapabilityFileSharing},
+		Capabilities: []string{
+			tailcfg.CapabilityFileSharing,
+			tailcfg.CapabilityAdmin,
```
evenh, line 724: Not sure if `CapabilityAdmin` is needed 🤔

db48x: This is one of the important open questions. If it isn’t present, the tailscale client won’t let you use `--ssh`. However, I don’t know what else it enables or if it should be included only for certain users.

acls.go
```diff
@@ -187,6 +198,111 @@ func (h *Headscale) generateACLRules() ([]tailcfg.FilterRule, error) {
...
+	}
+
+	for index, sshACL := range h.aclPolicy.SSHs {
+		action := rejectAction
+		switch sshACL.Action {
+		case "accept":
+			action = acceptAction
+		case "check":
+			checkAction, err := sshCheckAction(sshACL.CheckPeriod)
+			if err != nil {
```
db48x, line 240: In the error case we leave action in the reject state and continue processing the rule. Perhaps it would be better to continue to the next rule instead?

acls.go
```diff
@@ -187,6 +198,111 @@ func (h *Headscale) generateACLRules() ([]tailcfg.FilterRule, error) {
...
+			Principals:  principals,
+			SSHUsers:    userMap,
+			Action:      &action,
+		})
+	}
+
+	return rules, nil
+}
+
+func sshCheckAction(duration string) (*tailcfg.SSHAction, error) {
```
db48x, line 289: We should make similar constructor functions for the accept and reject actions as well.

db48x (commented):
It’s starting to look pretty good.

acls.go
evenh: Something like this?
```go
var action tailcfg.SSHAction

switch sshACL.Action {
case "accept":
	action = sshAcceptAction()
case "check":
	checkAction, err := sshCheckAction(sshACL.CheckPeriod)
	if err != nil {
		log.Error().
			Msgf("Error parsing SSH %d, check action with unparsable duration '%s'", index, sshACL.CheckPeriod)
	} else {
		action = *checkAction
	}
	fallthrough
default:
	action = sshRejectAction()

	log.Error().
		Msgf("Error parsing SSH %d, unknown action '%s'", index, sshACL.Action)

	return nil, err
}
```

db48x: No, that fallthrough is wrong. There needs to be a case for `"reject"` as well.

evenh: From my understanding of [the docs](https://tailscale.com/kb/1193/tailscale-ssh/#ensure-tailscale-ssh-is-permitted-in-acls), only `accept` or `check` is valid. Happy to be proven wrong as I haven't read all the available documentation :-)

db48x: Oh, that’s right. I had forgotten.

kradalby (commented):
I think this looks like sensible work on this feature, I have not worked too much on the ACLs and our coverage of the feature and unit tests are good, but it suffers from the problem of testing "our interpretation" of how it should work.

My main concern with this is that if we get it wrong, and users start relying on that, we have a hard time fixing it. With things like SSH, the worst possible scenario is that someone lock themselves out from a machine.

As mentioned in Discord, integration tests would make a lot of sense for this feature, so we have the opportunity to actually verify that the client does what we expect and intend. I am not sure if Tailscale SSH will play nicely in Docker, but I think it is worth a shot.

I don't _want_ to block this feature, the question is, how should we roll it out to minimise the impact if we get something wrong? (This hesitant approach comes from the Namespace/User/Tailnet issue we had earlier)

iSchluff:
@kradalby maybe put it behind a feature toggle to let users know that the feature is experimental for now?

kradalby:
Progress update: This is looking good, we are missing a few more test cases and @evenh and I have figured that we at least want:

- Given empty ACL file, no SSH should work
- Verify that given two namespaces, which only allow `:22` access in its own namespace, cross namespace SSH should not work
- SSH ACL rules are configured, but no rules to allow `:22`, should not allow SSH access.

I think that should be sufficient to call it beta and add more tests later as we find broken behaviour.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - Fix OIDC registration issues [#960](https://github.com/juanfont/headscale/pull/960) and [#971](https://github.com/juanfont/headscale/pull/971)
 - Add support for specifying NextDNS DNS-over-HTTPS resolver [#940](https://github.com/juanfont/headscale/pull/940)
 - Make more sslmode available for postgresql connection [#927](https://github.com/juanfont/headscale/pull/927)
+- Add support for [SSH ACL](https://tailscale.com/kb/1018/acls/#tailscale-ssh) blocks [#847](https://github.com/juanfont/headscale/pull/847)
```
kradalby, line 32:
```suggestion
- Add experimental support for [SSH ACL](https://tailscale.com/kb/1018/acls/#tailscale-ssh) (see docs for limitations) [#847](https://github.com/juanfont/headscale/pull/847)
```

How about something like this? Not sure where the best part to put docs would be actually.

kradalby: This comment is superseded by https://github.com/evenh/headscale/pull/6

kradalby (approved):
I'm happy with this, letting @juanfont and then we can get it in.

juanfont (approved):
Excellent!

dedene:
@juanfont should the ssh "check" feature in ACLs work? I never seems to be getting the `Tailscale SSH requires an additional check.` although the ssh ACL uses "check" instead of "accept" with a low checkPeriod.

PR #856, opened 2022-10-13
Integration tests v2

kradalby:
This is the beginning of reworking the integration tests into reusable components:

Goals:
- reusability
- consistency
- lower barrier for contributors 
- alternative control servers to verify that our tests are correct
- no sleep, more deterministic
- do not need to bind local ports

These should be consider an unstable API, as the content of this PR will not work for all our current tests, but serve as a starter.

This PR is the beginning and it ports the "Ping all nodes" test.


kradalby:
Depends on #865 to make the old tests pass

integration/control.go
```diff
@@ -0,0 +1,13 @@
+package integration
+
+import v1 "github.com/juanfont/headscale/gen/go/headscale/v1"
+
+type ControlServer interface {
```
juanfont, line 5: Why are you not passing any Context in the interface?

Could help for timeouts...

integration/scenario.go
```diff
@@ -210,3 +211,20 @@ func (s *Scenario) RunTailscaleUp(
 
 	return fmt.Errorf("failed to up tailscale node: %w", errNoNamespaceAvailable)
 }
+
+func (s *Scenario) GetIPs(namespace string) ([]netip.Addr, error) {
```
juanfont, line 319: Should we have GetIPv4 and GetIPv6?

integration/scenario.go
```diff
@@ -19,27 +19,29 @@ import (
...
-	"1.16.2",
-
-	// These versions seem to fail when fetching from apt.
-	// "1.14.6",
-	// "1.12.4",
-	// "1.10.2",
-	// "1.8.7",
-}
+var (
+	errNoHeadscaleAvailable = errors.New("no headscale available")
```
juanfont, line 27: why this is no const?

kradalby: Function calls can't be stored to const

integration/scenario.go
kradalby: Can add helpers to filter after, but since netip holds both in the same type, we might as well get all in one command call since that's expensive

integration/control.go
kradalby: Currently it's all based in errors, so not getting a reply in time is built into the retry functionality and doesn't seem to care about context

integration/scenario.go
juanfont: Argh. Sorry. Too tired. Read Error("foobar"), like in the regular code.

Issue #912, opened 2022-11-02
Bug during registration new macos node 


CNLHC:
**Bug description**

macOS standalone client can not login to the headscale server.

**To Reproduce**

1. setup v0.16.4 headscale server
2. install a standalone version of MacOS Client from [tailscale sites](https://pkgs.tailscale.com/stable/#macos) (using version 1.32)
3. using `defaults write io.tailscale.ipn.macsys ControlURL http://<XXX>:8080` to change the control server
4. Open APP and click log in 
5. The server return 401 and `Unauthorized`

**Context info**

The relevant context

- Version of headscale used：0.16.4

- Version of tailscale client：macOS Standalone GUI Client, Version 1.32

**The Reason I Guess**

1.  A URL like `http://XXX/register/<key_hex>` will send to the client from here. 
https://github.com/juanfont/headscale/blob/e112514a3bce6c1592a9e1514ddf750459244803/protocol_common.go#L496

2. But a URL like `http://XXX/register/node:<key_hex>` is required during the client login.
https://github.com/juanfont/headscale/blob/1eea9c943c8e1174795dd504ba490091495d7339/api.go#L98

3. And finally, The unexpected content was returned from here 
https://github.com/juanfont/headscale/blob/1eea9c943c8e1174795dd504ba490091495d7339/api.go#L111

**The Fix I Submitted**

https://github.com/juanfont/headscale/pull/911

PR #914, opened 2022-11-02
fix bug in #912

CNLHC:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -26,6 +26,7 @@ linters:
     - ireturn
     - execinquery
     - exhaustruct
+    - nolintlint
 
     # We should strive to enable these:
     - wrapcheck
--- a/cmd/headscale/cli/debug.go
+++ b/cmd/headscale/cli/debug.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 
+	"github.com/juanfont/headscale"
 	v1 "github.com/juanfont/headscale/gen/go/headscale/v1"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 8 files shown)


CNLHC:
Oh the test broken
I will try to solve the problem

kradalby:
I think this fails the tests because it changes how the ID is put into the cache for only one registration method, I think we should probably not change that, or normalise it so it is _always_ one thing for all methods. 

I notice that there was another fix submitted, but then withdrawn which might be related to a patch I did a while ago: https://github.com/juanfont/headscale/pull/918

I am happy with one of the following solutions:

- The ID used in reg url and the cache does _always_ contain `nodekey:`
- It _never_ contains `nodekey:`

CNLHC:
> I think this fails the tests because it changes how the ID is put into the cache for only one registration method, I think we should probably not change that, or normalise it so it is _always_ one thing for all methods.
> 
> I notice that there was another fix submitted, but then withdrawn which might be related to a patch I did a while ago: #918
> 
> I am happy with one of the following solutions:
> 
> * The ID used in reg url and the cache does _always_ contain `nodekey:`
> * It _never_ contains `nodekey:`

Thanks for your suggestion。

I have already made the test pass locally by prefixing everything with a `nodekey`, but I am not sure which style is preferred.

Actually, I am asking this question on the discord just before 3 minutes you post a comment here 😆

CNLHC:
I think there still remains one test to fix : `test_integration_v2_general`

kradalby:
@juanfont has some work on integration testing this, and he will test this branch with his work in #929 before we proceed.

kradalby (approved):

PR #927, opened 2022-11-03
Make more sslmode available for postgresql connection (Fixes #925)

### LINKED ISSUE #925: Headscale can not connect to postgresql with SNI in used.
**Bug description**

Can't get postgresql works with TLS and SNI.

**To Reproduce**

Try connect to a postgresql database with SNI in used.

**Context info**

Refs to [this]() document of neon, a hosted postgresql service.

```
If your application or service uses golang PostgreSQL clients like pgx and lib/pg you can set sslmode=verify-full, which will cause SNI info to be sent. Most likely, this was not intentional but happened inadvertently due to the golang's TLS library API design.
``` 

But in headscale, `db_ssl` is defined as a bool `DBssl`.

https://github.com/juanfont/headscale/blob/ca8bca98edca4dced2331071ef67fa0e0b0d0942/config.go#L543

Therefore `ssl_mode` can only be `disabled` or `require` (psql's default), and user won't be able to connect to any hosted postgresql server depend on SNI in TLS.

I suggest we change `DBssl` to string to fix this issue.

OrvilleQ:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

#925 

Change `DBssl` to `String` while still compatible with old boolean configuration.

### DIFF
--- a/app.go
+++ b/app.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -127,8 +128,12 @@ func NewHeadscale(cfg *Config) (*Headscale, error) {
 			cfg.DBuser,
 		)
 
-		if !cfg.DBssl {
-			dbString += " sslmode=disable"
+		if s, err := strconv.ParseBool(cfg.DBssl); err == nil {
+			if !s {
+				dbString += " sslmode=disable"
+			}
+		} else {
+			dbString += fmt.Sprintf(" sslmode=%s", cfg.DBssl)
 		}
 
 		if cfg.DBport != 0 {
--- a/app.go
+++ b/app.go
@@ -128,8 +128,8 @@ func NewHeadscale(cfg *Config) (*Headscale, error) {
 			cfg.DBuser,
 		)
 
-		if s, err := strconv.ParseBool(cfg.DBssl); err == nil {
-			if !s {
+		if sslEnabled, err := strconv.ParseBool(cfg.DBssl); err == nil {
+			if !sslEnabled {
 				dbString += " sslmode=disable"
 			}
 		} else {
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -137,14 +137,15 @@ db_path: ./db.sqlite
 
 # # Postgres config
 # If using a Unix socket to connect to Postgres, set the socket path in the 'host' field and leave 'port' blank.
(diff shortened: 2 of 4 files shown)


OrvilleQ:
@juanfont @kradalby Hello guys.

I know you guys are busy working on the next release, but would you like to take a moment to look at this pull request? Is a small one, and I really hope it can be included in the next release as someone who really needs this fix.

config-example.yaml
```diff
@@ -137,6 +137,8 @@ db_path: ./db.sqlite
 
 # # Postgres config
 # If using a Unix socket to connect to Postgres, set the socket path in the 'host' field and leave 'port' blank.
+# Also if other 'sslmode' is required instead of 'require(true)' and 'disabled(false)', set the 'sslmode' you need
+# in the 'db_ssl' field. Refers to https://www.postgresql.org/docs/current/libpq-ssl.html Table 34.1.
```
kradalby, line 141: Can you move this line so it belongs to the `db_ssl:` entry?

app.go
```diff
@@ -127,8 +128,12 @@ func NewHeadscale(cfg *Config) (*Headscale, error) {
 			cfg.DBuser,
 		)
 
-		if !cfg.DBssl {
-			dbString += " sslmode=disable"
+		if s, err := strconv.ParseBool(cfg.DBssl); err == nil {
+			if !s {
+				dbString += " sslmode=disable"
+			}
```
kradalby, line 134: Do we need the inverse of this?

If `s` is true, do we need to add `sslmode=enable` ?

```suggestion
		if sslEnabled, err := strconv.ParseBool(cfg.DBssl); err == nil {
			if !sslEnabled {
				dbString += " sslmode=disable"
			}
```

kradalby (commented):
Apologise for the delay, one question, two nits.

config-example.yaml
OrvilleQ: Sure I'll.

app.go
OrvilleQ: `sslmode=on` is the default behavior so we'll be fine without it.

kradalby (dismissed):

kradalby:
oh, apologies, could you also add a `CHANGELOG.md` entry for this change? otherwise good!

OrvilleQ:
> oh, apologies, could you also add a `CHANGELOG.md` entry for this change? otherwise good!

Sure, I've done it.

kradalby (approved):

PR #929, opened 2022-11-03
Add auth web flow integration tests

juanfont:
We had missing the integration tests for the web flow auth. We had already for auth keys and OIDC.

This PR adds those tests, using the v2 of the integration system. As soon as it works...


.github/workflows/test-integration-v2-general-auth.yml
```diff
@@ -0,0 +1,35 @@
+name: Integration Test Authentication v2
```
kradalby, line 1: Lets not split these anymore, the new tests can run in parallell, so lets not start by splitting them.

Makefile
```diff
@@ -68,6 +68,17 @@ test_integration_v2_general:
...
+test_integration_v2_auth_web_flow:
+	docker run \
+		-t --rm \
+		-v ~/.cache/hs-integration-go:/go \
+		--name headscale-test-suite \
+		-v $$PWD:$$PWD -w $$PWD/integration \
+		-v /var/run/docker.sock:/var/run/docker.sock \
+		golang:1 \
+		go test ./... -timeout 60m -parallel 6 -run TestAuthWebFlow
+
```
kradalby, line 81:
```suggestion

```

integration/auth_web_flow_test.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (s *AuthWebFlowScenario) runTailscaleUp(
+	namespaceStr, loginServer string,
+) error {
+	log.Printf("running tailscale up's for namespace %s", namespaceStr)
```
kradalby, line 114:
```suggestion
	log.Printf("running tailscale up for namespace %s", namespaceStr)
```

PR #940, opened 2022-11-07
Add support for NextDNS resolver

### LINKED ISSUE #939: Support NextDNS DoH resolver
**Feature request**

[NextDNS](https://nextdns.io/) is a hosted adblocking DNS resolver that is supported in the hosted Tailscale. It would be nice if this would be supported in Headscale too.

The support is baked into tailscaled (along with other DNS-over-HTTPS resolvers) so there is not much that needs to be done on the Headscale side.

Each configuration in NextDNS has an ID (an example `abcdef` from now on) that is needed to identify the user. After snooping around the tailscale code it looks like everything needed is to set a resolver (in DNSConfig) with addr `https://dns.nextdns.io/abcdef` and the tailscale clients will use NextDNS as their resolver.

In the config a nice way of using it would be:
```yaml
dns_config:
  nameservers:
    - nextdns:abcdef
```

Is this something that could be considered?

arnarg:
This adds support for easily using [NextDNS](https://nextdns.io/) as a DoH resolver in tailscale network controlled by Headscale. More info in #939 

To enable you would put the following in config:

```yaml
dns_config:
  nameservers:
    - nextdns:abcdef # Where abcdef is your NextDNS ID
```

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #939

### DIFF
--- a/config.go
+++ b/config.go
@@ -383,10 +383,21 @@ func GetDNSConfig() (*tailcfg.DNSConfig, string) {
 		if viper.IsSet("dns_config.nameservers") {
 			nameserversStr := viper.GetStringSlice("dns_config.nameservers")
 
-			nameservers := make([]netip.Addr, len(nameserversStr))
-			resolvers := make([]*dnstype.Resolver, len(nameserversStr))
+			nameservers := []netip.Addr{}
+			resolvers := []*dnstype.Resolver{}
 
-			for index, nameserverStr := range nameserversStr {
+			for _, nameserverStr := range nameserversStr {
+				// Search for explicit DNS-over-HTTPS resolvers
+				if strings.HasPrefix(nameserverStr, "https://") {
+					resolvers = append(resolvers, &dnstype.Resolver{
+						Addr: nameserverStr,
+					})
+
+					// This nameserver can not be parsed as an IP address
+					continue
+				}
+
+				// Parse nameserver as a regular IP
 				nameserver, err := netip.ParseAddr(nameserverStr)
 				if err != nil {
 					log.Error().
@@ -395,10 +406,10 @@ func GetDNSConfig() (*tailcfg.DNSConfig, string) {
 						Msgf("Could not parse nameserver IP: %s", nameserverStr)
 				}
 
-				nameservers[index] = nameserver
-				resolvers[index] = &dnstype.Resolver{
+				nameservers = append(nameservers, nameserver)
+				resolvers = append(resolvers, &dnstype.Resolver{
 					Addr: nameserver.String(),
-				}
+				})
 			}
 
 			dnsConfig.Nameservers = nameservers
--- a/dns.go
+++ b/dns.go
@@ -3,11 +3,13 @@ package headscale
 import (
 	"fmt"
 	"net/netip"
+	"net/url"
 	"strings"
 
 	mapset "github.com/deckarep/golang-set/v2"
 	"go4.org/netipx"
 	"tailscale.com/tailcfg"
+	"tailscale.com/types/dnstype"
 	"tailscale.com/util/dnsname"
 )
 
@@ -18,6 +20,10 @@ const (
 const (
 	ipv4AddressLength = 32
 	ipv6AddressLength = 128
+)
+
+const (
+	nextDNSDoHPrefix = "https://dns.nextdns.io"
 )
 
 // generateMagicDNSRootDomains generates a list of DNS entries to be included in `Routes` in `MapResponse`.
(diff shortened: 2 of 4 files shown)


config.go
```diff
@@ -383,10 +383,21 @@ func GetDNSConfig() (*tailcfg.DNSConfig, string) {
...
+			for _, nameserverStr := range nameserversStr {
+				// Search for a nextdns pattern
+				if strings.HasPrefix(nameserverStr, "nextdns:") {
+					nextdnsID := strings.TrimPrefix(nameserverStr, "nextdns:")
+					resolvers = append(resolvers, &dnstype.Resolver{
+						Addr: fmt.Sprintf("https://dns.nextdns.io/%s", nextdnsID),
+					})
+
+					continue
+				}
```
arnarg, line 398: This parsing logic might be a bit naive. Would you want this to be improved and maybe even use regex?

dns.go
```diff
@@ -184,5 +184,18 @@ func getMapResponseDNSConfig(
...
+	// Add device identifier to the nextdns resolvers, if there are any.
+	for _, resolver := range dnsConfig.Resolvers {
+		if strings.HasPrefix(resolver.Addr, "https://dns.nextdns.io/") {
+			resolver.Addr = fmt.Sprintf(
+				"%s?device_name=%s&device_model=%s&device_ip=%s",
+				resolver.Addr,
+				machine.GivenName,
+				machine.HostInfo.OS,
+				machine.IPAddresses[0],
+			)
```
arnarg, line 196: This just adds a nice machine identifier in the request log in the NextDNS dashboard.
Machine identifier example

config.go
arnarg: For context, the idea here is to provide this to the list of nameservers like this:

```yaml
dns_config:
  nameservers:
    - nextdns:abcdef # where abcdef is the NextDNS ID
```

arnarg:
I updated some of the code as I wasn't happy with everything and I realized that in `getMapResponseDNSConfig` the dnsConfigOrig is only cloned when if magic dns is enabled. I believe NextDNS resolver can be enabled without magic DNS but as is it will overwrite the resolver in the global config, appending machine specific config over and over again.

What do you think would be more appropriate to resolve that.

- Only do the machine config appending to the nextdns resolver when magic DNS is enabled
- _Always_ clone the `dnsConfigOrig` in case there is a nextdns resolver in there that needs to be modified for the machine

I personally think the second would be better as the DoH resolvers don't have much to do with magic DNS.

iSchluff:
As far as I understand the code in tailscale you should just be able to provide a nexdns ipv6 ip in the headscale dns configuration, and it should work without any changes

arnarg:
Yeah you are correct. I had not seen this. This will at least make it possible to use NextDNS DoH resolver as is, good to know!

However, from a usability point of view I would still argue that `nextdns:abcdef` would be nicer to specify than `2a07:a8c0::ab:cdef`. But more importantly the ipv6 way does not support adding the metadata so that hosts are identifiable in the request log.

Do you think you'd be open for adding this feature at all or not?

config.go
```diff
@@ -383,10 +383,28 @@ func GetDNSConfig() (*tailcfg.DNSConfig, string) {
...
+							Addr: fmt.Sprintf("https://dns.nextdns.io/%s", nextdnsID),
+						})
+					} else {
+						log.Error().
+							Str("func", "getDNSConfig").
+							Msgf("Could not parse nextdns nameserver: should match pattern 'nextdns:<your-id>'")
+					}
+
+					continue
+				}
```
kradalby, line 405: From what I understand, this logic is to convert:

`nextdns:abcd123` => `https://dns.nextdns.io/abcd123` ?

I am not really that keen on hardcoding dedicated providers into the config handling, this sounds like it could grow out of hand if NextDNS starts getting competitors in the DoH space.

It sounds like rather documenting how one can enter `https://dns.nextdns.io/abcd123` and make sure that experience is smoother is a better approach.

dns.go
```diff
@@ -184,5 +185,21 @@ func getMapResponseDNSConfig(
 		dnsConfig = dnsConfigOrig
 	}
 
+	// Add device identifier to the nextdns resolvers, if there are any.
+	for _, resolver := range dnsConfig.Resolvers {
+		if strings.HasPrefix(resolver.Addr, "https://dns.nextdns.io/") {
```
kradalby, line 190: This would have similar concerns, I'm not too keen on having dedicated vendor stuff all over the code, but for this particular one, I could be convinced that having a generic interface/options/middelware system to override with could be appropriate, which is called from this function, but not in this function.

config.go
arnarg: That's fair and I can agree with that.

So this if statement could just look for a `https://` prefix instead and put it into the resolvers slice as is. Then in `getMapResponseDNSConfig` anything specific can be done that is needed for adding NextDNS machine metadata (but in another function/interface like mentioned in the other comment).

Would you be ok with that?

kradalby: Yes in `getMapResponseDNSConfig`, we can call a function which modifies the resolver map.

`addNextDNSMetadata(resolvers *dnsConfig.Resolvers)` or something more appropriate.

arnarg: Got it! I'll modify this soon.

kradalby (approved):

arnarg:
> @arnarg looks good now, can you add a CHANGELOG entry and some docs?

I have added to the CHANGELOG and I couldn't find any good place for this in docs except as comments in the example config. Is this acceptable?

kradalby (approved):

PR #942, opened 2022-11-08
Add host info to machine

joshuataylor:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Hi,

I would like to have the hostInfo exposed, which allows gGRPC to show full host info.

See here: https://pkg.go.dev/tailscale.com/tailcfg@v1.32.2#Hostinfo

I'm unsure how to add tests around this.

I've tested this with my service which calls the gRPC endpoint, and it works :tada:.

image


kradalby:
Hi, while we do appreciate contribution, this one concerns me quite a bit as it hand types an upstream resource which will likely drift really fast and we would have the burden of making sure they stay in sync, which isnt really realistic.

This data could be exposed over gRPC, but then it should be a generic `interface` or JSON or something that allow you to unmarshal it to the upstream type on the other side.

joshuataylor:
Makes sense, and I agree.

My knowledge of Go is pretty limited, is there an easier way to do this (like a struct/interface/etc)? That way we could mimic how tailscale works, but if they change the way hostinfo works tailscale won't complain when the dep is bumped.

kradalby:
Closing this for now, it will require a large rebase after our code reorg work goes in. If picked up again, please open when ready.

PR #959, opened 2022-11-14
Releases: use flavor to set the tag suffix

snh:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Moves to using the `suffix` option in `flavor` from [metadata-action](https://github.com/docker/metadata-action) to ensure  the suffix is applied to all tags. Without this the Alpine image was ending up with the `<version>` tag instead of the Debian image.

Also removed the `type=raw,value=latest` and equivalent options from all three images, as the default behavior of `latest=auto` in `flavor` should take care of this. The `onlatest=true` in the `suffix` option will ensure the Debug and Alpine images will be `latest-debug` and `latest-alpine` respectively.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,7 +65,6 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-            type=raw,value=latest
             type=sha
       - name: Login to DockerHub
         uses: docker/login-action@v1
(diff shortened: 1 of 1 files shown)


snh:
Noting that this will still be required even if the Alpine image is removed (https://github.com/juanfont/headscale/pull/962), as otherwise the Debug image will end up being tagged with the `<version>` tag instead.

kradalby:
Looks reasonable, we have removed Alpine, still happy to have these changes.

snh:
@kradalby Thanks 😄 I've rebased now that the Alpine removal has been merged.

PR #974, opened 2022-11-16
Bump meta action and always include latest

snh:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

A regression from https://github.com/juanfont/headscale/pull/959 is that the auto `latest` handling will intentionally avoid labeling a pre-release version, such as a beta, as `latest`.

I gather we want `latest` to point to the latest release regardless of if it is a pre-release like the previous behavior? If not, then the new behavior can be left alone, with only stable releases being tagged as `latest`. Though we'll probably want to create a new tag to use to reference the latest pre-release version for those who want to following along with development.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,12 +55,14 @@ jobs:
             ${{ runner.os }}-buildx-
       - name: Docker meta
         id: meta
-        uses: docker/metadata-action@v3
+        uses: docker/metadata-action@v4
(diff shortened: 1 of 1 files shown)


snh:
Closing after discussion on Discord, as we'll leave `latest` pointing at stable, and I'll make a new PR to attach a tag to all new pre-release versions.

PR #988, opened 2022-11-19
Proposal: Better routing management

juanfont:
This PR includes a proposal to improve how we handle the routing and subnet routers.

### DIFF
--- a/docs/proposals/002-better-routing.md
+++ b/docs/proposals/002-better-routing.md
@@ -1,0 +1,48 @@
+# Better route management
+
+As of today, route management in Headscale is very basic and does not allow for much flexibility, including implementing subnet HA, 4via6 or more advanced features. We also have a number of bugs (e.g., routes exposed by ephemeral nodes)
+
+This proposal aims to improve the route management.
+
+## Current situation
+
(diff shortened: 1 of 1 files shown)


docs/proposals/002-better-routing.md
```diff
@@ -0,0 +1,48 @@
...
+
+In case of colliding routes (i.e., subnets that are exposed from multiple nodes), we are currently just sending all of them in `PrimaryRoutes`... and hope for the best. (`PrimaryRoutes` is the field in `Node` used for subnet failover).
+
+## Proposal
+
+The core part is to create a new `Route` struct (and DB table), with the following fields:
+
+```go
+type Route struct {
+	ID          uint64 `gorm:"primary_key"`
```
kradalby, line 21: This is very sidenote, but we can probably stop defining a bunch of these fields and inherit them: https://gorm.io/docs/models.html#gorm-Model

kradalby (approved):

docs/proposals/002-better-routing.md
juanfont: Right!

PR #1023, opened 2022-11-28
Add nolint for t.parallel

juanfont:
Add nolint for some calls to t.Parallel

### DIFF
--- a/integration/scenario_test.go
+++ b/integration/scenario_test.go
@@ -21,7 +21,7 @@ func IntegrationSkip(t *testing.T) {
 	}
 }
 
-func TestHeadscale(t *testing.T) {
+func TestHeadscale(t *testing.T) { // nolint
 	IntegrationSkip(t)
 	t.Parallel()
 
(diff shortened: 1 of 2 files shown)


integration/scenario_test.go
```diff
@@ -21,7 +21,7 @@ func IntegrationSkip(t *testing.T) {
 	}
 }
 
-func TestHeadscale(t *testing.T) {
+func TestHeadscale(t *testing.T) { // nolint
```
kradalby, line 24: Drop this?

integration/scenario_test.go
```diff
@@ -21,7 +21,7 @@ func IntegrationSkip(t *testing.T) {
...
 	IntegrationSkip(t)
 	t.Parallel()
```
kradalby, line 26:
```suggestion
	t.Parallel() // nolint
```

kradalby (commented):
This looks nicer?

Issue #1010, opened 2022-11-23
--snat-subnet-routes=true not work


pingod:
**Bug description**
--snat-subnet-routes=true not work

**To Reproduce**
1. tailscale up --login-server=http://server_url --accept-routes=true --accept-dns=false --advertise-routes=10.0.0.234/32--snat-subnet-routes=true

2. headscale routes enable -i 12  -r "10.0.0.234/32"

3. headscale node route list -i 12
Route         | Enabled
10.0.0.234/32 | true

**Context info**
i use tcpdump at one of node(ip:10.0.0.234) :

tcpdump -i ens160 icmp and host 10.0.9.3 -nn -vvv
tcpdump: listening on ens160, link-type EN10MB (Ethernet), capture size 262144 bytes
13:52:26.379681 IP (tos 0x0, ttl 63, id 42914, offset 0, flags [DF], proto ICMP (1), length 84)
    10.0.9.3 > 10.0.0.234: ICMP echo request, id 18, seq 93, length 64
13:52:26.379754 IP (tos 0x0, ttl 64, id 29816, offset 0, flags [none], proto ICMP (1), length 84)
    10.0.0.234 > 10.0.9.3: ICMP echo reply, id 18, seq 93, length 64
13:52:27.392447 IP (tos 0x0, ttl 63, id 42921, offset 0, flags [DF], proto ICMP (1), length 84)
    10.0.9.3 > 10.0.0.234: ICMP echo request, id 18, seq 94, length 64

So, --snat-subnet-routes=true not work

Version of headscale used: v0.17.0-beta4
Version of tailscale client:  1.32.2
OS: Ubuntu 18

Issue #260, opened 2021-12-28
Remove old routes


12nick12:
Is there any way to remove old routes from tailscale/headscale? 

I deleted an old device and created a new one in it's place, when trying to enable a subnet router the device goes offline. This appears to be because of a loop I'm thinking.

kradalby:
Is this still a problem?

epichub:
I have a route that I cannot disable, if i "--reset" join the router node without advertising the route, the other nodes still gets the route added (after restarting headscale and tailscale on all nodes) - the route is no longer listed by headscale cli for the subnet router, but added to the other nodes anyway.

casdr:
I have the same issue, the problem can be reproduced as following:

- Add a route from one of the nodes
- Enable the route using headscale cli
- Remove the route advertisement by removing the route from the `--advertise-routes` list

Now, when you check `headscale route ls -i (node)` the route doesn't show up, however the route is still listed in the sqlite database and also being configured on the other nodes.

casdr:
I found a workaround for this, when you enable all routes for the node (`headscale r enable -i 2 -a`), the old route will be removed.

PsyMan2000:
I am running 0.19.0 in a container and am also seeing a few routes sticking around after deleting a node and re-adding it (I did it to remove the IPv6 addresses and just use IPv4 after disabling it in the server 3 firewalls in) - now when I "headscale route ls" is gives me 3 Advertised routes with an empty Machine entry with the new 3 Machines and same routes tagged on the end of the list. Am I experiencing a bug or did I do something wrong. Also the commands seem to have changed a bit in 0.19.0. is there a "headscale route delete -r x" option or is there a simpler way to remove the ghosts from the machine? :-) 

Maybe I should have re-upped it removing the advertised routes prior to deleting the node with the --reset flag? I will try that with another node and see what happens. I am only testing so no huge problem to wipe and start over. Looking forward to adding the UI on top once I get HS mastered

robertorubioguardia:
Hi there.

Having the same issue here. Routes are still listed and shown as advertised in the headscale server, without a node label, even when the tailscale node has been removed from the network.

The headscale CLI seems not to have a command to delete routes :( but one just to disable routes.

Any ideas/insight?

evenh:
I've seen the same behaviour

jsbrain:
Same ... cannot remove old routes from deleted device.

thejinx0r:
I was having the same issue and discovered that this might work:
[...]

PR #1024, opened 2022-11-28
Refactor routing code and add support for subnet failover (HA)

juanfont:
This PR implements https://github.com/juanfont/headscale/pull/988, for better route management.

It enables Headscale to:

- Enable and disable advertised subnets
- Handle exit nodes
- Manage multiple nodes offering the same route 
- Failover to another node when the primary node for a subnet is disconnected

Should also implement https://github.com/juanfont/headscale/issues/1010, and https://github.com/juanfont/headscale/issues/260.


machine.go
```diff
@@ -1106,9 +1109,9 @@ func (h *Headscale) EnableAutoApprovedRoutes(machine *Machine) error {
 		}
 	}
 
-	for _, approvedRoute := range approvedRoutes {
-		approvedRoute.Enabled = true
-		err = h.db.Save(&approvedRoute).Error
+	for i, approvedRoute := range approvedRoutes {
+		approvedRoutes[i].Enabled = true
+		err = h.db.Save(&approvedRoutes[i]).Error
```
kradalby, line 1114: Why?

cmd/headscale/cli/routes.go
```diff
@@ -64,28 +59,39 @@ var listRoutesCmd = &cobra.Command{
...
+		var routes []*v1.Route
 
-		response, err := client.GetMachineRoute(ctx, request)
-		if err != nil {
-			ErrorOutput(
-				err,
-				fmt.Sprintf("Cannot get nodes: %s", status.Convert(err).Message()),
-				output,
-			)
+		if machineID == 0 {
```
kradalby, line 68: Why is `0` special?

kradalby: oh is zero when you didnt pass the flag?

juanfont: yes!

machine.go
juanfont: The linters were complaining (rightfully so) of a common pitfall of Go. The pointer of `approvedRoute` is always be the same, so just in case we use here the index.

kradalby (dismissed):

kradalby (approved):

PR #1058, opened 2022-12-09
Fix duplicate nodes due to incorrect implementation of the protocol

### LINKED ISSUE #1054: Logout and login back got different IP address
**Bug description**
I use latest code with a 3rd party OIDC connection. I found that if I logout and login again with the same user, Tailscale client will get a different IP address. 

I think it's related to TS2021 protocol. The reason is that after logout, the content of /var/lib/tailscale/tailscaled.state file is like this:

```json
{
  "_current-profile": "cHJvZmlsZS0yZjFh",
  "_machinekey": "cHJpdmtleTpjMGQzNzNlMGVmYzRhNjc5MTgyMTNkMGE1NzgzYWNkMGRhZTUyMGI5NzBhYjMxODQyYTExNzQxNDRmMDc1ODUz",
  "_profiles": "e30=",
  "profile-2f1a": null
}
```

When logging back in, the only anchor left in state file is machine key, which is set to all 0 when using TS2021, so Headscale is not able to relate this machine to any previous record. 

It looks very like the problem lies in Tailscale client, which should preserve nodekey and rotate it as old nodekey after logout. But the same Tailscale client, if I connect to Tailscale control server, logout and login again, it will get back same IP address. Logout behaviour related to state file is the same (only machine key is remained after logout). 

**To Reproduce**

1.  `rm -rf /var/lib/tailscale`
2. start tailscaled
3. `tailscale up --login-server http://my.login.server`
4. finish login process
5. `tailscale status`, show IP address
6. tailscale logout
7. `tailscale up --login-server http://my.login.server`
8. finish login process
9. `tailscale status`, the IP address is different from step 5 above.
**Context info**

juanfont:
Under certain conditions we could get duplicated nodes (e.g., after logout/login, reported in #1054) - due to a misunderstanding from my side on the role of MachineKey in TS2021.

`MachineKey` in TS2021 (the permanent-ish key associated to a client) is no longer sent in the body of the protocol requests, but obtained from the method `Peer()` in the `controlbase.Conn` struct. I was not aware of this (cheers @awsong for the heads up!). Our TS2021 was completely dropping `MachineKey`

Fixes #1054

### DIFF
--- a/integration/auth_web_flow_test.go
+++ b/integration/auth_web_flow_test.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"net/netip"
 	"net/url"
 	"strings"
 	"testing"
@@ -59,7 +60,6 @@ func TestAuthWebFlowAuthenticationPingAll(t *testing.T) {
 	}
 
 	success := 0
-
 	for _, client := range allClients {
 		for _, ip := range allIps {
 			err := client.Ping(ip.String())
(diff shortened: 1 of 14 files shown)


awsong:
Cheers to you @juanfont . 

Thanks for the quick fix. Looking forward to trying it.

awsong:
I tested lates Linux Tailscale client with your branch [machinekey-in-noise](https://github.com/juanfont/headscale/tree/machinekey-in-noise). I could only login when Headscale DB is empty. Once I login and logout, next login will fail.

juanfont:
@awsong can you describe the steps you followed? I am not being able to replicate your issue.

awsong:
> @awsong can you describe the steps you followed? I am not being able to replicate your issue.
Here's my steps:
Server side:
1. `git checkout origin/machinekey-in-noise`
2. `go build -v cmd/headscale/headscale.go`
3. `rm /var/lib/headscale/db.sqlite`
4. `./headscale serve`

Client side:
1. Compile latest linux CLI from source
2. `rm /var/lib/tailscale/ -rf`
3. `./tailscaled`
4. `./tailscale up --login-server http://raz.3fire.org:444 --reset` , you can actually connect this server, user name `test`, password `VGVzdDEyMzQ1QA==` (please base64 decode it)
5. `./tailscale logout`
6. `./tailscale up --login-server http://raz.3fire.org:444 --reset`

juanfont:
@awsong I have managed to replicate it! 

Looking at it...

awsong:
> @awsong I have managed to replicate it!
> 
> Looking at it...

Tested your latest fix, now it works fine. Logout/login multiple times, IP stays the same, Headscale reports same machine key and different node key each time.

juanfont:
@awsong thanks for checking out!  It turned out to be a little bit harded to fix than I thought :) 

I am now finishing integration tests for this, and then I will send it to @kradalby for review.

evenh:
Looking forward to this fix 🙌🏻

integration/auth_web_flow_test.go
```diff
@@ -59,7 +60,122 @@ func TestAuthWebFlowAuthenticationPingAll(t *testing.T) {
...
+	for _, client := range allClients {
+		ips, err := client.IPs()
+		if err != nil {
+			t.Errorf("failed to get IPs for client %s: %s", client.Hostname(), err)
+		}
+		clientIPs[client] = ips
+	}
+
+	for _, client := range allClients {
+		_, _, err = client.Execute([]string{"tailscale", "logout"})
```
kradalby, line 144: Maybe make a nice `.Logout()` for this

kradalby (dismissed):

kradalby (approved):

Issue #1050, opened 2022-12-06
Suspected bug with ip_prefixes


mhameed:
Hi,

Thanks for your work on headscale.
 
I would be grateful if someone could reproduce the the following, which seems like a bug to me:

headscale v0.17.0
tailscale v1.34.0
Ubuntu 22.04
 

Following headscale config file:

 
        ---
        noise:
          private_key_path: /var/lib/headscale/noise.key
        
        # The url clients will connect to.
        # Typically this will be a domain like:
        #
        # https://myheadscale.example.com:443
        #
        server_url: https://test1.ephemeral.host
        
        # Address to listen to / bind to on the server
        #
        listen_addr: 0.0.0.0:443
        
        # Address to listen to /metrics, you may want
        # to keep this endpoint private to your internal
        # network
        #
        metrics_listen_addr: 0.0.0.0:9090
        
        # Address to listen for gRPC.
        # gRPC is used for controlling a headscale server
        # remotely with the CLI
        # Note: Remote access _only_ works if you have
        # valid certificates.
        grpc_listen_addr: 0.0.0.0:50443
        
        # Allow the gRPC admin interface to run in INSECURE
        # mode. This is not recommended as the traffic will
        # be unencrypted. Only enable if you know what you
        # are doing.
        grpc_allow_insecure: false
        
        # Private key used encrypt the traffic between headscale
        # and Tailscale clients.
        # The private key file which will be
        # autogenerated if it's missing
        private_key_path: /var/lib/headscale/private.key
        
        # List of IP prefixes to allocate tailaddresses from.
        # Each prefix consists of either an IPv4 or IPv6 address,
        # and the associated prefix length, delimited by a slash.
        ip_prefixes:
          #- fd7a:115c:a1e0::/48
          - 172.19.0.0/23
          #- 100.64.0.0/16
        
        # DERP is a relay system that Tailscale uses when a direct
        # connection cannot be established.
        # https://tailscale.com/blog/how-tailscale-works/#encrypted-tcp-relays-derp
        #
        # headscale needs a list of DERP servers that can be presented
        # to the clients.
        derp:
          server:
            # If enabled, runs the embedded DERP server and merges it into the rest of the DERP config
            # The Headscale server_url defined above MUST be using https, DERP requires TLS to be in place
            enabled: true
        
            # Region ID to use for the embedded DERP server.
            # The local DERP prevails if the region ID collides with other region ID coming from
            # the regular DERP config.
            region_id: 999
        
[...]

PR #1068, opened 2022-12-15
Add test to reproduce #1050

kradalby:


kradalby:
This is probably a wontfix for now, see #1135.

PR #1076, opened 2022-12-20
fix(routes): ensure routes are correctly propagated

### LINKED ISSUE #1075: Routes are not propagated in v0.18.0-beta1
Hello,

**Bug description**

It seems routes are not correctly propagated when enabling or disabling them through the CLI.
Encountered that while trying https://github.com/juanfont/headscale/pull/1024

**To Reproduce**

Advertise the same route from multiple nodes (to test HA failover). Enable or disable these routes.
Route does not appear in "ip route show table 52". It appears only if I restart tailscale on the node.

**Context info**

Headscale version: v0.18.0-beta1
Tailscale version: v1.34.1
OS: Ubuntu 22.04

fatih-acar:
When using Tailscale v1.34.1, enabling or disabling a route does not effectively add or remove the route from the node's routing table. We must restart tailscale on the node to have a netmap update.

Fix this by refreshing last state change so that a netmap diff is sent.

Also do not include secondary routes in allowedIPs, otherwise secondary routes might be used by nodes instead of the primary route.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #1075

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -675,18 +675,13 @@ func (h *Headscale) toNode(
 		[]netip.Prefix{},
 		addrs...) // we append the node own IP, as it is required by the clients
 
-	enabledRoutes, err := h.GetEnabledRoutes(&machine)
-	if err != nil {
-		return nil, err
-	}
-
-	allowedIPs = append(allowedIPs, enabledRoutes...)
-
 	primaryRoutes, err := h.getMachinePrimaryRoutes(&machine)
 	if err != nil {
 		return nil, err
 	}
 	primaryPrefixes := Routes(primaryRoutes).toPrefixes()
+
+	allowedIPs = append(allowedIPs, primaryPrefixes...)
 
 	var derp string
 	if machine.HostInfo.NetInfo != nil {
@@ -1057,6 +1052,7 @@ func (h *Headscale) EnableRoutes(machine *Machine, routeStrs ...string) error {
 		}
 	}
 
+	h.setLastStateChangeToNow()
 	return nil
 }
 
--- a/routes.go
+++ b/routes.go
@@ -215,6 +215,7 @@ func (h *Headscale) handlePrimarySubnetFailover() error {
 		log.Error().Err(err).Msg("error getting routes")
 	}
 
+	routesChanged := false
 	for pos, route := range routes {
 		if route.isExitRoute() {
 			continue
@@ -235,6 +236,7 @@ func (h *Headscale) handlePrimarySubnetFailover() error {
 					return err
 				}
 
+				routesChanged = true
 				continue
 			}
 		}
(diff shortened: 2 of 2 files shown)


routes.go
```diff
@@ -309,6 +309,7 @@ func (h *Headscale) handlePrimarySubnetFailover() error {
 		}
 	}
 
+	h.setLastStateChangeToNow()
```
juanfont, line 312: This should only run if subnets were actually failed over.

machine.go
```diff
@@ -675,19 +675,14 @@ func (h *Headscale) toNode(
...
 		addrs...) // we append the node own IP, as it is required by the clients
 
-	enabledRoutes, err := h.GetEnabledRoutes(&machine)
-	if err != nil {
-		return nil, err
-	}
-
-	allowedIPs = append(allowedIPs, enabledRoutes...)
-
 	primaryRoutes, err := h.getMachinePrimaryRoutes(&machine)
```
juanfont, line 678: > Also do not include secondary routes in allowedIPs, otherwise secondary routes might be used by nodes instead of the primary route.

I reckon that is not correct. Primary routes handled by that node are specified in the dedicated 
`PrimaryRoutes` field of `tailcfg.Node` https://github.com/juanfont/headscale/blob/main/machine.go#L741, even if we are including _all_ the enabled routes in `AllowedIPs`.

machine.go
```diff
@@ -1057,6 +1052,7 @@ func (h *Headscale) EnableRoutes(machine *Machine, routeStrs ...string) error {
 		}
 	}
 
+	h.setLastStateChangeToNow()
```
juanfont, line 1055: This looks good.

juanfont (commented):
Thanks for spotting it. 

Added some comments.

routes.go
fatih-acar: Should be fixed.

machine.go
fatih-acar: This is what I assumed first after reading the structure's documentation. It states
```
// PrimaryRoutes are the routes from AllowedIPs that this node
// is currently the primary subnet router for, as determined
// by the control plane. It does not include the self address
// values from Addresses that are in AllowedIPs.
PrimaryRoutes []netip.Prefix `json:",omitempty"`
```

But it seems this field is used as an informative one so that subnet routers know that they are the primary router of an HA route. It is not used to populate the routing table of other peers... `AllowedIPs` is used for that.
I don't know Tailscale's internals as much as you, so I will let you double-check.

juanfont:
You have a very good point! Otherwise old clients (pre-subnet HA) would not work...

routes.go
juanfont: Thanks!

juanfont:
thanks!!

PR #1101, opened 2022-12-27
domains/restricted_nameservers: check dnsConfig.Resolvers instead

johnae:
With this change, domains are applied if there are any `dnsConfig.Resolvers` rather than any `dnsConfig.Nameservers`. Without this change, using nextdns doh would disallow setting any domains (because it checked `dnsConfig.Nameservers` which would be empty in the case of NextDNS DOH).

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


juanfont (approved):
Thanks!

Issue #804, opened 2022-09-14
tailscale GUI on macOS doesn't display exit nodes from headscale


nsauk:
**Bug description**

I use headscale as a control server and an official tailscale client from the App Store as a client, and I can't choose exit node from the client's GUI, list of available exit nodes is empty.
<img width="455" alt="image" src="https://user-images.githubusercontent.com/3472119/190027761-74035e06-46f1-4fce-a7de-84cd92a454ec.png">

At the same time, I can see the list of exit nodes available and can choose an exit node using CLI:
```bash
/Applications/Tailscale.app/Contents/MacOS/Tailscale up --login-server=https://my-headscale --exit-node=apx
```
As a result, it works great! But in GUI I can see only the chosen exit node and it shown as offline, despite it works correctly:
<img width="455" alt="image" src="https://user-images.githubusercontent.com/3472119/190024880-f7ebb55c-1bac-4922-a83c-0cb4c382bf08.png">
I reported this to tailscale/tailscale#5628 but they [suppose that there is a problem in control server](https://github.com/tailscale/tailscale/issues/5628#issuecomment-1246061076):
> I bet Headscale isn't setting the online/offline bits or something.

**To Reproduce**

1. Install and configure headscale on a remote server (this step may be omitted, please [contact me](mailto:me@nsauk.in) if you'd like to reproduce this using my headscale instance).
2. Install a Tailscale client on macOS machine.
3. Configure the Tailscale client using .mobileprofile or a CLI command:
```bash
defaults write io.tailscale.ipn.macos ControlURL https://my-headscale
```
4. Login to control server and ensure you can see enabled exit nodes using a CLI command:
```bash
/Applications/Tailscale.app/Contents/MacOS/Tailscale status
```
5. Check the list of exit nodes in GUI.

**Actual result**
List in GUI is empty.

**Expected result**
List in GUI contains the same exit nodes as in console.

**Context info**

Control server: headscale v0.16.4 on CentOS Linux 8
Client: tailscale 1.30.1, App Store version on macOS 12.5.1
Kernel version: 4.18.0-305.12.1.el8_4.x86_64

[headscale and caddy configs](https://gist.github.com/nsauk/b95b748535b4c0879c6df319ecbbf544)

juanfont:
@nsauk hey, can you try to run 0.17.0-alpha2?

nsauk:
@juanfont issue still reproduced on 0.17.0-alpha2.

kazauwa:
[...]

PR #1104, opened 2022-12-31
Fix exit nodes funcionality

juanfont:
Just a small PR  to send in AllowedIPs both primary routes AND enabled exit routes


juanfont:
Also fixes https://github.com/juanfont/headscale/issues/804

kradalby (approved):

PR #1107, opened 2023-01-02
use smallzstd and sync pool

### LINKED ISSUE #1014: Expotential CPU usage from allowed peer checks
**Bug description**

CPU usage grows exponentially as the number of peers grows, to the point where headscale cannot respond to updates fast enough for clients to remain connected. 

This appears to be due to recalculating allowable peers for every update, which is an O(n) operation, for n peers. The allowed peer list should be static except when new peers are added, so updating the peer list once for each new peer would be a huge performance win. 

Enabling ACLs makes this worse because there is more work per peer to check if it's valid, but the namespace only checks do eventually cause performance issues too with 1000s of peers. 

**To Reproduce**

Create a network with 400-600 peers, the exact number where the performance curve becomes a problem depends on the system specs, but with a 4 core server 600 is usually enough to overwhelm the system. 

**Context info**

<!-- Please add relevant information about your system. For example:
- Headscale: 0.16.0, 0.17.0-beta4
- Tailscale: 1.32.3
- Ubuntu 20.04, 22.04

jsiebens:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

This PR tweaks the zstd encoding by using smallzstd package from tailscale and a sync pool, instead of creating a new encoder.
While this won't fix #1014 entirely, it could reduce CPU usage when sending mapresponses to the clients

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
 
           # When updating go.mod or go.sum, a new sha will need to be calculated,
           # update this if you have a mismatch after doing a change to thos files.
-          vendorSha256 = "sha256-SuKT+b8g6xEK15ry2IAmpS/vwDG+zJqK9nfsWpHNXuU=";
+          vendorSha256 = "sha256-8p5NFxXKaZPsW4B6NMzfi0pqfVroIahSgA0fukvB3JI=";
 
           ldflags = ["-s" "-w" "-X github.com/juanfont/headscale/cmd/headscale/cli.Version=v${version}"];
         };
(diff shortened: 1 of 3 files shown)


juanfont:
Hi @jsiebens, first and foremost... thanks! Very happy to see you around here :)

Can you have a look into the linting/build errors? Thanks!

juanfont:
@jsiebens the sha256 sum in `vendorSha256` of `flake. nix` needs to be updated:

```
error: hash mismatch in fixed-output derivation '/nix/store/m9mr72wj7hb0qwgjf3hv7gaxvld4hajb-headscale-48e2a82-go-modules.drv':
         specified: sha256-SuKT+b8g6xEK15ry2IAmpS/vwDG+zJqK9nfsWpHNXuU=
            got:    sha256-8p5NFxXKaZPsW4B6NMzfi0pqfVroIahSgA0fukvB3JI=
```

jsiebens:
ok, I'll try to find the time to set up a nix environment, never done that before :-)

juanfont (approved):
Awesome! Thanks :)

PR #1108, opened 2023-01-02
Remove Tailscale v1.18.2 from test matrix

evenh:
This PR removes Tailscale v1.18.2 from the test matrix, as it seems to fail downloading from the apt package repository.


juanfont (approved):
good catch! thanks!

PR #1127, opened 2023-01-10
oidc: allow reading the client secret from a file

motiejus:
Currently the most "secret" way to specify the oidc client secret is via an environment variable `OIDC_CLIENT_SECRET`, which is problematic[1]. Lets allow reading oidc client secret from a file. For extra convenience the path to the secret will resolve the environment variables.

[1]: https://systemd.io/CREDENTIALS/

- [ ] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md


config.go
```diff
@@ -528,6 +532,25 @@ func GetHeadscaleConfig() (*Config, error) {
...
+		secretFile, err := os.Open(os.ExpandEnv(oidcClientSecretPath))
+		if err != nil {
+			return nil, err
+		}
+		defer secretFile.Close()
+
+		secretBytes, err := io.ReadAll(secretFile)
+		if err != nil {
+			return nil, err
+		}
```
kradalby, line 550: This can be done in one smooth motion with [`os.ReadFile()`](https://pkg.go.dev/os#ReadFile)

```suggestion
		secretBytes, err := os.ReadFile(os.ExpandEnv(oidcClientSecretPath))
		if err != nil {
			return nil, err
		}
```

I like the use of `os.ExpandEnv`.

docs/oidc.md
```diff
@@ -20,6 +20,10 @@ oidc:
   # Specified/generated by your OIDC provider
   client_id: "your-oidc-client-id"
   client_secret: "your-oidc-client-secret"
+  # alternatively, set `client_secret_path` to read the secret from the file.
+  # It resolves environment variables, making integration to systemd's
+  # `LoadCredential` straightforward:
+  #client_secret_path: "${CREDENTIALS_DIRECTORY}/oidc_client_secret"
```
kradalby, line 26: We also need an entry in `config-example.yaml`.

docs/oidc.md
```diff
@@ -20,6 +20,10 @@ oidc:
   # Specified/generated by your OIDC provider
   client_id: "your-oidc-client-id"
   client_secret: "your-oidc-client-secret"
+  # alternatively, set `client_secret_path` to read the secret from the file.
```
kradalby, line 23: I want to discuss: Can we just make the "old" option understand paths?

It would be nice not having to deal with having mutual exclusive options and having to deal with errors and user support for those. 

If the old entry is changed to accept a path, and then it can first check if it is a path, and then fall back to the secret?

motiejus: I understand the sentiment, but I can't figure out how to do this.

How can we determine if something is a path? A few ideas:
- "If it looks like a path, it's a path", we may get false positives. Is "supersecret" a path? How about `/&!!.`?
- If we `stat` and the file exists? Sounds good. But if the file does not exist due to a configuration error, we treat the string as a password. But understanding that headscale uses an incorrect OIDC password is very non-trivial.

motiejus:
Updated the PR as per @kradalby 's comments.

I was able to reproduce the integration test failure locally and will investigate it in a couple of days.

motiejus:
I've updated the test and it passed locally.

motiejus:
Apologies for missing to check the linter. Fixed a linter error.

docs/oidc.md
kradalby: I think this is fair. Lets keep it as a separate option for now.

kradalby (approved):

motiejus:
FYI the subsequent change to nixpkgs: https://github.com/NixOS/nixpkgs/pull/217482

Issue #1114, opened 2023-01-03
[OIDC] Allowed_groups directive issue


yaroslavkasatikov:
Hey team,

I faced with the issue while testing new `allowed_groups` directive. 
I have tried to use it with auth0 and Keycloak, but received `unauthorized principal (allowed groups)` and an error in Headscale log.
```
2023-01-03T18:38:46Z ERR authenticated principal not in any allowed groups
```

My oidc config in config.yaml:

Keycloak version:
```
   oidc:
      client_id: headscale
      client_secret: xxxxxxxxxx
      enabled: true
      issuer: http://keycloak.tld/realms/headscale
      only_start_if_oidc_is_available: true
      scope:
      - openid
      - profile
      - email
      strip_email_domain: true
      allowed_groups:
      - /headscale
      
```
auth0.com Version:

```    
  oidc:
      client_id: 7zxh8KmXXXXXXXXXXXX
      client_secret: X-2AYCpXXXXXXXXXXXXXs-od3
      enabled: true
      issuer: https://dev-r4uxxxxxxxxx.us.auth0.com/
      only_start_if_oidc_is_available: true
      scope:
      - openid
      - profile
      - email
      strip_email_domain: true
      allowed_groups:
      - headscale
      - /headscale ### Added just in case if auth0 needs leading slash too
  ```

Some screnshots from Keycloak and auth0:

<img width="1149" alt="image" src="https://user-images.githubusercontent.com/7746585/210446458-7b0e6b8d-d12a-403f-af37-77847155d1ae.png">

<img width="1250" alt="image" src="https://user-images.githubusercontent.com/7746585/210445246-3e28213d-6cc1-4c5a-b60c-7ce936d65a06.png">

**Comment by madjam002** [comment]
This is not a bug and rather a configuration issue.

For keycloak:
- Create your headscale client in keycloak
- Go to client configuration, roles tab
- Add a role (e.g `allow_access`)
- Go to client scopes tab, click the dedicated client scope config, add new mapper, from predefined mappers, choose client roles
- Change client ID to the same as your headscale client ID, change token claim name to `groups`, make sure add to ID token is checked
- Go to groups in Keycloak config, create a new group that will have access to headscale, in group config go to role mapping, select the role that you added to the headscale client. Add your keycloak user to be part of the group.
- Make sure the id token is valid, go back to your headscale client, client scopes tab, click the "evaluate" tab, choose your user from the dropdown, go to "generated ID token" and you should see:
```
  "groups": [
[...]

PR #1128, opened 2023-01-10
Fix SIGSEGV crash related to map of state changes

evenh:
See https://github.com/juanfont/headscale/issues/1114#issuecomment-1373698441 (probably caused by race conditioning)


juanfont (approved):

juanfont:
Great! Thanks!

PR #1129, opened 2023-01-11
Repeat ToStringSlice

QZAiXH:
…e cpu usage

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

I've been watching the cpu situation for the past few days, and I found that the cpu occasionally suddenly rises to 500%, so I used pprof to check and found that ToStringSlice causes the cpu to rise.

After optimizing it, the cpu is still rising occasionally, but not very much.

The final solution should be to use the cache to store the machine information for comparison, and modify the database data and cache information only when there are changes. I'm not familiar with it, so I haven't made the appropriate changes yet.

https://user-images.githubusercontent.com/23068780/210695313-54af7475-d527-4274-b33a-50838c44ed98.svg
https://user-images.githubusercontent.com/23068780/210695393-754f2a55-6894-42ac-8ecc-6294109e2f19.svg

### DIFF
--- a/machine.go
+++ b/machine.go
@@ -194,6 +194,7 @@ func getFilteredByACLPeers(
 	peers := make(map[uint64]Machine)
 	// Aclfilter peers here. We are itering through machines in all namespaces and search through the computed aclRules
 	// for match between rule SrcIPs and DstPorts. If the rule is a match we allow the machine to be viewable.
+	machineIPs := machine.IPAddresses.ToStringSlice()
 	for _, peer := range machines {
 		if peer.ID == machine.ID {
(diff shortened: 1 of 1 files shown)


juanfont (approved):
Looks cool!

juanfont:
This will probably help in #1014.

QZAiXH:
> This will probably help in #1014.

Yes, the 2-core CPU is now driving 550 machines. The cpu usage is currently a third of what it was before.
But this is not the optimal solution, after that I will try to replace gorm's queries with cache.

PR #1144, opened 2023-01-17
Rename [Nn]amespace -> [Uu]ser

### LINKED ISSUE #329: Rename Namespace to User
Depends on the release of #320 and #311, and resolving #326, #327 and #328 (to reduce code needed to be updated).

This is a tracking ticket for what we need to do to achieve a migration from Namespace to User as outlined in #311.

kradalby:
This PR sorts the longstanding annoyance of the concept "Namespace"

closes #329


db.go
```diff
@@ -184,7 +187,9 @@ func (h *Headscale) initDB() error {
 		return err
 	}
 
-	err = db.AutoMigrate(&Namespace{})
+	_ = db.Migrator().RenameTable("Namespace", "User")
```
juanfont, line 190: Should we add a warning message if this migration happens? 

e.g., "Namespaces have been renamed into Users, to better represent Tailscale.com naming".

kradalby: if we migrate, then it is already migrating, a warning is a bit late at that point?

juanfont: I know. It is meant for UX, so they are aware that now Namespaces are Users. 

Not everyone will be reading the changelog I am afraid.

juanfont (approved):

PR #1160, opened 2023-01-22
Add note on Apache as unsupported reverse proxy

juanfont:
Apache overwrites the upgrade header in the WebSockets connections, replacing the custom Tailscale header for the usual `Upgrade`.

This PR adds a small note in the community docs.


hrtkpf:
Hey,

first of all, thanks for your work on Headscale, really appreciate it!

I actually have Headscale running behind an Apache. You can tell it to accept arbitrary `Upgrade` header values, so it will forward the header as-is.
I could open a PR adding my Apache config if you want.

juanfont:
@hrtkpf oh! I based my info on a user from Discord.

Can you send us a PR with your config? I will close this one...

PR #1218, opened 2023-02-16
Fix OpenBSD build docs

mikejsavage:
- OpenBSD 7.2 installs go 1.19 by default
- Doubt you can run nix on OpenBSD so skip the makefile/nix build and just go build directly

(this is only an auxiliary docs change so none of the checklist seems relevant)


juanfont (approved):

juanfont:
Thanks!

Issue #1217, opened 2023-02-15
Documentation issues


Norodix:
**Issue description**

I really like this software! I just set up it on my VPS and I would like to give some feedback on the installation instructions. Issues I ran into that were not very obvious from the documentation.

**To Reproduce**

Install headscale on linux following the docs.

**Context info**

I love the software, and I hope I can help with this feedback.

### File permissions not explained

I think it is not unreasonable to think someone with little to no linux experience would try to set up their own VPN. The docs dont go into detail with what the file permissions need to look like. There is mention of running it with a headscale user but not that this user needs to be able to write to some dirs and read others. Should the config files under /etc be read-only, belong to root or headscale, etc. I think it might be worth mentioning what it should look like.

### Running from the same directory

This might be a bug, but when I started headscale for the first time from the command line, in the new terminal I could not list the users. I could wget the metrics as shown in the docs but I had to cd to `/etc/headscale` to be able to `headscale users list`. If it is not a bug, it should be mentioned explicitly.

### Systemd not printing all errors.

This might be a bug, but when I ran headscale as a systemd service it could not start but in the status it did not show the errors. I had to run it with headscale user to see that I had some permission errors. I dont know why it was not printed to systemd's status.

### Some files need to be created manually

Some files like `/var/run/headscale/headscale.sock` need to be created manually. It is not created by headscale if it does not exist. It might be better to explicitly mention it in the docs.

### How to keep up-to-date

I dont see any mention in the docs how to upgrade between version of headscale. Given how fast it is developed it might be a good idea to add a short guide. I assume back up stuff, replace bin, run again, but I am just guessing.

kradalby:
Hi, it would be great if you could help improve the docs with your feedback :), we are very happy for pull requests to update the docs!

Thanks

Norodix:
I'm happy to do so but to be perfectly honest I'm not sure about some of the points. Should I just create  PR for the ones I'm confident about and leave this issue up for the rest?

kradalby:
Yes that’s better than the current state :) Sent from Proton Mail for iOS On Sat, Feb 18, 2023 at 14:13, Norodix ***@***.***> wrote:  
I'm happy to do so but to be perfectly honest I'm not sure about some of the points. Should I just create  PR for the ones I'm confident about and leave this issue up for the rest?

Norodix:
I opened a PR.

The only missing thing that I found is the general description on how to update headscale when a new version comes out. I'm curious about that myself.

[...]

PR #1221, opened 2023-02-19
Improve docs

Norodix:
Add file permission details to running headscale on linux documentation.
Fix nologin path.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

https://github.com/juanfont/headscale/issues/1217

### DIFF
--- a/docs/running-headscale-linux.md
+++ b/docs/running-headscale-linux.md
@@ -36,14 +36,19 @@ useradd \
 	--home-dir /var/lib/headscale/ \
 	--system \
 	--user-group \
-	--shell /usr/bin/nologin \
+	--shell /usr/sbin/nologin \
 	headscale
 ```
+
+This directory needs to be writable by the user running headscale. If you are not running headscale as root make sure that `/var/lib/headscale` is owned by the headscale user. If you used useradd to create the directory it should be done by default.
 
 4. Create an empty SQLite database:
 
 ```shell
 touch /var/lib/headscale/db.sqlite
+
+# if you created a headscale user make it the owner:
+chown headscale:headscale /var/lib/headscale/db.sqlite
 ```
 
 5. Create a `headscale` configuration:
@@ -51,6 +56,7 @@ touch /var/lib/headscale/db.sqlite
 ```shell
 touch /etc/headscale/config.yaml
 ```
+The config file should be owned by root with read permission for the headscale user. `-rw-r--r-- root root`
 
 It is **strongly recommended** to copy and modify the [example configuration](../config-example.yaml)
 from the [headscale repository](../)
@@ -77,6 +83,13 @@ Verify `headscale` is available:
 ```shell
 curl http://127.0.0.1:9090/metrics
 ```
+
+Verify you can issue commands to `headscale`:
+
+```shell
+headscale users list
+```
+When running headscale as a normal user it is possible that the metrics page is available but you cannot issue commands to headscale. In that case you need to change to the directory from where you started headscale.
 
 8. Create a user ([tailnet](https://tailscale.com/kb/1136/tailnet/)):
 


docs/running-headscale-linux.md
```diff
@@ -78,6 +84,13 @@ Verify `headscale` is available:
 curl http://127.0.0.1:9090/metrics
 ```
 
+Verify you can issue commands to `headscale`:
+
+```shell
+headscale users list
+```
+When running headscale as a normal user it is possible that the metrics page is available but you cannot issue commands to headscale. In that case you need to change to the directory from where you started headscale.
```
almereyda, line 92: Here it is not obvious what the last sentence intends to convey.

Norodix: When I was trying to test my configuration for the first time I ran headscale from `/etc/headscale` as a non-privileged user. Then according to the docs I opened a second terminal (first was busy running headscale) and I wanted to issue a command, `headscale users list`. It did not work. It only works if I `cd /etc/headscale` in the second terminal.
I tested some more and `/etc/headscale` is not the special case, as a normal user any directory where you start headscale is the only one from where you can issue commands to it.

I don't think that this behaviour is intended, and also mostly people run it from systemd or as a root user, neither case has any problems. But I think it is not unreasonable to test your config as a normal user so I think this quirk should be documented. I would have been happy if I didn't have to figure it out myself.

Maybe a senctence like: In that case you need to change to the directory in the second terminal where (you issue the command) to the same directory from where you started headscale in the first terminal.

I'm not a great writer as you can see, if you have a clearer explanation in mind I'm very open to it!

docs/running-headscale-linux.md
```diff
@@ -36,21 +36,27 @@ useradd \
 	--home-dir /var/lib/headscale/ \
 	--system \
 	--user-group \
-	--shell /usr/bin/nologin \
+	--shell /usr/sbin/nologin \
```
kradalby, line 39: Is this true for alle distros? which one is the reference?

docs/running-headscale-linux.md
kradalby: I think the reason behind why you needed to be in the directory is slightly incorrect.

I think that you need to make sure that `/etc/headscale/config.yaml` is reachable for the unprivileged user, it will need to know how to find the headscale socket, which is located in this config.

The error might have been that you could read the config, by being in the same dir, but you could not access the actual directory with the config (`/etc/headscale`).

docs/running-headscale-linux.md
Norodix: I assume for many distributions. This is certainly the case for debian based ones. According to this post for some more too. https://superuser.com/questions/830656/debian-linux-usr-bin-nologin-missing

docs/running-headscale-linux.md
```diff
@@ -36,21 +36,27 @@ useradd \
...
+This directory needs to be writable by the user running headscale. If you are not running headscale as root make sure that `/var/lib/headscale` is owned by the headscale user. If you used useradd to create the directory it should be done by default.
+
 4. Create an empty SQLite database:
 
 ```shell
 touch /var/lib/headscale/db.sqlite
+
+# if you created a headscale user make it the owner:
+chown headscale:headscale /var/lib/headscale/db.sqlite
 ```
```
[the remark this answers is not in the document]
Norodix: That can work too, but I think it is nice that reading through the docs one can see the most important files that headscale interacts with.

Maybe a separate troubleshooting section at the bottom if you dont like cluttering the main description. Something listing the files with the proper permissions could work.

docs/running-headscale-linux.md
Norodix: I just talked to a user on discord. In his case he was running the headscale server as root and was running the headscale users command as root and it did not work, only when he cd-ed to `/etc/headscale`. I don't think permissions could be an issue in this case.... Maybe this should be a separate issue.

tony1661:
I wish these changes were merged as it would have saved me a lot of troubleshooting!

docs/running-headscale-linux.md
tony1661: would it be better to add `$(which nologin)` instead to be compatible across distros?

Norodix: You are right, that does not work but `su -s /bin/bash - headscale` still does with nologin.

kradalby:
Closing this as a lot has changed with the installation method. We appreciate the contribution.

Issue #1228, opened 2023-02-25
Deleting a node does not delete the associated route entry (v0.20.0)


NetworkEngineer509:
**Bug description**
Deleting a node does not delete the associated route entry

**To Reproduce**

1. Register a node with an advertised route
2. Delete the node manually via headscale cli
3. Check the routes, you will see the entry is still there, but the name is blank

**Context info**
- Version of headscale used: v0.20.0

Here is the output of headscale route list showing the issue
![image](https://user-images.githubusercontent.com/34389163/221382751-9762b8a8-8a35-4d6a-ba26-f4774d01496b.png)

iFargle:
Yep, seeing the same issue here.

kradalby:
Has anyone tried with @juanfont fix?

NetworkEngineer509:
> Has anyone tried with @juanfont fix in #1244?

Yes, it's working now. I have deleted the old routes.

PR #1244, opened 2023-03-02
Allow to delete routes

juanfont:
This PR adds support to delete routes from the CLI and the API.

It also deletes the routes of a machine when it is deleted (https://github.com/juanfont/headscale/issues/1228)


machine.go
```diff
@@ -535,6 +535,11 @@ func (h *Headscale) RefreshMachine(machine *Machine, expiry time.Time) error {
 
 // DeleteMachine softs deletes a Machine from the database.
```
kradalby, line 536: What is the difference between hard and soft? why do we have them?

machine.go
```diff
@@ -535,6 +535,11 @@ func (h *Headscale) RefreshMachine(machine *Machine, expiry time.Time) error {
...
 func (h *Headscale) DeleteMachine(machine *Machine) error {
+	err := h.DeleteMachineRoutes(machine)
```
kradalby, line 538: Cant this be done by cascading database deletes?

machine.go
juanfont: Hard actually deletes the entry in the DB, soft just sets the deleted_at field. We do hard when deleting ephemeral nodes.

machine.go
juanfont: If we would use cascade, we would be using soft deleted in the routes https://stackoverflow.com/a/67751408 and I wanted to hard delete them.

kradalby (approved):

Issue #938, opened 2022-11-07
Option to use Username instead of email when logging in via oidc


Polsaker:
From testing on headscale 0.17, the email field is used when logging in via oidc.

In some providers like headscale, the email is not immutable (it can be changed by the user at any time) but the username is not. It would be nice to have a way to make headscale use that field instead of email.

almereyda:
Is this about supporting claims?

- https://openid.net/specs/openid-connect-core-1_0.html#Claims

meson800:
I'm working on an update for #1287 now, looking at how the code has been reorged.

FStelzer:
Thanks for working on this. I think this needs to be a bit more flexible than just choosing between preferred_username & email.
See: https://learn.microsoft.com/en-us/entra/identity-platform/migrate-off-email-claim-authorization
There should probably be some unique user identifier claim (like MS suggests using TID+OID) and additional info about the user (email/username) to make them easier to identify

meson800:
I'll poke around the current auth code, but I think right now there isn't an assumption of a unique user identifier claim in Headscale. Perhaps there's a nice way to link this. The username is nice because you could migrate between OIDC providers technically without recreating users.

I don't know what the best way to handle username updates would be though, as node MagicDNS names are tied to username.

meson800:
I'm considering implementing something similar to Matrix Authentication Service's user attribute mapping, which would make it generalizable:
https://matrix-org.github.io/matrix-authentication-service/setup/sso.html#user-attributes-mapping

xaemiphor:
I'm not sure where is the best place to document an explicit use case, I'm still surprised #1287 is simply closed.

I've just setup headscale backed by LLDAP + Authelia. I'm not pairing any email solutions that would require my users maintain restricted `username@domain.com` in the `email` field of their profile, like we would see if I deployed headscale using Google OIDC. Other applications I'm using with Authelia appear to be importing both the `preferred_username` and `email` fields, the `email` is specifically loaded by applications that expect to be able to send outbound emails.

So with this setup, someone could simply update their email in the LLDAP dashboard to match another user's email username in order to enroll nodes under their user account. IE `user-a/coolkid1992@hotmail.com` and `user-b/coolkid1992@aol.com` would both be interacting with headscale as `coolkid1992`.

[...]

PR #1287, opened 2023-03-26
Allow use of the preferred_username OIDC claim.

meson800:
Previously, Headscale would only use the `email` OIDC claim to set the Headscale user. In certain cases
(self-hosted SSO), it may be useful to instead use the `preferred_username` to set the Headscale username. This also closes the existing issue #938.

This adds a config setting to use this claim instead. The OIDC docs have been updated to include this entry as well. In addition, this adds an Authelia OIDC example to the docs.

I didn't see any existing unit or integration tests for OIDC, so wasn't sure where to add such tests. If I'm wrong, I'm happy to add them!

I did test this against my SSO server, and confirm that the `preferred_username` grant if you set it works as expected.

- [x] raised a GitHub issue or discussed it on the projects chat beforehand (one already exists)
- [ ] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md


kradalby:
I think this looks sensible, the tests for oidc are located here: https://github.com/juanfont/headscale/blob/main/integration/auth_oidc_test.go

Would be good if we could have one.

meson800:
Thanks. It looks like the current integration tests don't explicitly set claims, but the underlying [mockoidc](https://github.com/oauth2-proxy/mockoidc) does let you queue up successive users with potentially different email/preferred_username claims. I'll try my hand at creating an integration test for this.

vbrandl:
Were you able to implement a testcase? I'd love to see this feature since it is blocking me from using OIDC for headscale.

meson800:
@vbrandl I'm working on it, but I'm new to Go and this codebase, and still figuring out how to more efficiently run the integration tests locally without just pushing to Github. The current OIDC tests don't actually check the claim responses at all, so I also need to thread those checks in.

On my stuff, I'm running from these docker images if you want to try them out: https://github.com/meson800/headscale/pkgs/container/headscale

vbrandl:
I just tested your image and it does work as expected, but I'd prefer to run the official image for my infrastructure. So I'll wait until this gets merged.

jonathanspw:
Thank you for this work @meson800

jonathanspw:
I did some testing on this and it appears to work great.  I noticed the email is still cited on the "congrats you've registered your machine" page or whatever it is which might cause some confusion.

That should probably also reflect the username if the setting is enabled.

meson800:
I'll look into changing that welcome page template @jonathanspw  to respect the setting as well. On my local machine I have the mockoidc up and running with new mock users. this will be mergable hopefully within a week or so.

meson800:
@kradalby I added integration tests as requested, so this should be good for review. In sum, this PR includes:

1. Added the option to use the username claim as above, and updated the documentation.
2. Updates to the mockoidc runner; now you can give it a series of users in the format `username_claim <email_claim@example.org>` and have the mockoidc report those users.
3. Using that functionality, there are now integration tests that register clients under `test-user <test-email@example.org>` with the different claim options. I do this in a kind of sketchy way, by checking the FQDN string of `TailscaleClient`s (e.g. makes sure they end in `test-user.headscale.net` or `test-email.headscale.net`, depending on the setting. I picked this method of validating since in the existing `TailscaleClient` interface that `HSIC` implements, there isn't a user field and I didn't want to add them.
4. As suggested by @jonathanspw, I also updated the post-login HTML template to say "Authenticated as `username`, you can now close this window", where `username` is filled in from the preferredUsername or email claim depending on the setting.

Even with a full `nix develop` on both actual Linux and WSL on Windows, I was unable to run one of the go formatters as suggested in the README, and couldn't figure out the path munging needed to install the tool on my own. If I'm failing some format or lint check, I can try to fix on my end if it's easier.

cmd/headscale/cli/mockoidc.go
```diff
@@ -63,6 +65,29 @@ func mockOIDC() error {
 		}
 		accessTTL = newTTL
 	}
+	
+	mockUsers := os.Getenv("MOCKOIDC_USERS")
```
kradalby, line 69: Would be great if this is prefixed with "HEADSCALE_INTEGRATION_"

integration/auth_oidc_test.go
```diff
@@ -36,6 +37,181 @@ type AuthOIDCScenario struct {
...
+
+	oidcConfig, err := scenario.runMockOIDC(defaultAccessTTL, user_str)
+	if err != nil {
+		t.Errorf("failed to run mock OIDC server: %s", err)
+	}
+
+	oidcMap := map[string]string{
+		"HEADSCALE_OIDC_ISSUER":             oidcConfig.Issuer,
+		"HEADSCALE_OIDC_CLIENT_ID":          oidcConfig.ClientID,
+		"CREDENTIALS_DIRECTORY_TEST":        "/tmp",
```
kradalby, line 71: Is this for Headscale or mockoidc? if its headscale, then prefix would be good.

kradalby:
Otherwise, looks great!

cmd/headscale/cli/mockoidc.go
meson800: This is (technically) not just useful for integration tests, as this environment variable controls what mockoidc does if you e.g. set the env variable and ran `headscale mockoidc` yourself. I followed the naming of the other variables in the `mockOIDC` function (`MOCKOIDC_CLIENT_ID`, `MOCKOIDC_CLIENT_SECRET`, and so on). I can still prefix this if desired.

integration/auth_oidc_test.go
meson800: Are you referring to `CREDENTIALS_DIRECTORY_TEST`? I copied this environment map from the TestOIDCAuthenticationPingAll test. For whatever historical reason, that test passes in the client secret with a file instead of directly, which is the reason for the directory. I could simplify if needed (I don't see why `CREDENTIALS_DIRECTORY_TEST` is a separate env variable).

loprima-l:
Isn't it better if we add an internal mail domain for this case, i think this would be easier to deal with in the api and al taht stuff :

ex: 

preferred_username : loprima would make an email like loprima@headscale.sso

It would be more accurate with the Tailscale way of managing sso I guess

meson800:
I think that is something that should be up to your OIDC provider. If
you're self hosting, you can return those types of fake/internal email
addresses with the email grant.

Other self hostable software (Nextcloud, Matrix) deals with the possible
non-uniqueness of the preferred username claim by binding usernames to the
OIDC provider, which is something Headscale doesn't currently do. If
Headscale allows multiple OIDC providers in the future that will have ti be
fixed.

loprima-l:
I agree with that, but aren't we there to implement best practices to follow ? I think using multiple ways of storing a user "id" will be a mess that will complicate the interconnection between Headscale and other pieces of software.

My advice is that we should have one way to identify a user, if we already have email based user distinction, wouldn't it be bad to implement other ways that doesn't fit with the Tailscale way ?

meson800:
The default OIDC settings prior to this PR already strips the domain from
the email claim unless you configure it not to. If you e.g. point Google's
OIDC provider at it, you'll get a Headscale user with @gmail.com removed.
There's still only one type of user ID with or without this PR.

I think there is an advantage for allowing the option of using the
preferred_username claim to create users; this is how OIDC supports letting
SSO users/administrators pick stable usernames despite changing a backing
email address. There is a unique source of truth on which user is which,
the OIDC provider.

I've only used Headscale; can you elaborate on what the Tailscale way to do
this is?

loprima-l:
As you can see here : See here

Tailscale is using the provider domain, I think that's a better way to do it, maybe we must implement this way also in Headscale, but it would be necessary to use a headscale prefix, so we don't break up the compatibility with native Tascale APP, I don't know how they refer to the login methods.

meson800:
Thanks for the screenshot!

My feeling is that I think this probably belongs with multi-provider
support, not in this PR. I could work on it though in a future PR though.

I would imagine that with multiple providers, the headscale admin could
configure a suffix for each OIDC provider. E.g. Github could get a @github
prefix, self-hosted Authelia could get a @internal.sso prefix, whatever.

loprima-l:
I think that would be AMAZING <3 

Like it will help a lot in the infinite quest of having a standardized system, I'm really busy to work on it but if you need help let me know I could write the docs if you want, I'm already planing to open a PR to add a gitbook.

meson800:
I merged against main so this can run tests again.

@kradalby do you want me to change the environment variables as in the review above? I named them to be consistent with the surrounding code.

@juanfont do you want me to update/edit this PR in some way/can you review? I see that there are other OIDC PRs that this might conflict with this (#1435), I can incorporate those changes if needed.

leeaash:
When will this PR be merged to main branch, that feature is acutely I want.

joachimtingvold:
Yes, please. Immutable identifier to be used by headscale would be nice to have, as most IDPs allows users to change their emails, and thus break their headscale setup/login.

meson800:
I can try to update this branch to get it considered again; a lot has changed since I submitted the PR.

jonathanspw:
> I can try to update this branch to get it considered again; a lot has changed since I submitted the PR.

Please do.

kradalby:

meson800:
@kradalby I can't reopen the PR myself, but this is ready to be considered again; I rebased and tested [meson800:oidc_username_claim](https://github.com/meson800/headscale/tree/oidc_username_claim) on top of the re-orged main.

# Summary

- [x] read the [CONTRIBUTING guidelines](https://github.com/juanfont/headscale/pull/README.md#contributing)
- [x] raised a GitHub issue or discussed it on the projects chat beforehand (one already exists)
- [ ] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

This PR adds the ability to use the `preferred_username` claim, closing #938. I added documentation, updated the CHANGELOG.md, and added integration tests; I didn't add a unit test but can if needed.

The short version is that you can now set the `use_username_claim` config setting inside the `oidc` category. If set to true, the `preferred_username` claim is used to set the Headscale username instead of the `email` claim. I updated the way `mockoidc` is called to let integration tests set these claims.

## Integration tests
In particular, the integration tests are implemented by setting the mock OIDC user to `test-user <test-email@example.com>`, then it checks that the FQDN of the user is [`test-user.headscale.net`](https://github.com/meson800/headscale/blob/oidc_username_claim/integration/auth_oidc_test.go#L104) or [`test-email.headscale.net`](https://github.com/meson800/headscale/blob/oidc_username_claim/integration/auth_oidc_test.go#L191).

#### Test failures
Of note, I ran [all of the integration tests](https://github.com/meson800/headscale/pull/1) in a stub PR in my repo; some integration tests appear flaky so I had to rerun them, but one still fails despite reruns, `TestPingAllByIP`. This one also seem to be failing in other PR runs in this repo so I don't think it's related ([example 1](https://github.com/juanfont/headscale/actions/runs/6647003340/job/18061574738))

## Real tests
I generated a Docker image for this PR and tested it out in my tailnet, and it works, correctly using the `preferred_username` claim:
image

LarssonOliver:
Looks awesome! Would really help my auth setup.

Helvio88:
Any reason this wasn't merged? I see it's closed. I think it's a very welcome improvement.

meson800:
I think that one of the concerns is it wasn't generalizable enough. See this comment: https://github.com/juanfont/headscale/issues/938#issuecomment-1952697228

PR #1293, opened 2023-03-27
Build static site with documentation

Orhideous:
This PR builds available docs into static site on Github Pages, [example here](https://orhideous.github.io/headscale/).

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


kradalby:
Could you post some screenshots?

Orhideous:
@kradalby Sure. Also, there is a livv vesion of this documentation, built from fork here: https://orhideous.github.io/headscale

image
image

kradalby:
I think this looks pretty good; two requests;

We need to ensure there is information about adding newly created pages to the configuration for the doc generator

Can something be done about the colour so the logo works better?

Orhideous:
> We need to ensure there is information about adding newly created pages to the configuration for the doc generator

Sure. Right now MkDocs only complains about files, not mentioned in navigation tree, but it could be configured to fail in such cases (`build --strict`)

> Can something be done about the colour so the logo works better?

I used existing logo. Maybe, it worth to create monochrome variant or so.

Orhideous:
> Yes, at least we want the build to fail if someone adds a new .md file under docs/ which is not on the docs page.
Then I need to move `/docs/examples`  somewhere… to `/examples`, say. Is it OK?

> but the orange/red is not ideal

I choose closest one to `hsl(340, 83%, 84%)`. Propose any color scheme which you prefer, I have (almost) no preference — just to avoid bikeshedding. Such kind of tweaks could be done in subsequent PRs, of course.

Orhideous:
@kradalby Done. Right now I used default scheme.

juanfont (approved):

docs/acls.md
```diff
@@ -1,4 +1,15 @@
-# ACLs use case example
+Headscale implements the same policy ACLs as Tailscale.com, adapted to the self-hosted environment.
+
+For instance, instead of referring to users when defining groups you must
+use users (which are the equivalent to user/logins in Tailscale.com).
```
tianon, line 4: Is it intentional that this says "instead of referring to users ... you must use users" ? 😅

PR #1304, opened 2023-03-31
Add tests to verify "Hosts" aliases in ACL

### LINKED ISSUE #941: ACL src/dst with host-alias not working
**Bug description**

Using host alias as dst/destination in ACLs does not work. In this case no node can talk to each other. The command "tailscale status" only shows itself. 
Same applies by using alias in src/source as well.

When using the IP instead host alias, the ACL works.

**To Reproduce**

A sample ACL which is not working:

```
{
    "hosts": {
        "client1": "100.64.0.1",
        "client2": "100.64.0.2",
        "server":  "100.64.0.4",
    },

    "acls": [
        {
            "action": "accept",
            "src": [
                "*",
            ],
            "dst": [
                "server:22,80",
            ],
        },
    ],
}
```

**Context info**

- Version of headscale used: 0.17.0-beta2
- Version of tailscale client: 1.32.2 (Windows), 1.32.0 (Linux)
- OS (e.g. Linux, Mac, Cygwin, WSL, etc.) and version: Windows 10, Linux DietPi v8.9.2
- Kernel version: 5.15.61-v8+
- The relevant config parameters you used: see above
- Log output: no

kradalby:
This should fix #941.

### DIFF
--- a/acls.go
+++ b/acls.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/tailscale/hujson"
+	"go4.org/netipx"
 	"gopkg.in/yaml.v3"
 	"tailscale.com/envknob"
 	"tailscale.com/tailcfg"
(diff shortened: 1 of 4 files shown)


juanfont (approved):
lgtm!

PR #1333, opened 2023-04-09
Add unix_socket directive to the "likely wanted changes" area

qub:
The default config places the control socket in /var/lib/headscale/headscale.sock, causing an error. As noted for the other files this area is not writable in the docker image. I also fixed a double space while I was at it.

I'm not sure if protocol is to also create an issue and link it to this, but please let me know and I'll do that if it's needed.

- [ ] read the [CONTRIBUTING guidelines](README.md#contributing)
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/docs/running-headscale-container.md
+++ b/docs/running-headscale-container.md
@@ -61,9 +61,11 @@ private_key_path: /etc/headscale/private.key
 # The default /var/lib/headscale path is not writable in the container
 noise:
   private_key_path: /etc/headscale/noise_private.key
-# The default /var/lib/headscale path is not writable  in the container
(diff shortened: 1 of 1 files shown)


kradalby:
Hi @qub, I think the more correct way to do this would be to create a mount for the socket, not to put it elsewhere, if not, you wont be able to access the headscale instance with the CLI outside of the container.

What the docs should probably say is to create a mount from `/var/run/headscale` to `/var/run/headscale` in the container.

kradalby:

PR #1338, opened 2023-04-13
Port embedded DERP integration tests to new model

juanfont:
This PR ports the DERP integration tests to the new scenario model, in preparation to expand the capabilities of the DERP server (i.e., verify clients).

### DIFF
--- a/integration/hsic/hsic.go
+++ b/integration/hsic/hsic.go
@@ -15,6 +15,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
@@ -23,6 +24,7 @@ import (
 	"github.com/juanfont/headscale/integration/dockertestutil"
 	"github.com/juanfont/headscale/integration/integrationutil"
 	"github.com/ory/dockertest/v3"
+	"github.com/ory/dockertest/v3/docker"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


integration/tsic/tsic.go
```diff
@@ -555,6 +575,62 @@ func (t *TailscaleInContainer) Ping(hostnameOrIP string, opts ...PingOption) err
...
+
+		if strings.Contains(result, "is local") {
+			return nil
+		}
+
+		if !strings.Contains(result, "pong") {
+			return backoff.Permanent(errTailscalePingFailed)
+		}
+
+		if !strings.Contains(result, "via DERP") {
```
kradalby, line 617: Taking a second look at this function, it seems unnecessary, we can achieve the same by using `WithPingUntilDirect` and wrapping this if check in `if !args.direct` in the original function

integration/hsic/hsic.go
```diff
@@ -108,6 +112,19 @@ func WithPort(port int) Option {
...
 }
 
+// WithExtraPorts exposes additional ports on the container (e.g. 3478/udp for STUN)
+func WithExtraPorts(ports []string) Option {
+	return func(hsic *HeadscaleInContainer) {
+		hsic.extraPorts = ports
+	}
+}
+
+func WithHostPortBindings(bindings map[string][]string) Option {
```
kradalby, line 122: We cannot reintroduce HostBindings, it breaks in subtle ways and it prevents us from running tests in parallel.

integration/tsic/tsic.go
juanfont: I agree :)

kradalby (approved):

PR #1348, opened 2023-04-19
refactor: Support any PostgreSQL database

ayush6624:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

We wanted to use CockroachDB with headscale, which was not possible with the current config options. This PR adds support to directly input a Postgres dialect URL

### DIFF
--- a/app.go
+++ b/app.go
@@ -122,6 +122,11 @@ func NewHeadscale(cfg *Config) (*Headscale, error) {
 	var dbString string
 	switch cfg.DBtype {
 	case Postgres:
+		if cfg.DBurl != "" {
+			dbString = cfg.DBurl
+			break
+		}
+		
 		dbString = fmt.Sprintf(
 			"host=%s dbname=%s user=%s",
 			cfg.DBhost,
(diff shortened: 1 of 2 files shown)


kev-the-dev:
Just curious, what options did cockroachDB require that aren't currently provided? What does the db url string you use look like?

app.go
```diff
@@ -122,6 +122,11 @@ func NewHeadscale(cfg *Config) (*Headscale, error) {
 	var dbString string
 	switch cfg.DBtype {
 	case Postgres:
+		if cfg.DBurl != "" {
```
kev-the-dev, line 125: Any particular reason for this to only apply to Postgres? This seems like this could be a generic feature to override DBURL for any DBType

config.go
```diff
@@ -61,6 +61,7 @@ type Config struct {
 	DBuser string
 	DBpass string
 	DBssl  string
+	DBurl  string
```
kev-the-dev, line 64:
```suggestion
        // DBurl explicitly sets the url passed to the database driver
        // If set, DBpath, DBhost, DBport, DBname, DBuser, DBpass, DBssl is ignored
	DBurl  string
```

kradalby:

Issue #1372, opened 2023-04-22
Headscale does not start with large networks in ACLs


pkrivanec:
Hi,

headscale v0.22.1 does not work with large networks (10.0.0.0/8) in ACLs.
The v0.21.0 version had no problem with this configuration.

**Bug description**

Headscale does not start correctly and uses 100% CPU and memory.

**To Reproduce**
Use acl.hujson:

    {
        "groups": {
            "group:test": [
                "testuser"
            ]
        },
        "acls": [
            {
                "action": "accept",
                "src": [
                    "group:test"
                ],
                "dst": [
                    "tag:exitnode:*",
                    "10.0.0.0/8:*"
                ]
            }
        ]
    }

**Context info**
- Version of headscale used 0.22.1
- OS Rocky Linux release 9.1 
- Kernel version 5.14.0-162.23.1.el9_1.x86_64

Chluz:
Same issue here.

pkrivanec:
Hi,

I started debugging and found a possible problem with the generation of the ACLPeerCacheMap.

The generateACLPeerCacheMap function in acls.go creates a struct for each destination IP.
This will not scale for large network ranges.

codyro:
This unfortunately is a blocker for our organization too.

kradalby:
I have rewritten this whole thing in https://github.com/juanfont/headscale/pull/1381, this is an unoptimised, but "way smarter and cleaner"  solution, I have added @Chluz case as a test case so we don't regress.

The initial version was a lot of cladding on a suboptimal temp solution, but this should be closer to something we can live with. I _think_ it should not _need_ optimalisations before it's useful, but I'll look into that as part of ACL refactoring.

The previous few PRs was mostly to get tests that we could be confident in so we now can refactor it without too much worry. I appreciate that it lead to one really bad decision, but now we can write a faster solution without breaking things.

Would be great if someone tests #1381 while I am working on it, and apologies for the inconvenience.

kradalby:
resolved in next release

PR #1377, opened 2023-04-24
optimize generateACLPeerCacheMap

pkrivanec:
Hi,

this is a possible optimization for the problem in issue #1372.
Not the most elegant solution but it should work.


jonathanspw:
Added this patch into the RPMs in 0.22.1-2.

codyro:
I've tested both the proposed changes and @jonathanspw RPM. This seems to resolve this issue.

kradalby (approved):

PR #1239, opened 2023-02-28
Optimize getFilteredByACLPeers

QZAiXH:
- [x] read the [CONTRIBUTING guidelines](README.md#contributing)
- [ ] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated documentation if needed
- [ ] updated CHANGELOG.md

Use map to speed up lookups. This can effectively reduce cpu usage

PR #1381, opened 2023-04-26
ACL refactor

kradalby:
General QoL improvements to the ACL code, naming, splitting up larger functions, some more tests at different layers.

Rewrite of the filtering of peers served as part of MapResponses, fixing #1372, undoing #1239, #1377, redoing #320.

Small contribution to laying the groundwork for improving SSH support.


acls.go
```diff
@@ -820,8 +682,255 @@ func expandGroup(
...
+
+type Match struct {
+	Srcs  []netip.Prefix
+	Dests []netip.Prefix
+}
+
+func MatchFromFilterRule(rule tailcfg.FilterRule) Match {
+	match := Match{
+		Srcs:  []netip.Prefix{},
+		Dests: []netip.Prefix{},
```
kradalby, line 896: Replace with [IPSet](https://pkg.go.dev/go4.org/netipx#IPSetBuilder)

ventsislav-georgiev:
@kradalby Is there any chance of an image pushed to a public registry for trying out the changes directly?

kradalby:
@ventsislav-georgiev sorry, its being continuously updated, so its not really feasible with our current release pipeline, we dont support or build docker officially. There is a binary attached to the build job on this PR.

loprima-l:
Hi, about the potential ACL API, what is the current plan ?

kradalby:
@loprima-l no plan, I have no intention of implementing an API for ACLs.

The current state is that some people want it, and have discussed how to do it, and we will accept an implementation that fits into the project. But it is currently not part of our priorities, and I do not think this will change for a long time.

juanfont (approved):

PR #1387, opened 2023-04-28
Switch docker release to goreleaser

kradalby:
WIP PR to see if we can switch away from the complicated docker image build to using goreleaser.


.goreleaser.yml
```diff
@@ -80,6 +80,36 @@ nfpms:
...
+    platforms:
+      - linux/amd64
+      - linux/386
+      - linux/arm64
+      - linux/arm/v7
+      - linux/arm/v6
+      - linux/arm/v5
+    tags:
+      - latest
+      - '{{.Tag}}'
```
ohdearaugustin, line 97: Add all the tags we have now to not break it.
```yaml
- '{{ .Major }}.{{ .Minor }}'
- '{{ .Major }}'
```

PR #1390, opened 2023-04-28
Send peer deltas in MapResponse

### LINKED ISSUE #1383: FR: Send Peers delta in MapResponse
**Feature request**

**I'm currently working on this feature and will submit my patch upstream, but creating a ticket to prevent duplicate work**

This is a ticket to implement the sending of Peers list as deltas rather then the full list each time, per this TODO https://github.com/juanfont/headscale/blob/d0113732fe1a3248456da84259f01967b4d09b49/api_common.go#L72.

In addition to improving performance, there is at least one bug I've found associated with not sending deltas:
* There are two nodes in a tailnet, A and B
* Node B is deleted
* A MapResponse is generated for node A. Node A no longer has any peers, so `len(MapResponse.Peers) == 0`
* Tailscale interprets `len(MapResponse.Peers) == 0` as no change, meaning it isn't notified of node B being deleted https://github.com/tailscale/tailscale/blob/43819309e18ac7d496643fcd86489a2ee0927964/control/controlclient/map.go#L227
* Node B remains on tailnet after being deleted, causing a security vulnerability for deleted nodes

kev-the-dev:
# Summary 
Implement the sending of peer list deltas in MapResponse, rather then sending the full list each time. Newer versions of tailscale support this, but it had not yet been implemented in Headscale and was left as a TODO. Sending peer list deltas reduces data usage of the control protocol, especially for larger tailnets. 

This also fixes a known bug described in #1383, where a peer would not be removed from the network if it is deleted and there are only 2 peers in the network. 

This is implemented in a way intended to later be expanded to other deltas being sent in MapResponse. It introduces a new struct `mapResponseStreamState` that is forwarded to sequential calls to generate a MapResponse. Over time, this struct can evolve to track more state to move more of the MapResponse to supporting the deltas.

Closes #1383

# Testing
This was tested locally by creating a small tailnet with 2 peers, and deleting one to verify that the other peer is notified of the deleted peer.

CI unit and integration tests should give reasonable assurance of no other regressions with this change.

### DIFF
--- a/hscontrol/api_common.go
+++ b/hscontrol/api_common.go
@@ -7,9 +7,15 @@ import (
 	"tailscale.com/tailcfg"
 )
 
+// generateMapResponse generate a map response message for this machine.
+//
+// streamState is persistent state maintained across a stream of mapResponse messages.
+// For first message in stream, send the address of a zero-initialized struct,
+// then send the same address in subsequent messages to enable sending of delta-only messages.
 func (h *Headscale) generateMapResponse(
 	mapRequest tailcfg.MapRequest,
 	machine *Machine,
+	streamState *mapResponseStreamState,
 ) (*tailcfg.MapResponse, error) {
 	log.Trace().
 		Str("func", "generateMapResponse").
@@ -39,17 +45,6 @@ func (h *Headscale) generateMapResponse(
 
 	profiles := h.getMapResponseUserProfiles(*machine, peers)
 
-	nodePeers, err := h.toNodes(peers, h.cfg.BaseDomain, h.cfg.DNSConfig)
-	if err != nil {
-		log.Error().
-			Caller().
-			Str("func", "generateMapResponse").
-			Err(err).
-			Msg("Failed to convert peers to Tailscale nodes")
-
-		return nil, err
-	}
-
 	dnsConfig := getMapResponseDNSConfig(
 		h.cfg.DNSConfig,
 		h.cfg.BaseDomain,
@@ -65,17 +60,6 @@ func (h *Headscale) generateMapResponse(
 
 		// TODO: Only send if updated
 		DERPMap: h.DERPMap,
-
-		// TODO: Only send if updated
-		Peers: nodePeers,
-
-		// TODO(kradalby): Implement:
-		// https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L1351-L1374
-		// PeersChanged
-		// PeersRemoved
-		// PeersChangedPatch
-		// PeerSeenChange
-		// OnlineChange
 
 		// TODO: Only send if updated
 		DNSConfig: dnsConfig,
@@ -103,6 +87,20 @@ func (h *Headscale) generateMapResponse(
 		},
 	}
 
+	toNodes := func(machines Machines) ([]*tailcfg.Node, error) {
+		return h.toNodes(machines, h.cfg.BaseDomain, h.cfg.DNSConfig)
+	}
+	resp, err = applyMapResponseDelta(resp, streamState, peers, toNodes)
+	if err != nil {
+		log.Error().
+			Caller().
+			Str("func", "generateMapResponse").
+			Err(err).
+			Msg("Cannot apply map response deltas")
+
+		return nil, err
+	}
+
 	log.Trace().
 		Str("func", "generateMapResponse").
 		Str("machine", mapRequest.Hostinfo.Hostname).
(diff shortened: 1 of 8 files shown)


machine.go
```diff
@@ -588,6 +597,8 @@ func (h *Headscale) DeleteMachine(machine *Machine) error {
 		return err
 	}
 
+	h.setLastStateChangeToNow()
```
kev-the-dev, line 600: I'm not so sure about this change, but I was surprised to see `setLastStateChangeToNow` called in all other changes to machines, but not in `DeleteMachine`

api_common.go
```diff
@@ -7,9 +7,22 @@ import (
 	"tailscale.com/tailcfg"
 )
 
+// mapResponseStreamState tracks state associated with a stream of MapResponse messages,
+// which may optionally send only deltas from the previous message
```
loprima-l: Add a point to complete the lint check ^^

api_common.go
```diff
@@ -7,9 +7,22 @@ import (
...
+type mapResponseStreamState struct {
+	// peersByID is the peers sent in the last stream message,
+	// for comparison in generating deltas in the new message
+	peersByID map[uint64]Machine
+}
+
+// generateMapResponse generate a map response message for this machine
+//
+// streamState is optional persistent state maintained across a stream of mapResponse messages.
+// send the same streamState in each message to enable sending of delta-only messages
```
loprima-l: Same here

api_common.go
```diff
@@ -103,6 +94,62 @@ func (h *Headscale) generateMapResponse(
...
+				Str("func", "generateMapResponse").
+				Err(err).
+				Msg("Failed to convert peersChanged to Tailscale nodes")
+
+			return nil, err
+		}
+		resp.PeersChanged = nodesChanged
+
+		// Update PeersRemoved with any peers which are no longer present
+		for id, _ := range streamState.peersByID {
```
loprima-l: Obscur lint error, potential fix : `$ gofumpt -w api_common.go`

machine.go
```diff
@@ -88,6 +88,15 @@ type (
 	MachinesP []*Machine
 )
 
+// machinesByID converts a slice of machines to a map from node id to machine
+func machinesByID(machines Machines) map[uint64]Machine {
+	byID := make(map[uint64]Machine)
+	for _, machine := range machines {
+		byID[machine.ID] = machine
+	}
+	return byID
```
loprima-l, line 98: Add a blank line before the return here ^^

machine.go
```diff
@@ -88,6 +88,15 @@ type (
 	MachinesP []*Machine
 )
 
+// machinesByID converts a slice of machines to a map from node id to machine
```
loprima-l: Add a dot at the end of the comment

loprima-l (changes requested):
Nice job here ! I added some comments to help you pas the tests. Your only real mistake here is forgetting dots at the end of your comments. (I think lint don't like it, but we do xD)

loprima-l (commented):
My bad GIT didn't show me the lint commit at first, thanks for your implication

machine.go
kradalby: Yes makes sense, it kind of highlight how fragile the current "remember to update the state" approach is.

api_common.go
```diff
@@ -103,6 +94,62 @@ func (h *Headscale) generateMapResponse(
 		},
 	}
 
+	// Send only deltas if this is the next map response in a streamed sequence
```
kradalby, line 97: Can you split this out to a function and add some tests?

nil (full update), node added, node removed, no change.

kradalby: please write table driven tests, and do not take inspiration from our old terrible tests :)

kradalby:
Looks sensible 👍

gps949:
just a reminder, need to ensure there's no delta may miss if there's any network unstable.

kradalby:
> just a reminder, need to ensure there's no delta may miss if there's any network unstable.

@gps949, this approach _should_ work as the state that determines if we should send the delta is stored in a function that only lives while the connection is alive.

That said, it would be have released as testing for a while.

api_common.go
kev-the-dev: Sounds good... can you point me in the right direction in terms of where to put this test and and preferred framework for table driven tests? I've only written table driven tests in JS

kradalby: You don't need a framework, but a structure like this should suffice https://github.com/kradalby/headscale/blob/acl-refactor/acls_test.go#L1698C1-L1834

juanfont:
@kev-the-dev cna you rebase to main?

kradalby (commented):
I think this looks reasonable, I think there is no point in nitting to much on how we want to strucutre the already silly large mapresponse function as I am about to start work on restructuring it to support some needed changes for ACLs.

If  you resolve the few nits, and the tests passes, I'm happy to merge this, and then we can urge users to help test it.

kradalby:
if you can rebase this of master and fix the lint error, that should be good.

kev-the-dev:
Oh I think it's related to the integration test testing multiple versions of tailscale client, some of which don't have support for the map deltas. We'll need to either drop support for these versions or introduce code to do different things depending on client version

kev-the-dev:
When a maintainer gets the chance, I'm hoping the next run of the integration tests with new debug lines will clarify what versions of tailscale this is failing for

kradalby:
> Oh I think it's related to the integration test testing multiple versions of tailscale client, some of which don't have support for the map deltas. We'll need to either drop support for these versions or introduce code to do different things depending on client version

Most of them seem to pass, so that would be strange, I would expect to see more of them failing. It would have to be done with [CapVersion](https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L37)

kev-the-dev:
Given it was 1.36.2 that failed, I think it wasn't an issue with Tailscale versions. My current guess is it's just a timing issue.. the test sleeps for a fixed 30 seconds. Depending on CPU load of CI, etc that may not work. I updated test to explicitly wait for the condition we're looking for.

hscontrol/api_common.go
```diff
@@ -111,3 +108,71 @@ func (h *Headscale) generateMapResponse(
...
+// currentPeers list of peers currently available for the node that this mapResponse is for.
+// toNodes a function to convert the Headscale Machines structure to Tailscale Nodes structure.
+func applyMapResponseDelta(
+	mapResponse tailcfg.MapResponse,
+	streamState *mapResponseStreamState,
+	currentPeers Machines,
+	toNodes func(Machines) ([]*tailcfg.Node, error)) (tailcfg.MapResponse, error,
+) {
+	// Full update
+	if streamState == nil {
```
gps949, line 134: It seems streamState won't be nil when the applyMapResponseDelta called first time in the "case <-updateChan" branch. But there's no peers has been stored in streamState.peersByID. So that may cause the 2nd map response send as delta but generated from an empty state. May that cause some duplicated peers and some peers won't be remove forever in this poll stream?

kev-the-dev: Ah, I think you're right! Good find. This could explain the integration test failure

kev-the-dev (commented):
@kradalby sorry for all the churn... I believe I now fixed the integration test as it passes locally for me

kev-the-dev:
Hm now failures with message:

```
=== FAIL: integration TestTailscaleNodesJoiningHeadcale/create-tailscale (unknown)
2023/05/16 07:38:36 failed to create tailscale node: could not start tailscale container (version: 1.30.2): : The command '/bin/sh -c apt-get update     && apt-get install -y gnupg curl ssh dnsutils ca-certificates     && adduser --shell=/bin/bash ssh-it-user' returned a non-zero code: 100
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xecb78b]
```

I can't imagine this being related to the PR... perhaps a re-run?

kradalby:
It is not the best timing, we started quite a large code reorg, and we will likely not cut a new version for quite some time. We will have to discuss it a bit. 

The second part of the code reorg is in https://github.com/juanfont/headscale/pull/1458, and that brings in the biggest changes. However, the changes touches so much of the code that I suspect we will run it as beta for quite some time, if not alpha.

PR #1395, opened 2023-04-30
Fix: unable to start without bind mount

### LINKED ISSUE #1300: Better Docker Support PR
**Feature request**

I would like to edit the configuration logic as well as the Dockerfile to make it possible to configure headscale using the environment variables only. I'd need this for my portainer setup.

I'd suggest adding a configuration "SKIP_CONFIG_READ" which would skip checking the config file and instead rely on the environment variables only.

I also think it would be nice to use a double underscore as environment separator, like some other projects like gitea do, as I think that helps differentiate the variables better, but this would break backwards compatibility, so it's probably a bad idea.

If you are ok with my suggestion, I'd get to work on a PR.

Edit: I should probably clarify my intent a bit better. My goal would be to run headscale without any additional configuration as bind mounts so it's easier to deploy. If you'd want this feature for consistency, I could  also autocreate the sqlite db if it hasn't been initialized yet.

acul009:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #1300 - I got no answer so I just made this pull request instead.
If you need anything else before merging this, please tell me.

### DIFF
--- a/config.go
+++ b/config.go
@@ -161,7 +161,7 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetEnvPrefix("headscale")
 	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 	viper.AutomaticEnv()
-	
+
 	viper.SetDefault("skip_config_file", false)
 
 	viper.SetDefault("tls_letsencrypt_cache_dir", "/var/www/.cache")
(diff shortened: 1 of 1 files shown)


babyhuey:
Why did you close this? This is useful for me

Issue #740, opened 2022-08-15
add --verify-clients environment variable to docker version


vampywiz17:
**Feature request**

PLease add option to set private option to embedded DERP server in docker container.

kradalby:
Hi, can you please try to reword and explain this feature request?

christian-heusel:
> .... but we do not officially support the docker setup, so it will not be prioritised for now.

@kradalby this is a general feature request, not specific for docker 😊 
See https://tailscale.com/kb/1118/custom-derp-servers/#optional-restricting-client-access-to-your-derp-node for reference

joejose97:
@juanfont

The tailscale derp server gets the valid client list by sending a GET request to http://local-tailscaled.sock/localapi/v0/status
I'm thinking of emulating what the tailscale client does by creating a http listener on the said socket.

If this is acceptable, I'll open a PR with the same.

icb-:
Emulating the tailscale control socket API doesn't sound very straightforward, and would come with some downsides (like not being able to easily run tailscale on the headscale system).

It may be better to see about factoring out how the DERP server validates node keys (https://github.com/tailscale/tailscale/blob/main/derp/derp_server.go#L1126-L1142) and make it possible to pull those from headscale rather than the tailscale client socket.

Maybe passing a `Verifier` function, rather than setting a boolean for whether to verify?

I don't know if that sort of change could be made in a way that would be accepted upstream, or if it would mean maintaining a fork of the derp server.

mritd:
@icb- After reading the derper code, it is feasible to simulate the API; I have created a [branch](https://github.com/mritd/headscale/tree/feat/derp-verify-client) and tested it; derper successfully completed the verification of the Client.

<img width="2558" alt="image" src="https://github.com/juanfont/headscale/assets/13043245/93251b64-c434-4545-92ca-7ed538467d93">


- 1、Derper's verification of its client is mainly completed through [the node public key in the `status` struct](https://github.com/tailscale/tailscale/blob/f398712c006a8afeb165942baae9e7c526de7a9e/derp/derp_server.go#L1130).
- 2、Node public key is obtained from [local socket api status request](https://github.com/tailscale/tailscale/blob/f398712c006a8afeb165942baae9e7c526de7a9e/ipn/localapi/localapi.go#L105)
- 3、We only need to simulate a local status api and return the public keys of all nodes.
- 4、After setting up client authentication, derper will automatically connect to this API to obtain the public keys of all nodes.

[...]

PR #1400, opened 2023-04-30
PR: Add a config to verify clients on DERP

loprima-l:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md`

Solution to the #740 issue, my objective is to add an option in the config to enable or not the DERP client verification 

TO DO : 

- [ ] Check if native tailscale DERP client verify option works 
- [ ] Create the OPTION
- [ ] Bind the option to the actual DERP client verify function 
- [ ] Update the docs & changelog
- [ ] (if needed) create tests

### DIFF
--- a/derp_server.go
+++ b/derp_server.go
@@ -33,6 +33,9 @@ type DERPServer struct {
 func (h *Headscale) NewDERPServer() (*DERPServer, error) {
 	log.Trace().Caller().Msg("Creating new embedded DERP server")
 	server := derp.NewServer(key.NodePrivate(*h.privateKey), log.Info().Msgf)
+	
+	server.SetVerifyClient(true)
+	
 	region, err := h.generateRegionLocalDERP()
 	if err != nil {
 		return nil, err


juanfont:
Hi, this will not work.

The derper server requires to talk to a local tailscale client daemon to check if the DERP clients are part of the same tailnet.

We would need to replicate the local API of tailscale - or, more easily, check if each request to the DERP server comes from one of the endpoints we are aware of from the headscale clients.

loprima-l:
Thanks for the advice, I guess I'm going to take a deeper look at it, I guess following the tailscale API is the better solution, but I don't see how to do it as the Tailscale DERP server already implement that, I'm concerned about not coding the same thing twice and not use a part of it, maybe we should copy the DERP api files in Headscale, so we can do a workaround and connect the DERP directly with the Headscale instance.

Are some people using DERP instances on multiple hosts to have a scalable solution ?

joejose97:
@juanfont  

The tailscale derp server gets the valid client list by sending a GET request to http://local-tailscaled.sock/localapi/v0/status
I'm thinking of emulating what the tailscale client does by creating a http listener on the said socket.

If this is acceptable, I'll open  a PR with the same.

PR #1406, opened 2023-05-01
Update the docker container doc

loprima-l:
Update the docker container documentation to be more useful and accurate 


docs/running-headscale-container.md
```diff
@@ -6,72 +6,98 @@
...
+# 2. Create an empty [SQLite](https://www.sqlite.org/) datebase in the headscale directory:
 
 ```shell
 touch ./config/db.sqlite
 ```
 
-3. **(Strongly Recommended)** Download a copy of the [example configuration][config-example.yaml](https://github.com/juanfont/headscale/blob/main/config-example.yaml) from the headscale repository.
-
-Using wget:
+# 3. Create an ACL file in the headscale directory:
```
ohdearaugustin, line 37: Just point out that this is optional as the example-config does not point to a acl path.  Further down you add the according config. Good addition. Maybe add it as an optional point.

docs/running-headscale-container.md
```diff
@@ -6,72 +6,98 @@
...
+
+!!! warning "/var/lib isn't an editable path in the container"
+
+    Because /var/lib isn't an editable path in the container, we will need to replace
+    in the config file all those /var/lib paths to /etc.
+    
+    Example :
+    ```yaml
+    # The default /var/lib path is not writable in the container
+    private_key_path: /etc/headscale/private.key
```
ohdearaugustin, line 89: Same thing needs to be done for the noise key.

docs/running-headscale-container.md
```diff
@@ -105,67 +136,26 @@ Verify `headscale` is available:
...
-The `headscale/headscale` Docker container is based on a "distroless" image that does not contain a shell or any other debug tools. If you need to debug your application running in the Docker container, you can use the `-debug` variant, for example `headscale/headscale:x.x.x-debug`.
-
-### Running the debug Docker container
-
-To run the debug Docker container, use the exact same commands as above, but replace `headscale/headscale:x.x.x` with `headscale/headscale:x.x.x-debug` (`x.x.x` is the version of headscale). The two containers are compatible with each other, so you can alternate between them.
-
-### Executing commands in the debug container
+    The `headscale/headscale` Docker container does not contain a shell or any other debug tools. 
+    If you need to debug your application running in the Docker container, you can use the 
+    `-debug` variant, for example `headscale/headscale:0.22.1-debug`.
```
ohdearaugustin, line 149: Some information is lost like it is a distroless image and that is just a busybox shell, because that can be crucial information in my opinion.

ohdearaugustin (changes requested):
All in all it is great to have an updated version of this again. Thank you.

docs/running-headscale-container.md
loprima-l: Every guys familiar with docker will understand it and the new to docker will just use the debug doc below, adding to much details can lead us to a pain reading the documentation. What's your opinion on that ? I think that docker needs to be an easy way to try Headscale, so we want it as simple as possible.

docs/running-headscale-container.md
loprima-l: Don't everybody want to use ACLs ? And I'm pretty sure that the docker image won't boot without acls, but I need to check

docs/running-headscale-container.md
loprima-l: I'm not familiar with that concept, I'm gonna look at it.

loprima-l:
Thanks for your Reviews ^^

docs/running-headscale-container.md
ohdearaugustin: The noise key should also be in a writeable directory. 
```yaml
noise:
  private_key_path: /etc/headscale/noise_private.key
```

docs/running-headscale-container.md
ohdearaugustin: It will boot without ACLs. I don't mean to remove it just, as an extra point that you can optionally add a acl. It is not needed for a basic setup.

docs/running-headscale-container.md
ohdearaugustin: Fair point.

docs/running-headscale-container.md
loprima-l: I'm not sure, but is the noise key optional ? I can't find any information in the documentation

docs/running-headscale-container.md
loprima-l: After a time thinking to it, I came up with that : Initial work is a good understanding way. I'm sure that pretty much no one will use Headscale without ACLs, and even if it's the case, if them want to use them later (they would probably want), it'll be more simple for them to just edit a fill than finding the solution in the

loprima-l: What do you think ?

docs/running-headscale-container.md
juanfont: The Noise key is absolutely required :)

juanfont:
@loprima-l can you run prettier?

loprima-l:
I'll run prettier when we've found the best solution for ACLs file

docs/running-headscale-container.md
ohdearaugustin: This wording makes it sound like it is necessary, that is why I think it would be better it's own point and say it is optional.

loprima-l:
I can't run lint correctly with nix, and without nix I don't run into any lint problem... I will try a bit more tomorrow, maybe

PR #1417, opened 2023-05-03
Revert and fix Unix socket location

juanfont:
Reverts juanfont/headscale#1414

### DIFF
--- a/config.go
+++ b/config.go
@@ -175,7 +175,7 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetDefault("derp.server.enabled", false)
 	viper.SetDefault("derp.server.stun.enabled", true)
 
-	viper.SetDefault("unix_socket", "/var/run/headscale/headscale.sock")
+	viper.SetDefault("unix_socket", "/var/run/headscale.sock")
 	viper.SetDefault("unix_socket_permission", "0o770")
 
 	viper.SetDefault("grpc_listen_addr", ":50443")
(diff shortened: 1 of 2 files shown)


ohdearaugustin:
This whole thing broke the installation for most people who use docker. Just changing the default unix_socket for packing with debian. We probably should fix the debian package and not just change the unix_socket location. This is a breaking change.

juanfont:
> This whole thing broke the installation for most people who use docker. Just changing the default unix_socket for packing with debian. We probably should fix the debian package and not just change the unix_socket location. This is a breaking change.

Any idea how to fix both things?

ohdearaugustin:
I guess there are three options:

1. We change the the deb package build process, so it will also create the socket in /var/run/headscale.socket. I would still need to check if that is possible. This folder is created in the postinstall.sh of the package.

2. We add the /var/run/headscale/ folder directly in the docker container.

3. We only update the documentation for the install with docker that people need to change the unix_socket path in their setup to /var/run/headscale.socket

I personally prefer the first option as we would revert an unintentional breaking change of a default value. @kradalby Do you have any opinion on that? Or was this change necessary and I don't see why it was necessary?

config.go
```diff
@@ -174,7 +174,7 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetDefault("derp.server.enabled", false)
 	viper.SetDefault("derp.server.stun.enabled", true)
 
-	viper.SetDefault("unix_socket", "/var/run/headscale.sock")
+	viper.SetDefault("unix_socket", "/var/run/headscale/headscale.sock")
```
juanfont, line 177:
```suggestion
	viper.SetDefault("unix_socket", "/var/run/headscale.sock")
```

kradalby (approved):

PR #1420, opened 2023-05-05
[docs] Use modern apt command to install package

Daniel15:
`dpkg --install` is legacy 😃


Daniel15:
As a side note - any plans to create a Debian repo so we can just do `apt install headscale`? If not, is it OK if I create one?

ohdearaugustin:
Looks good to me. A debian repository would be indeed a nice to have, Probably integrated into the CI/CD would be awesome. (=

ohdearaugustin (approved):

ohdearaugustin:
f

PR #1422, opened 2023-05-05
Support hosting headscale with an HTTP path prefix

### LINKED ISSUE #1126: [feature] Support running under a HTTP path
**Feature request**

Tailscale client supports specifying not just host/port/protocol, but also a HTTP path when connecting to a control server:

```
tailscale up --login-server https://example.com/foo
```

Presumably this is to allow multiple control servers to be behind the domain, for example https://example.com/customer1 https://example.com/customer2, or to allow other services to also be behind a reverse proxy i.e  `https://example.com/headscale` for headscale backend, `https://example.com/my_api` for a different backend.

There doesn't seem to be any way to support this in Headscale currently. Headscale _does_ allow a path to specified in the config:

```
server_url: https://127.0.0.1:443/headscale
```

But the path in this URL doesn't seem to affect the application i.e. a client with `tailscale --login-server https://example.com/headscale`. The requests tailscale makes 404s since they are not provided by headscale.

**Implementation**
From what I can tell this feature would require the following changes:
* If `server_url` contains a path, the HTTP server should use this path as a prefix for all provided endpoints
* The [noise protocol implementation](https://github.com/juanfont/headscale/blob/93aca81c1cb20695d010766042a3abbe820364e8/noise.go#L66) should also use the path prefix, as Tailscale client also includes the path prefix in the Noise-encrypted API calls under /ts2021
* Perhaps same with embeded DERP server?
* Any other paths I may be missing?

I'm actively working on a patch for this to use myself, but if developers have any tips for getting this feature upstream let me know.

kev-the-dev:
Add support for hosting headscale endpoints with a HTTP path prefix. This is a feature requested by several developers, mainly to support running headscale behind a reverse proxy where there are other backends which headscale should not conflict with.

Unfortunately, through testing the tailscale client implementation has some strange behavior when adding a path to `--login-server`. It respects this prefix for certain endpoints, but does not for `/derp`, `/ts2021`, and `/bootstrap-dns`. 

Additionally, the admin API at `/api/v1` breaks if you host it with a path prefix as the paths are hardcoded into the generated code from swagger. 

These considerations were applied to this implementation so that only the supports paths will use the path prefix, and documentation was updated to clarify this.

Closes #1126

- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 - Fix systemd service file location in `.deb` packages [#1391](https://github.com/juanfont/headscale/pull/1391)
 - Improvements on Noise implementation [#1379](https://github.com/juanfont/headscale/pull/1379)
 - Replace node filter logic, ensuring nodes with access can see eachother [#1381](https://github.com/juanfont/headscale/pull/1381)
+- Support [hosting headscale behind a certain HTTP path](https://github.com/juanfont/headscale/issues/1126) [#1422](https://github.com/juanfont/headscale/pull/1422)
+  - For example, host at https://example.com/foo instead of https://example.com
+  - Some endpoints are always at the root of the server (`/`). See `config-example.yaml`
 
 ## 0.22.1 (2023-04-20)
 
--- a/app.go
+++ b/app.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"sort"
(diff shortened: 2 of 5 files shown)


juanfont:
Hi @kev-the-dev, unfortunately we cannot merge this. 

Having the API half-split in methods that work on a subpath and other than don't would add a lot of confusion and issues.

Reverse proxies are a source of user issues in Headscale, and this would add on top :(

--

Happy to have a chat on project priorities over DM in Discord :)

Issue #1409, opened 2023-05-02
ip_prefixes cannot be displayed properly


threerog:
When `ip_prefixes` is `100.64.0.0/24`, running `tailscale status` normally displays the IPs of all devices. However, when `ip_prefixes` is in other network segments, `tailscale status` cannot display the IPs of other devices,

Also, I have a question whether `Hedascale` supports multiple IPv4 `ip_prefixes`?
How to use multiple `ip_prefixes` if supported?

juanfont:
Ranger other than 100.64.0.0/10 and fd7a:115c:a1e0::/48 (or their subnets) are not supported, and it can/will lead to odd behaviour. 

This issue should be tackled in the Tailscale client.

PR #1425, opened 2023-05-06
Give a warning when users have set an unsupported prefix

juanfont:
The only supported prefixes to assign client IPs by the Tailscale client are 100.64.0.0/10 and fd7a:115c:a1e0::/48.

Despite that, we allow users to setup arbitrary ranges in the Headscale config. This kinda work, but also don't:

https://github.com/juanfont/headscale/issues/1409
https://github.com/juanfont/headscale/issues/1050
https://github.com/juanfont/headscale/issues/148
(and more)

This PR adds a warning in the config when users have other IP prefixes.

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 	"go4.org/netipx"
+	"tailscale.com/net/tsaddr"
 	"tailscale.com/tailcfg"
 	"tailscale.com/types/dnstype"
 )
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -510,11 +511,35 @@ func GetHeadscaleConfig() (*Config, error) {
 	configuredPrefixes := viper.GetStringSlice("ip_prefixes")
 	parsedPrefixes := make([]netip.Prefix, 0, len(configuredPrefixes)+1)
 
+	log.Warn().Msg("culo")
```
kradalby, line 514: What

kradalby (approved):

PR #1480, opened 2023-05-27
Fix hangs on SIGTERM.

### LINKED ISSUE #1461: Strange shutdown/restart behavior on 0.22.2
## Bug description

After upgrading to 0.22.2 restart or shutdown take like over 10 minutes on latest Debian. Installed from *.deb. Working like a charm on 0.22.1.

## Environment

- OS: Debian
- Headscale version: 0.22.2
- Tailscale version: 1.40.1 (not relevant)

## To Reproduce

Upgrade, do `systemd restart headscale`, wait like for 10 minutes to restart.

armooo:
This fixes #1461, where two bugs prevent main from existing on SIGTERM.

The first is a blocking receiving from a cancel chanel which never have a, cancelStateUpdateChan. It seems closing the chanel should be all the is needed to signal the watching goroutines to exit.

The other is the signal hander never exiting an infinite for loop. The sigFunc is called in an errorGroup which blocks exiting Serve and thus main. It looks like a refactor in #1382 removed an os.Exit(0), replacing it with a return breaks out of the loop.

- [ ] read the [CONTRIBUTING guidelines](README.md#contributing)
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -760,7 +760,6 @@ func (h *Headscale) Serve() error {
 				// Stop listening (and unlink the socket if unix type):
 				socketListener.Close()
 
-				<-h.cancelStateUpdateChan
 				close(h.stateUpdateChan)
 				close(h.cancelStateUpdateChan)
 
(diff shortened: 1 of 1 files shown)


kradalby:
Picking this up in #1492, added you as a co-author.

Issue #1473, opened 2023-05-23
[Code Freeze] Restructuring the code


kradalby:
Juan and I are currently undertaking a large code refactor to make the code more maintainable.

Currently the goal is to split a lot of the code into modules, which forces us to have separate concerns and make sure that more of the code does one thing, and does it well. We had a lot of cycling dependencies (allowed by one large package) and it made everything very tightly knitted, and testing very hard as you had to set up a massive dependency chain. 

Rough plan for how we want to do it:

1. Move control go code to package `hscontrol` (#1444)
2. Split "core" code into packages (types, db, policy) (#1458)
  - This is _mainly_ a reorg change, where the code change is only to be able to move code
  - `types` contains all our datatypes, and functions attached to them without dependencies
  - `db` contains all functions that manipulate and changes the database
  - `policy` contains all functions related to ACLs, a big change is that it does no longer depend on `db`
3. Split and potentially rewrite hscontrol into packages (this might change) (#1477)
  - `mapper`, all code related to generating MapResponses (the map clients need to find their peers)
  - ~`api`, the code that implements our gRPC interface~ (For now, leave grpc in root)
  - `derp`, derp related code
  - `dns`, related code
4. Rewrite policy package (#1487)
  - Generate filter and ssh rules per node
5. Rewrite polling and global state (#1492)
6. Rename Machine to Node (#1553)

kev-the-dev:
For clarity, this issue is tagged [Code Freeze] as the maintainers will wait to merge any non-reorg changes until after the re-org? Makes sense to me but wanted to clarify

kradalby:
@kev-the-dev, I realise that I have not actually written that out. Thanks for asking.

kradalby:
This has been implemented, and we are preparing to get these changes out as alpha soon, closing.

PR #1483, opened 2023-05-31
Bugfix: when db_type is set to postgres, tailscale client will be hang up for a long time

### LINKED ISSUE #764: Headscale fails to activate clients with postgresql backend

Yxnt:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #764 

When db_type is set to postgres, `tailscale login` and `tailscale up` can hang up for a long time. because expiry time is `0001-01-01 00:00:00`.

I'm not sure this change is make sense, but it‘s worked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,5 @@ integration_test/etc/config.dump.yaml
 /site
 
 __debug_bin
+.vscode
+.devcontainer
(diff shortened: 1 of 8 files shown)


kradalby:

PR #1521, opened 2023-08-06
fix invalid discord channel links to instead reference server invite and name of channel

unusualevent:
This edits the docs on faq.md and web-ui.md to remove channel links which were invalid and also not helpful if you aren't invited to the server already.

Instead, they are replaced with the server invite link, and the channel's name.


kradalby (approved):

kradalby:
This makes sense, sorry for missing this, can you rebase it?

Issue #1522, opened 2023-08-09
Bug in user switching during OIDC authentication


vsychov:
## Bug description
Hello,

I found an issue in headscale that seems like a bug to me. 
As a user, I can't change the OIDC account I'm already authorized with. I'm trying to do it this way:
1. login as user1
2. logout, by: `tailscale logout && tailscale down`
3. login as user2: `tailscale up` with the `--login-server` parameter. It redirects me to the OIDC authorization page (`https://my-server.com/oidc/register/nodekey:XXXXXXX`)

Upon successful authorization, I get an error in the web interface:
`could not register machine`, and in the app logs, there's also an error: `ERR go/src/headscale/hscontrol/oidc.go:712 > could not register machine error="machine was previously registered with a different user"`

The error seems to be here, where it's explicitly forbidden:
https://github.com/juanfont/headscale/blob/fb203a2e454a9907d017ed3c61c7002ed3065977/hscontrol/db/machine.go#L381-L386

vsychov:
Same issue is #1310, bug was caused by changes from https://github.com/juanfont/headscale/pull/754

NetworkEngineer509:
Can confirm I am coming across this as well. I was just about to submit a bug report but came across this.

micudaj:
same for me. just had this as well

kolaente:
Any update on this?

hannescl:
This bug "hit" me as well yesterday...

mannp:
I have just got this error after setting up my first user with oidc.

`headscale-start[167399]: 2024-01-22T16:08:55Z ERR github.com/juanfont/headscale/hscontrol/oidc.go:712 > could not register machine error="machine was previously registered with a different user"`

benmehlman:
I'm having a similar problem:

Headscale is installed and working (headscale v0.23.0-alpha5, on debian 12.4).

OIDC is set up using Google Workspace as the provider.
Connect tailscale client for linux, using my headscale server as the login-server.  Copy/paste oidc url, authenticate as a google workspace user.  Let's say "firstuser@mydomain.com".  It authenticates successfully.

Install tailscale client for windows (1.6.2).  Set registry values to point to my headscale server.  Select Login from the tailscale client menu.. it does OIDC auth in the browser.. authenticate using SAME google user as before, it works.

Everything works as configured (nodes can see each other, subnet route works, and configured dns server is used).  

(Amazing.. loving this.. thanks!).

[...]

PR #1492, opened 2023-06-21
Codereorg: Part 5, state 

kradalby:
This PR reworks the state management in headscale, instead of relying on timestamps and frequent checking, and sending the entire world every time, we now send incremental updates based on the change that has happened. 

Closes #1390
Fixes #1461

Issue #1310, opened 2023-04-03
Preauth keys for different users


QZAiXH:
**Bug description**
I have two preauth keys A and B, and a new client. I first use A to join headscale and this client successfully joins user A. After that I use B on the client and the client does not switch to user B, it is still on A. Is this correct?

DaClemens:
Furthermore fast account switching with multiple users on the same headscale instance is currently not working (afaict).

I suspect the same reason under the hood, headscale might be using the machine id/key for a more unique identifier than it should be.

I tested:
- joining via preauth key, subsequent `login` via preauth key for second user -> no user switch, no second account, just a renegotiation of node key
- joining via preauth key, subsequent `login` via OIDC -> oidc fails, in the log i see `could not register machine error="machine was previously registered with a different user"`
- joining via OIDC, subsequent `login` via preauth key -> same as #1. preauth key changes nothing, except for the node key.

pallabpain:
You should be able to fast-switch users seamlessly if you do the following after logging in with preauth key A.
```
tailscale logout
tailscale down

tailscale up --auth-key preauthkey-B ... 
```

DaClemens:
I tried it today on headscale 0.21.0 and it does not work. `tailscale switch --list` only shows a single user.

Carseason:
me to

DaClemens:
i believe it is still relevant however

ItsShadowCone:
Did anyone test this for v0.23.0?

kradalby:
> Furthermore fast account switching with multiple users on the same headscale instance is currently not working (afaict).
> 
> I suspect the same reason under the hood, headscale might be using the machine id/key for a more unique identifier than it should be.

Are you trying to login to the same headscale with the same node twice?

So [node a, login 1] and [node a, login 2] is both in your fast user switching menu and as two nodes in headscale?

Does Tailscale SaaS support this?

ItsShadowCone:
See my response in #1920

I think the whole point in fast user switching @ tailscale SaaS is same node multiple logins.

ItsShadowCone:
Can we confirm that this bug is either still existing or fixed in the latest beta?

stormshield-gt:
I can confirm that the bug is still present is the latest published image 2 months ago: sha256:392237fecf911ff101f56d92dbb2529eb8a1f065bdbdcd73744565c6e0744bad
When login with a new user it creates a new profile in `tailscaled.state` but also set the old one to `null`

[...]

PR #1528, opened 2023-08-15
Enhancement: Enable Account Switching for Non-Ephemeral Devices on Logout

vsychov:
Hello! I find myself saying this in almost every PR or ticket, but I'd like to take a moment to thank you once again for this fantastic product. I truly hope to see it thrive and evolve over time.

This PR addresses a user story and resolves a bug. Fixes: #1522 and #1310.

**User Story:**
As a user, I want the ability to switch between accounts that I'm logged into, either via OIDC or through an auth-key.

**Existing Issues:**
- When logging in via OIDC as User A, then logging out and trying to log in as User B, there's a web interface error stating `could not register machine`.
- If you log in using the auth-key for User A, log out, and then try to authenticate with the auth-key for User B, the authentication seems to work. However, you remain logged in as User A.

These scenarios are works for non-ephemeral devices. Ephemeral devices don't face this problem as they get removed upon logout. While these issues might be addressed in #1492, it got me thinking: Why not always remove devices on logout? What value do they have post-logout?

Hence, I'm suggesting this as either a temporary or permanent fix for the aforementioned issues.

### DIFF
--- a/hscontrol/auth.go
+++ b/hscontrol/auth.go
@@ -551,16 +551,13 @@ func (h *Headscale) handleMachineLogOut(
 		return
 	}
 
-	if machine.IsEphemeral() {
-		err = h.db.HardDeleteMachine(&machine)
-		if err != nil {
-			log.Error().
-				Err(err).
-				Str("machine", machine.Hostname).
-				Msg("Cannot delete ephemeral machine from the database")
-		}
-
-		return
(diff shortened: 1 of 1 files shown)


kradalby:

axxonadmin:
it is relevant as of v0.23.0-alpha1

PR #1460, opened 2023-05-11
Add sync errorgroup handling in integration tests

kev-the-dev:
Address TODOs in integration tests to handle errors from sync WaitGroup for things like bringing up a bunch of tailscale nodes. I ran into some confusing segfaults when running integration tests locally that would be helped by implementing this fail-on-error rather than log-on-error.

Closes #1459

- [x] read the [CONTRIBUTING guidelines](README.md#contributing)
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [x] added integration tests
- [x] updated documentation if needed
- [ ] updated CHANGELOG.md -> internal change, no changelog update

PR #1535, opened 2023-08-29
Handle errors in integration test setups

kradalby:
And use upstream containers for tests, should be a lot faster.

Thanks @kev-the-dev

Closes #1460


juanfont (approved):
♥️♥️

PR #1544, opened 2023-09-07
fix test fail keep user.Clients  from concurrent map writes

PatrickHuang888:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

```console
fatal error: concurrent map writes

goroutine 259 [running]:
github.com/juanfont/headscale/integration.(*EmbeddedDERPServerScenario).CreateTailscaleIsolatedNodesInUser.func1()
```

### DIFF
--- a/integration/embedded_derp_test.go
+++ b/integration/embedded_derp_test.go
@@ -197,7 +197,9 @@ func (s *EmbeddedDERPServerScenario) CreateTailscaleIsolatedNodesInUser(
 					)
 				}
 
+				s.headscaleLock.Lock()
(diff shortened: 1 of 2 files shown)


kradalby:
Ah thanks, I think this makes sense, but after just skimming this, should it have its own lock?

PatrickHuang888:
> Ah thanks, I think this makes sense, but after just skimming this, should it have its own lock?

yes, it can. but right now headscaleLock in senario just used in headscale creation, this just a test case init here, I think maybe use headscaleLock here simple enough and can work.

PR #1553, opened 2023-09-22
Rename Machine to Node

juanfont:
Almost three years ago I made the mistake of calling Machines to the Tailscale clients in Headscale.

Since then it has become very obvious that they should be referred as Nodes.

This massive PR fixes that mistake across the codebase.


hscontrol/db/db.go
```diff
@@ -78,49 +78,53 @@ func NewHeadscaleDatabase(
...
 
 	// GivenName is used as the primary source of DNS names, make sure
 	// the field is populated and normalized if it was not when the
-	// machine was registered.
-	_ = dbConn.Migrator().RenameColumn(&types.Machine{}, "nickname", "given_name")
+	// node was registered.
+	_ = dbConn.Migrator().RenameColumn(&types.Node{}, "nickname", "given_name")
 
-	// If the Machine table has a column for registered,
+	// If the MacNodehine table has a column for registered,
```
kradalby, line 101: Flagging this for later

kradalby (approved):

PR #1562, opened 2023-09-26
Restore foreign keys and add constraints

### LINKED ISSUE #1482: Orphaned routes accumulating in the routes table
## Bug description

Hello,

I'm using headscale with `autoApprovers` ACL and dynamic Tailscale subnet routers which frequently change their advertised routes. To handle traffic redirection to certain domain names with dynamic DNS, I've set up a script that monitors DNS changes and restarts Tailscale, adjusting the `advertise-routes` flag according to the resolved DNS. Additionally, the nodes running this setup are `ephemeral` and periodically replaced. This setup mostly works fine, except for a single issue I've discovered.

The routes table in the database seems to be filling up with entries for routes that are no longer relevant and have no association with any machines:

```sql
SELECT
  count(*)
FROM
  routes
  LEFT JOIN machines ON routes.machine_id = machines.ID 
WHERE
  machines.ID IS NULL
```

This returns approximately 20k entries that are no longer needed.

I believe this to be a bug, as I would expect the routes table to be cleaned up at some point (perhaps when a node is deleted), or when the `updated_at` field for a route is more than 24 hours old, and both `advertised` and `enabled` fields are set to `false`.

At present, I'm handling the table overflow issue with the following SQL:
```sql
DELETE 
FROM
  routes 
WHERE
  "id" IN ( SELECT routes."id" FROM routes LEFT JOIN machines ON routes.machine_id = machines."id" WHERE machines."id" IS NULL ) 
  OR ( updated_at <= ( NOW( ) - INTERVAL '1 DAY' ) AND advertised = 'f' AND enabled = 'f' ) 
```

Might it be beneficial to add a goroutine to handle this?
If this change is approved, I would be happy to create a pull request implementing the necessary adjustments.

Thank you for your attention to this matter.

Best regards.

vsychov:
Hello there, 

This pull request addresses the issue #1482, aiming to fix the anomalies regarding foreign keys and constraints in database models. Initially FK was disabled here: https://github.com/juanfont/headscale/pull/41 without any description why it was done.

These changes improve data integrity, maintain referential integrity between different tables, and ensure that the application's behavior remains consistent and predictable when handling database records, particularly in delete operations.

https://gorm.io/docs/gorm_config.html#DisableForeignKeyConstraintWhenMigrating

Fixes #1482

- [x] updated CHANGELOG.md


kradalby (commented):
Looks reasonable to me, I'll let @juanfont have the last say since he did the change, but I recon it would be hard to remember.

vsychov:
Hello, I was wondering if there have been any updates regarding the merge this?

kradalby:
Hi, I'm had some other work the previous weeks and I'm on holiday this week but hope to get back to the last things next week.

vsychov:
Hey @kradalby, could you check this out? I just resolved conflicts, and it seems like the tests are running smoothly now.

kradalby:
I think this looks reasonable, I'm running the tests now.

Are there any potential concerns we should think of that might not be covered by the current tests?

vsychov:
@kradalby , I couldn't think of any potential issues that might arise with this, and for which there are no tests already in place, the changes seem fairly safe.

vsychov:
@kradalby, I see that all tests passed except TestAuthKeyLogoutAndRelogin, which failed with the error:

`general_test.go:165: failed to have all clients sync up: reached retry deadline: [ts-1-40-muk9lb] peer count correct, but ts-head-pkmesn does not have a DERP`

and

`general_test.go:165: failed to have all clients sync up: [ts-1-40-rottie] peer count correct, but ts-head-roffmn does not have a DERP`

From what I understand, there seems to be an issue with DERP. Overall, is this test stable? Is there a chance that this instability is due to infrastructure issues with the tests, or is the test error caused by my changes specifically? Should I investigate further to find the root cause?

kradalby:
I'm seeing that particular test fail on one of my PRs too, but not locally, not sure what is up.

vsychov:
> I'm seeing that particular test fail on one of my PRs too, but not locally, not sure what is up.

So, can this PR be merged? If yes, let's do it please, to prevent future conflicts.

vsychov:
@kradalby, I found the reason why it works locally for you but not in pipelines. The issue lies in something broken in the Docker image of tailscale `head` and `unstable`, causing the `Peer.*.Relay` field to be empty in the clients status for these versions. Locally, you probably have a cached older version of this images, so the test works locally.

It might be worth reporting this in the issues section of tailscale. Since you're presumably affiliated with the tailscale team, perhaps you could inform them about this issue, as I don't have a full understanding of the reasons to describe the issue accurately.

Here's a snippet from the contents of `ts-1-60-q8fnnl_status.json`:
```
...
  "Peer": {
    "nodekey:04ac28657e6c28a0a71ab4c60e96d32921908e413de278eecae8693aff589445": {
      "HostName": "ts-unstable-ntcuuv",
      "Relay": "",
      ...
    },
    "nodekey:0de113990c6918c86cde623ec129c5977e5c904ecd80e4d9849bbb909fbfff04": {
      "HostName": "ts-1-40-kcnou1",
      "Relay": "par",
      ...
    },
    ...
}
```

I've also included a commit with a fix that adjusts the makefile command to run integration tests: https://github.com/juanfont/headscale/pull/1562/commits/18bd635e1cf66e76135ff8785c43735dea2351a4

kradalby:
I have not had too much time to look into it, but after clearing the cache I can confirm this behaviour, however, I dont get why I dont see it with any of the other tests because they have the same assertion.

kradalby:
What I find odd is that it all the other tests pass, and those two consistently fail. I've poked around in the tailscale client code and it doesnt seem like there is any change that should affect this.

vsychov:
It seems to have stopped working after version 1.64 because version 1.66 and anything newer no longer works with this test. I just tested it on 1.66.

vsychov:
I went through the versions of Tailscale and managed to find the version with the issue, `unstable-v1.65.188` (May 8, 2024 at 5:46 am) - this version has a problem. The previous version, `unstable-v1.65.179` (May 7, 2024 at 2:30 am), works fine. So, the issue likely occurred somewhere between these versions. Unfortunately, I don't know how to link the Docker image version to a tailscale Git commit, but I hope this helps in troubleshooting.

[Here's the link to the Docker Hub page with the versions](https://hub.docker.com/r/tailscale/tailscale/tags?page=&page_size=&ordering=&name=unstable-v1.65.)

kradalby (approved):

kradalby:
I'll take a look, thanks, since it is not related to this PR, I will merge it for now.

Issue #1514, opened 2023-07-25
When does derp.yaml support the InsecureForTests configuration item


hausen1012:
When does derp.yaml support the InsecureForTests configuration item

WhaleFell:
2025 Still unable to support InsecureForTests

PR #1565, opened 2023-09-28
Add Customization Options to DERP Map entry of integrated DERP server

alexhalbi:
This Pull request adds a draft for the possibility to add the IPv4 and IPv6 address to the DERP map for the integrated DERP server.
This enables the DERP server to be reachable by clients, if they are using an Exit-Node via the DERP server and loose DNS connection.

This happens for example when restarting Headscale and the Exit-Node on the same time. It is also possible that this happens if Headscale is restarted and one of the nodes, does not reconnect successfully to the integrated DERP server and tries to communicate DNS through it.

If you take a look into the [default derp map](https://login.tailscale.com/derpmap/default), all of them have their IPs in the list, the integrated DERP server does never have it's IP in the list.
See https://pkg.go.dev/tailscale.com/tailcfg#DERPNode for reference.

Furthermore this request adds the possibility to disable the automatic creation of the integrated DERP map entry completely and adds the possibility to use the integrated DERP server with a completely custom etry in the `derp.paths`.

This enables even more customization, by fully leveraging the official [DERPRegion](https://pkg.go.dev/tailscale.com/tailcfg#DERPRegion) and [DERPNode](https://pkg.go.dev/tailscale.com/tailcfg#DERPNode) configuration file, which is currently overwriten.
This should also implement #1514.

https://tailscale.com/kb/1118/custom-derp-servers/#step-2-adding-derp-servers-to-your-tailnet

- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md


config-example.yaml
```diff
@@ -95,6 +95,16 @@ derp:
     # For more details on how this works, check this great article: https://tailscale.com/blog/how-tailscale-works/
     stun_listen_addr: "0.0.0.0:3478"
 
+    # For better connection stability (especially when using an Exit-Node and DNS is not working), 
+    # it is possible to optionall add the public IPv4 and IPv6 address to the Derp-Map using:
+    IPv4: 1.2.3.4
+    IPv6: 2001:db8::1
```
kradalby, line 101:
```suggestion
    ipv4: 1.2.3.4
    ipv6: 2001:db8::1
```
 or maybe
```suggestion
    ip:
      v4: 1.2.3.4
      v6: 2001:db8::1
```

kradalby: What happens if only one is set? do we need to require both?

kradalby: It would be pretty cool if there was an option to resolve this on startup, but, on the other hand, that is kind of a can of worms on how to handle changes, how to resolve etc.

config-example.yaml
```diff
@@ -95,6 +95,16 @@ derp:
...
+
+    # This flag can be used, so the DERP map entry for the embedded DERP server is not written automatically,
+    # it enables the creation of your very own DERP map entry using a locally available file with the parameter DERP.paths
+    # If you enable the DERP server and set this to true, it is required to add the DERP server to the DERP map using DERP.paths
+    manual_derp_map: false
```
kradalby, line 106: I find this name a bit hard to get, I would guess that the majority of people would not use this option, so it could have a long and verbose name:

`disable_inject_of_private_derp_map_entry`: false

but double negative is hard, so we could have enabled by default:

`add_private_derp_map_entry`: true

kradalby: I'm not particularly convinced by any of my suggestions either, so open to suggestions.

config-example.yaml
alexhalbi: I would say having them as two seperate options is prettier.
If none is set we have the behaviour we have right now. If one is set it is just using IPv4 or IPv6, if both are set the client can choose.


I thought about getting them dynamically too, but that could be a problem, when the IP changes via DNS. I do not know if the DERP-Map is actually cached in Headscale and what happens if it is overridden with an internal IP on Headscale itself. So I thought better to simply add them to the config. 

It could be done dynamically though like in bootstrap-DNS: https://github.com/alexhalbi/headscale/blob/6e734bc864abc13f8d608cb7e566611bf9593b2f/hscontrol/derp/server/derp_server.go#L228
```go
addrs, err := resolver.LookupIP(resolvCtx, "ip", node.HostName)
```
It could be added here: https://github.com/alexhalbi/headscale/blob/6e734bc864abc13f8d608cb7e566611bf9593b2f/hscontrol/derp/server/derp_server.go#L86

config-example.yaml
alexhalbi: Since we are talking about the "embedded DERP server" in the config, what about:
`automatically_add_embedded_derp_region: true`

Default should then be changed to `true` obviously and if set `false` it needs to be defined using the `paths` variable.

config-example.yaml
kradalby: Sure, we dont have to do automatic resolving, but the option should be renamed the the latter example I think.

config-example.yaml
kradalby: I think that sounds reasonable.

alexhalbi: Hi @kradalby I added your requested change in 46d28909893ef18f1be38d99baec4a5ecfdae617

alexhalbi:
Hi @kradalby,
Please have a look, if the new variant including the changes in 46d28909893ef18f1be38d99baec4a5ecfdae617 could be merged.
Thanks,
Alex

kradalby:
@alexhalbi could you rebase and try again fixing the lint issues? The tests should be retrying a bit now and should pass.

kradalby:
oh and a changelog entry.

alexhalbi:
> @alexhalbi could you rebase and try again fixing the lint issues? The tests should be retrying a bit now and should pass.
The changes I made exclusively impact the following files: config.go, config-example.yaml, derp_server.go, and app.go. Despite running the "make lint" command and encountering numerous errors, it's important to note that none of these errors pertain to the modifications I made. Therefore, please ignore the lint job failures.

I fixed lint errors in config-example.yaml which I created

> oh and a changelog entry.

I added an entry, I hope it is okay for you.

kradalby:
@alexhalbi Looks good now, I rewrote some of the git history as it ended up with some massive commits in the middle proposing this as a 4000 LOC change. 

Please double check that I didnt miss anything in the rebasing.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+### Changes
+
+Added the possibility to manually create a DERP-map entry which can be customized, instead of automatically creating it. [#1565](https://github.com/juanfont/headscale/pull/1565)
```
kradalby, line 5: actually, can you move this under 0.23.0?

alexhalbi:
Hi, 
I moved the changelog entry and we checked everything. It should be all in the branch thanks.

kradalby (approved):

yqs112358:
Hello, so how can we set something like `"InsecureForTests": true` in local `derp.yaml`? 
Sorry but I have not found any examples and something like that in the repo🤔

PR #1566, opened 2023-09-28
Enhance pipeline stability and automatically retry unstable tests

vsychov:
Hello there,

This Pull Request aims to introduce changes designed to enhance the overall stability of the pipeline and implement an automatic retry mechanism for tests that are currently unstable.

There is example, pipeline with retry: https://github.com/vsychov/headscale/pull/2
There is pipeline without retry: https://github.com/juanfont/headscale/pull/1562

Both of them contains same headscale file changes.

Another way to do that, is use `--rerun-fails` and `--rerun-fails-max-failures` arguments of `gotestsum`.

### DIFF
--- a/cmd/gh-action-integration-generator/main.go
+++ b/cmd/gh-action-integration-generator/main.go
@@ -56,9 +56,13 @@ jobs:
             config-example.yaml
 
       - name: Run {{.Name}}
+        uses: Wandalen/wretry.action@master
         if: steps.changed-files.outputs.any_changed == 'true'
         run: |
-            nix develop --command -- docker run \
+        with:
+          attempt_limit: 5
+          command: |
+              nix develop --command -- docker run \
               --tty --rm \
               --volume ~/.cache/hs-integration-go:/go \
               --name headscale-test-suite \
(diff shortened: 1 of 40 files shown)


vsychov:
@kradalby, can you take a look?

kradalby:
Apologise, missed this one, I had it in my todo list to do something like this, but I think 50 is a bit excessive, how about we start with 5.

The goal needs to be to get the tests to a state where that (or less) is the upper threshold.

vsychov:
@kradalby , done, I changed it to 5.

cmd/gh-action-integration-generator/main.go
```diff
@@ -56,9 +56,12 @@ jobs:
             config-example.yaml
 
       - name: Run {{.Name}}
+        uses: Wandalen/wretry.action@master
         if: steps.changed-files.outputs.any_changed == 'true'
         run: |
-            nix develop --command -- docker run \
+        with:
+          attempt_limit: 5
+          command: |
```
kradalby, line 64: I think this removes a bit too much?
```suggestion
        with:
          attempt_limit: 5
          command: |
            nix develop --command -- docker run \
```

kradalby (approved):

PR #1580, opened 2023-10-21
feat(derp): support `verify-clients`

mritd:
- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

Add a local api to support derper's client validation #740 

Signed-off-by: kovacs <mritd@linux.com>

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -80,6 +80,9 @@ derp:
     # The Headscale server_url defined above MUST be using https, DERP requires TLS to be in place
     enabled: false
 
+    # If true, derp will reject other unknown tailscale client connections
+    verify_clients: true
+
     # Region ID to use for the embedded DERP server.
     # The local DERP prevails if the region ID collides with other region ID coming from
     # the regular DERP config.
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -83,6 +83,10 @@ derp:
     # If true, derp will reject other unknown tailscale client connections
     verify_clients: true
 
+    # If true, headscale will simulate a socket API, which allows client authentication to be
+    # completed without running tailscaled independently.
+    enable_socket_api: false
+
     # Region ID to use for the embedded DERP server.
     # The local DERP prevails if the region ID collides with other region ID coming from
     # the regular DERP config.
--- a/hscontrol/derp/server/derp_server.go
+++ b/hscontrol/derp/server/derp_server.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 	"net/url"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"tailscale.com/ipn/ipnstate"
@@ -316,6 +317,21 @@ func serverSTUNListener(ctx context.Context, packetConn *net.UDPConn) {
 
 func (d *DERPServer) ServeFakeStatus() error {
 	socketPath := paths.DefaultTailscaledSocket()
+	socketDir := path.Dir(socketPath)
+	st, err := os.Stat(socketDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = os.MkdirAll(socketDir, 0755)
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
+	}
+	if !st.IsDir() {
+		return fmt.Errorf("the socket dir path(%s) already exists, but is a file", socketDir)
+	}
 
 	log.Trace().Caller().Msg("Clean up fake status socket file")
 	if err := os.RemoveAll(socketPath); err != nil {
--- a/hscontrol/derp/server/derp_server.go
+++ b/hscontrol/derp/server/derp_server.go
@@ -349,7 +349,10 @@ func (d *DERPServer) ServeFakeStatus() error {
 				}
 				peer[*pk] = &ipnstate.PeerStatus{}
 			}
-			status := &ipnstate.Status{Peer: peer}
+			status := &ipnstate.Status{
+				Self: &ipnstate.PeerStatus{},
+				Peer: peer,
+			}
 			w.Header().Set("Content-Type", "application/json")
 			e := json.NewEncoder(w)
 			e.SetIndent("", "\t")
--- a/hscontrol/derp/server/derp_server.go
+++ b/hscontrol/derp/server/derp_server.go
@@ -14,6 +14,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"tailscale.com/client/tailscale"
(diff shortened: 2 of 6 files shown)


hscontrol/derp/server/derp_server.go
```diff
@@ -302,3 +314,72 @@ func serverSTUNListener(ctx context.Context, packetConn *net.UDPConn) {
...
+		} else {
+			return err
+		}
+	}
+	if !st.IsDir() {
+		return fmt.Errorf("the socket dir path(%s) already exists, but is a file", socketDir)
+	}
+
+	log.Trace().Caller().Msg("Clean up fake status socket file")
+	if err := os.RemoveAll(socketPath); err != nil {
```
6ixfalls, line 337: It looks like this removes the socket at the default tailscale socket location. If the host is also running a seperate tailscale client, would this not break that?

mritd: This is indeed a difficult decision. Since we are using a simulated socket API solution, we must be able to manage the socket files regardless of who created them before.

This implementation does not allow Tailscale and headscale DERP to coexist. If we consider the Tailscale client, we have only two options:

Headscale DERP fails to start to minimize the impact on the Tailscale client behavior.
Take direct control of the socket API to meet our own requirements, but this will cause the Tailscale client to fail to run.

6ixfalls: Theoretically, I'd assume if there's a Tailscale client on the same host as Headscale, the client is connected to the Headscale server. In that case, would it be possible to completely skip socket emulation, and use the existing socket for client verification? I'm not a maintainer of Headscale so you may want to wait for their opinion - but that seems to be the most optimal method. If the Tailscale DERP server includes an option for a Tailscale socket, that may be even better, to use a separate socket to not conflict with the default client. You could even do some detection where if the socket is found from the Tailscale client, it uses that to verify instead.

mritd: I can suggest adding two security measures:

- 1、Add an enable_socket_api field in the configuration file to allow users to explicitly enable the simulation of the socket API.
- 2、Before performing the cleanup, call tailscale.Status() to verify if it returns a result. If it does, it indicates that another tailscaled process is listening on the socket file. In this case, a warning should be issued or the execution should be terminated.

6ixfalls (approved):
LGTM

hscontrol/derp/server/derp_server.go
kradalby: Hi, thanks for this work, I think the way to move this forward is to require tailscale to run on the node hosting headscale for this feature to work.

In stead of messing around with sockets and potentially removing them, breaking things in a way that might trigger support, I think we should:

- Drop the fake status
- Check if the tailscaled socket is present
- Verify that it is running and is logged into the headscale of this machine, 
- Fatal if these conditions do not hold

This means that the config option `verify_clients` needs to state how it works and what the prerequisites are, and we need good error messages for the different failure conditions.

`enable_socket_api` is dropped.

mritd: It is not difficult to detect whether tailscaled is running locally, as it has already been implemented in PR #1400. By enabling the derper setting option, derper will automatically perform the detection.

However, we need to clarify what the ultimate goal of this feature is:

- If we still require deploying a separate tailscaled instance, it means we are not considering portability.
- If we verify whether tailscaled is logged into the current headscale, it means we are changing the operational mode of derper(the official derper doesn't care about which server tailscale is logged into).

**If our goal is to "support verify-clients like the official derper" we simply need to add a configuration and make code modifications similar to PR #1400.**

**If our goal is to "support verify-clients like the official derper while making deployment more convenient" then we only need to simulate a local socket API.**

mritd:
How can this PR be improved? Is there any clear direction?

hscontrol/derp/server/derp_server.go
kradalby: I think I would not have it mess with the official socket, I'm not oppose to the simulation, but I do not want a scenario where sockets might be deleted causing very hard to debug scenarios.

An alternative could be that the simulation uses a different socket, and if that option is set, then derper is told to look at that socket, instead of the official path. e.g. /var/lib/headscale/derp_sim.sock

Even better, if you can trick it to not need a socket, that would be ideal.

config-example.yaml
```diff
@@ -80,6 +80,13 @@ derp:
     # The Headscale server_url defined above MUST be using https, DERP requires TLS to be in place
     enabled: false
 
+    # If true, derp will reject other unknown tailscale client connections
+    verify_clients: true
+
+    # If true, headscale will simulate a socket API, which allows client authentication to be
+    # completed without running tailscaled independently.
+    enable_socket_api: false
```
kradalby, line 88: If you managed to implement the alternative socket route as proposed, I'm tempted to say that we don't have this option, and always use the simulated socket if someone turns on verify_clients.

If not, I think this needs a better name.

hscontrol/derp/server/derp_server.go
mritd: Changing the socket file path is indeed the best way to go, but it cannot be achieved currently without modifying the official derper code of tailscale.

The official derper will always use [tailscale.Status()](https://github.com/tailscale/tailscale/blob/86c8ab7502a38b4de05308355fe0c847e4e78167/derp/derp_server.go#L1130) to get the status, and this code will always use the [default socket path](https://github.com/tailscale/tailscale/blob/86c8ab7502a38b4de05308355fe0c847e4e78167/client/tailscale/localclient.go#L105).

config-example.yaml
mritd: It is currently unlikely to adjust the fake socket file path 😥. Do you have any suggestions for naming this field? (I apologize for English not being my native language.)

zsio:
Hi all,

Just a quick check-in, what's the current status of this PR? Do we have a specific direction we're heading in? It seems like there hasn't been any progress lately. Could anyone please provide an update?

huanshiwushuang:
I hope someone can solve this problem or come up with a better solution, thank you

zhangguanzhang:
any update?

mritd:
> any update?

No... I've given up.

VisoTC:
I noticed that the latest version supports [authentication via webhook](https://github.com/tailscale/tailscale/blob/375617c5c804134ad28a34122e072e4bbb009675/cmd/derper/derper.go#L60). Does this mean that verification can be satisfied without breaking anything?

mritd:
Since there has been no progress for too long, I decided to close this PR.

PR #1583, opened 2023-10-30
🐛 fix(db): sets max open and idle connections for postgres

pallabpain:
### Description
When Postgres is used as the backing database for headscale, it does not set a limit on maximum open and idle connections which leads to hundreds of open connections to the Postgres server. This commit fixes that by setting up a connection pool while creating the gorm DB instance.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### Depends
* https://github.com/juanfont/headscale/pull/1700


hscontrol/db/db.go
```diff
@@ -360,10 +361,17 @@ func openDB(dbType, connectionAddr string, debug bool) (*gorm.DB, error) {
...
-		return gorm.Open(postgres.Open(connectionAddr), &gorm.Config{
+		db, err := gorm.Open(postgres.Open(connectionAddr), &gorm.Config{
 			DisableForeignKeyConstraintWhenMigrating: true,
 			Logger:                                   dbLogger,
 		})
+
+		sqlDB, _ := db.DB()
+		sqlDB.SetMaxOpenConns(10)
+		sqlDB.SetMaxIdleConns(10)
+		sqlDB.SetConnMaxIdleTime(time.Hour)
```
kradalby, line 372: Should we make this user configurable?

pallabpain: > Should we make this user configurable?

Thanks, that's a good idea. I'll create another commit with the changes required to make these fields configurable.

pallabpain: @kradalby I've pushed the commit. Please take a look. 🙇🏼‍♂️ 

Thanks.

config-example.yaml
```diff
@@ -142,11 +142,17 @@ db_path: /var/lib/headscale/db.sqlite
 # db_name: headscale
 # db_user: foo
 # db_pass: bar
+# db_max_idle_conns: 5
+# db_max_open_conns: 5
+# db_conn_max_idle_time_secs: 3600
```
kradalby, line 147: I think we need to prefix this with postgres, like `db_postgres_` or something as they do not apply to SQLite.

I might try to sneak a PR in before this to restructure the db config params because currently it is a bit chaotic, something like:

```yaml
db:
	type: postgres
	name: headscale
	user: foo
	password: bar
	postgres:
		max_idle_conns: 5
		max_open_conns: 5
```

or something like:

```yaml
db:
	type: postgres
	url: sql-style-url	

	sqlite:
		path: /var/run/db.sqlite

	# OR
	
	postgres:
		name: headscale
		user: foo
		password: bar
		max_idle_conns: 5
		max_open_conns: 5
```

I think I like the second option, and it can be implemented backwards compatible for a couple of versions with a warning of the change. What do you think? Makes it easier to implement dbURL style configs and sqlite specific stuff.

kradalby: Sorry for the late reply, I will see if I have time to make the change today/tomorrow and then you can rebase against it.

pallabpain: Thank you.

kradalby: https://github.com/juanfont/headscale/pull/1700 is up now, please take a look.

kradalby:
Sorry for the delay, this should be up and running

pallabpain:
> Sorry for the delay, this should be up and running

@kradalby Thank you. I've updated the PR.

I may have also run `make fmt` that has resulted in a larger diff in the PR. 😅

kradalby (approved):

SuperSandro2000:
@pallabpain does that mean headscale is now usable with postgres?

pallabpain:
> @pallabpain does that mean headscale is now usable with postgres?

@SuperSandro2000 It was already usable with Postgres. It now makes use of connection pooling as well.

SuperSandro2000:
I constantly ran into https://github.com/juanfont/headscale/issues/1584 which rendered headscale completely broken for me. It's that issue now fixed and can be closed?

pallabpain:
Yes, that should be resolved with this fix.

PR #1600, opened 2023-11-14
Add DERP generate_204 endpoint for captive portal detection.

tomtom5152:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

When connecting to a DERP server the client makes a request to `/generate_204` endpoint with the header `X-Tailscale-Challenge`. In order to pass the captive portal test, the server must provider a 204 NoContent response with `X-Tailscale-Response` set.

The source for this simple endpoint was taken directly from [Derper source at the time of commit](https://github.com/tailscale/tailscale/blob/955e2fcbfb4fe7ff9b8dbd665ba24ef2008c676e/cmd/derper/derper.go#L324)

### DIFF
--- a/.github/workflows/test-integration-v2-TestDERPValidateEmbedded.yaml
+++ b/.github/workflows/test-integration-v2-TestDERPValidateEmbedded.yaml
@@ -1,0 +1,67 @@
+# DO NOT EDIT, generated with cmd/gh-action-integration-generator/main.go
+# To regenerate, run "go generate" in cmd/gh-action-integration-generator/
+
+name: Integration Test v2 - TestDERPValidateEmbedded
+
+on: [pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  TestDERPValidateEmbedded:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - uses: satackey/action-docker-layer-caching@main
+        continue-on-error: true
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v34
+        with:
+          files: |
+            *.nix
+            go.*
+            **/*.go
+            integration_test/
+            config-example.yaml
+
+      - name: Run TestDERPValidateEmbedded
+        uses: Wandalen/wretry.action@master
+        if: steps.changed-files.outputs.any_changed == 'true'
+        with:
+          attempt_limit: 5
+          command: |
+            nix develop --command -- docker run \
+              --tty --rm \
+              --volume ~/.cache/hs-integration-go:/go \
+              --name headscale-test-suite \
+              --volume $PWD:$PWD -w $PWD/integration \
+              --volume /var/run/docker.sock:/var/run/docker.sock \
+              --volume $PWD/control_logs:/tmp/control \
+              golang:1 \
+                go run gotest.tools/gotestsum@latest -- ./... \
+                  -failfast \
+                  -timeout 120m \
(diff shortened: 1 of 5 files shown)


kradalby:
Is there a way we can have an integration test for this?

kradalby:
@tomtom5152 Could you rebase this?

tomtom5152:
Sorry, only just seen this as been having to put my attentions elsewhere. I don't have time to rebase currently, but I see there was some force push just. I also looked at adding tests but couldn't figure out how best to do this with the current repo structure. If you've got any hints I ~might~ hopefully, maybe have some time to look at it again after this coming week.

kradalby:
hmm, I wonder if I have seen an error message referring to the generate_204 endpoint when running `tailscale debug derp <region>`, so one option could be to setup a integration test (`integration/` folder, look at `embedded_derp_test.go`):

- Set up headscale with this new endpoint
- Setup a tailscale client
- execute `tailscale debug derp 999`
- unmarshal it (there is probably a type in the tailscale source code you can use)
- check if there is a warning/err for the generate_204 interface

I would appreciate if you have a go, if not, at somepoint I might have time, but that could be weeks to months these days.

I greatly appreciate this addition, but it is a feature that is hard to debug if it is wrong, so I am extra conservative for these kind of things.

kradalby:
I've added a testcase as I've described for endpoint, it is currently failing, and I have not had time to investigate, feel free to take a look @tomtom5152.

cc @TotoTheDragon since we talked about this the other day.

kradalby:
The DERP report looks as follows:

```json
{
         "Info": [
          "Region 999 == \"headscale\""
         ],
         "Warnings": [
          "Having only a single DERP region (i.e. removing the default Tailscale-provided regions) is a single point of failure and could hamper connectivity",
          "Error making request to the captive portal check \"http://headscale/generate_204\"; is port 80 blocked?"
         ],
         "Errors": [
          "Error connecting to node \"headscale\" @ \"headscale:80\" over IPv4: dial tcp4: lookup headscale on 100.100.100.100:53: no such host",
          "Error connecting to node \"headscale\" @ \"headscale:80\" over IPv6: dial tcp6: lookup headscale on 100.100.100.100:53: no such host"
         ]
}
```

I recon we might need to listen specifically on port 80 for that endpoint.

TotoTheDragon:
I am not sure if the tailscale client actually uses this. But within DERPNode, there is a field called CanPort80. Maybe we should look into setting that field?

kradalby:
> I am not sure if the tailscale client actually uses this. But within DERPNode, there is a field called CanPort80. Maybe we should look into setting that field?

It sounds useful, but searching the clients code it doesnt look like it is used.

balki:
Looks like tailscale client falls back to tailscale servers without this feature in headscale DERP servers. Ref: https://github.com/tailscale/tailscale/issues/13788#issuecomment-2407737147

kradalby:
There has been some work on the DERP integration tests to fix the verify client url and websockets, this might be helpful for these tests, anyone able to pick this up again?

Leseratte10:
Not sure if anyone plans to pick up this PR again since it's already fairly old, but I skimmed over the code and I see a potential issue. In there it says that "[the listening port] is not configurable as captive portal busting requires http/80."

And then it hardcodes the listen address to be 0.0.0.0:80; and that will break IPv6 setups. Can this be replaced with [::]:80 and a dual-stack-capable socket so it works for both IPv4 and IPv6 setups? Or two separate sockets for IPv4 and IPv6?

Headscale has quite a few places where it's currently only using IPv4 by default, but they can all be fixed by replacing "0.0.0.0" with "[::]" and "127.0.0.1" with "[::1]" in the config file. Please don't add another hardcoded dependency for IPv4 that can't be changed by the user. 

Or is there a compelling reason why this endpoint should only be reachable over IPv4? Same issue also applies to the other PR (#1768).

kradalby:
Happy for this to be attempted to pick up, I have not read up more on it since last time. There has been quite some changes to the code so there is substantial rebase work to get it on. I'm going to close it, so feel free to create a new PR when it might be ready for a new attempt.

PR #1609, opened 2023-11-17
Use tailscale key types instead of strings

kradalby:


hscontrol/auth.go
```diff
@@ -456,12 +456,12 @@ func (h *Headscale) handleNewNode(
 		resp.AuthURL = fmt.Sprintf(
 			"%s/oidc/register/%s",
 			strings.TrimSuffix(h.cfg.ServerURL, "/"),
-			registerRequest.NodeKey,
```
juanfont, line 463: I think we should use NodeKey - or we might break fast user switching

kradalby: We need a test for fast user switching.

Node keys can change.

juanfont (approved):

PR #1490, opened 2023-06-17
Allow setting ForcedTags using the acls.hujson config file

networkException:
This pull request introduces a declarative way to set ForcedTags for machines.

If a "forcedTags" field is set in the acls.hujson file, the ForcedTags of machines will be overriden in the database based on the tags defined in the config file.

```json
{
    "hosts": {
        "a": "fd7a:115c:a1e0::1",
        "b": "fd7a:115c:a1e0::2"
    },
    "forcedTags": {
        "tag:some-tag": [
            "a",
            "tag:some-other-tag"
        ],
        "tag:some-other-tag": [
            "b"
        ]
    }
}
```

- [x] read the [CONTRIBUTING guidelines](README.md#contributing)
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated documentation if needed
- [ ] updated CHANGELOG.md

PR #1617, opened 2023-11-26
Allow setting ForcedTags using the acls.hujson config file

networkException:
This pull request introduces a declarative way to set `ForcedTags` on nodes by recursively expanding aliases. These declarative forced tags are considered separately from the ones stored in the database.

Continuation of #1490 

```json
{
    "hosts": {
        "a": "fd7a:115c:a1e0::1",
        "b": "fd7a:115c:a1e0::2"
    },
    "forcedTags": {
        "tag:some-tag": [
            "a",
            "tag:some-other-tag"
        ],
        "tag:some-other-tag": [
            "b"
        ]
    }
}
```

- [x] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/policy/acls.go
+++ b/hscontrol/policy/acls.go
@@ -767,6 +767,15 @@ func (pol *ACLPolicy) expandIPsFromTag(
 		}
 	}
 
+	for _, alias := range pol.ForcedTags[alias] {
+		ips, err := pol.ExpandAlias(nodes, alias)
+		if err != nil {
+			return nil, err
+		}
+
+		build.AddSet(ips)
+	}
+
 	// find tag owners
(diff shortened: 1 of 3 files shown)


PizzaProgram:
Dear @networkException ,

Can you please explain a little bit more about this approach?
- What is it good for?
- How can it be used better than the "ones stored in the database" ?
- Are these extra tags show/used in some of the UI ?

Maybe reviewers would understand it better too, why did you commit this.
_(Personally I'm not.)_

kradalby:
I do not think we want to support more things in the policy file. It is currently has a lot of moving parts, and it does not work as intended. Adding more will make it harder for us to get it into shape. If there is a good case for this, it can be revisited in the future.

PR #1620, opened 2023-11-27
Do not overwrite the derpmap with default region

Orhideous:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

There are cases when some regions are already specified in `derp.yaml` (including those with ID 900) - in such cases, you should not overwrite them with the generated ones.

This problem [was discussed](https://discordapp.com/channels/896711691637780480/896711692120129540/1117183778704588800) at Discord.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ after improving the test harness as part of adopting [#1460](https://github.com/
 
 ### Changes
 
+Use versioned migrations [#1644](https://github.com/juanfont/headscale/pull/1644)
 Make the OIDC callback page better [#1484](https://github.com/juanfont/headscale/pull/1484)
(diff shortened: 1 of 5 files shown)


hscontrol/app.go
```diff
@@ -489,13 +489,18 @@ func (h *Headscale) Serve() error {
 			return errSTUNAddressNotSet
 		}
 
-		region, err := h.DERPServer.GenerateRegion()
-		if err != nil {
-			return err
+		regionsCount := len(h.DERPMap.Regions)
```
kradalby, line 492: I think that we should return an error if there is a collision, as far as I understand this will cause custom DERPs to only be added when you have no other DERPs.

If the upstream or current map has the ID, we should probably log.Fatal.

PR #1621, opened 2023-11-27
Add private key path to example config

1drturtle:
Adds private key path to the example config. 

This fixes an issue found in the discord where the service did not have a default private key path, so it errored with very little information for the end user. 

This config settings was previously only shown in the container installation docs.

- [ ] raised a GitHub issue or discussed it on the projects chat beforehand (N/A)
- [ ] added unit tests (N/A)
- [ ] added integration tests (N/A)
- [ ] updated CHANGELOG.md (N/A)

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -40,6 +40,10 @@ grpc_listen_addr: 127.0.0.1:50443
 # are doing.
 grpc_allow_insecure: false
 
+# Private Key path for the service. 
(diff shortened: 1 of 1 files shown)


kradalby:
This is intentionally removed as it is being dropped in the next version. You should use the config-example from the git tag equal to your version.

1drturtle:
the docs at `headscale.net` point to the config example on the main branch. Should it maybe point to the config example on the stable version?

PR #1652, opened 2023-12-14
add 1.54 and 1.56 to integration tests

kradalby:


kradalby:
These are failing because 1.56 has not been released to dockerhub yet.

juanfont (approved):

PR #1685, opened 2024-01-12
WIP DB work

kradalby:
nothing to see yet.

TLDR: make sure we use SQL Transactions to ensure data consistency (do not use locks). Move "notifying"/ sending state updates out of the database, should reduce risk of deadlock.

### DIFF
--- a/hscontrol/db/addresses_test.go
+++ b/hscontrol/db/addresses_test.go
@@ -7,10 +7,16 @@ import (
 	"github.com/juanfont/headscale/hscontrol/util"
 	"go4.org/netipx"
 	"gopkg.in/check.v1"
+	"gorm.io/gorm"
 )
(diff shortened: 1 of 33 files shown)


hscontrol/db/addresses_test.go
```diff
@@ -41,10 +47,13 @@ func (s *Suite) TestGetUsedIps(c *check.C) {
 		AuthKeyID:      uint(pak.ID),
 		IPAddresses:    ips,
 	}
-	db.db.Save(&node)
-
-	usedIps, err := db.getUsedIPs()
+	db.Write(func(rx *gorm.DB) error {
```
kradalby, line 50: tx

kradalby:
Superseeded by #1701

PR #1698, opened 2024-01-23
docs/reverse-proxy: use standard map, correct X-Forwarded-Proto varia…

SuperSandro2000:
…ble for nginx

I tested this on NixOS. After switching from postgres back to sqlite and with the standard websocket proxy config, everything started to work.

Initially the none standard map which differs from upstreams docs https://nginx.org/en/docs/http/websocket.html got me very confused and I thought this would be somehow required.

Also $http_x_forwarded_proto doesn't exist, NixOS uses $scheme instead and I am going with that.

proxy_redirect also didn't work for me but that might be on my rather big config.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/docs/reverse-proxy.md
+++ b/docs/reverse-proxy.md
@@ -33,8 +33,7 @@ The following example configuration can be used in your nginx setup, substitutin
 
 ```Nginx
 map $http_upgrade $connection_upgrade {
-    default      keep-alive;
-    'websocket'  upgrade;
+    default      upgrade;
     ''           close;
 }
 
@@ -61,7 +60,7 @@ server {
         proxy_buffering off;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
+        proxy_set_header X-Forwarded-Proto $scheme;
         add_header Strict-Transport-Security "max-age=15552000; includeSubDomains" always;
     }
 }


ohdearaugustin:
> Also $http_x_forwarded_proto doesn't exist, NixOS uses $scheme instead and I am going with that.

Yeah probably only NixOS uses this config that way. You PR seems like it would only work for your NixOS setup. This wouldn't be helpful config for people not using NixOS. 

My suggestion would be that you can add a comment, that if you run nginx on NixOS you would probably need to replace $http_x_forward_proto for $schema

SuperSandro2000:
> Yeah probably only NixOS uses this config that way.

That's wrong. NixOS' nginx config is pretty standard. http_x_forwarded_proto is a variable coming from the k8s nginx ingress controller, see https://docs.nginx.com/nginx-ingress-controller/configuration/virtualserver-and-virtualserverroute-resources/#:~:text=variables%3A%20%24scheme%20and-,%24http_x_forwarded_proto,-.%20Variables%20must%20be . In normal nginx this var doesn't exist.

> You PR seems like it would only work for your NixOS setup.

I didn't test this but I expect the snippet now to work with the majority of setups unlike before.

The NixOS specific code would look like

```nix
services.nginx.virtualHosts."headscale.example.com" = {
  locations."/" = {
  proxyPass = "http://127.0.0.1:${toString services.headscale.port}";
  proxyWebsockets = true;
  extraConfig = ''
    proxy_buffering off;
  '';
};
```

as you can see no special sauce is needed and the standard options defined for other websockets and proxies just work.

> My suggestion would be that you can add a comment, that if you run nginx on NixOS you would probably need to replace $http_x_forward_proto for $schema

Unless you are running headscale on k8s with the nginx controller then I don't know a scenario where that variable exists and even then it defaults to scheme https://github.com/nginxinc/kubernetes-ingress/blob/c503483ef20476f522850dbd6db32e38e2565cee/internal/configs/oidc/oidc_common.conf#L11-L14

ohdearaugustin:
Also did more research now. You are right. This mapping is mainly used by the nginx ingress and as you pointed out in the end it is just a remapping of `$schema`

From your wording:
> Also $http_x_forwarded_proto doesn't exist, NixOS uses $scheme instead and I am going with that.
It sounded like it is just a NixOS thing for me.  Sorry for the misunderstanding.

This PR makes sense.

SuperSandro2000:
I've accidentally deleted my fork, so I can't reopen this PR. Continuing in #1790

PR #1700, opened 2024-01-25
Restructure database config

kradalby:
This PR restructures the database part of the configuration to follow the indent style of the rest of the config. It gives a better indication for which setting belongs to `sqlite` and `postgres`.

It is intended for making adding specific features like #1583 less confusing. 

The old keys are aliased to the new settings keys so this is not breaking, but marked as deprecated in the changelog so we can remove them later.

New config example:

```yaml
database:
  type: sqlite3

  # SQLite config
  sqlite:
    path: /var/lib/headscale/db.sqlite

  # # Postgres config
  # postgres:
  #   # If using a Unix socket to connect to Postgres, set the socket path in the 'host' field and leave 'port' blank.
  #   type: postgres
  #   host: localhost
  #   port: 5432
  #   name: headscale
  #   user: foo
  #   pass: bar

  #   # If other 'sslmode' is required instead of 'require(true)' and 'disabled(false)', set the 'sslmode' you need
  #   # in the 'db_ssl' field. Refers to https://www.postgresql.org/docs/current/libpq-ssl.html Table 34.1.
  #   ssl: false
```


config-example.yaml
```diff
@@ -138,24 +138,26 @@ ephemeral_node_inactivity_timeout: 30m
...
-# db_pass: bar
-
-# If other 'sslmode' is required instead of 'require(true)' and 'disabled(false)', set the 'sslmode' you need
-# in the 'db_ssl' field. Refers to https://www.postgresql.org/docs/current/libpq-ssl.html Table 34.1.
-# db_ssl: false
+database:
+  type: sqlite3
+
+  # SQLite config
+  sqlite:
```
pallabpain, line 145: todo: Should be `sqlite3` here, right?
```suggestion
  sqlite3:
```

config-example.yaml
```diff
@@ -138,24 +138,26 @@ ephemeral_node_inactivity_timeout: 30m
...
-# db_port: 5432
-# db_name: headscale
-# db_user: foo
-# db_pass: bar
-
-# If other 'sslmode' is required instead of 'require(true)' and 'disabled(false)', set the 'sslmode' you need
-# in the 'db_ssl' field. Refers to https://www.postgresql.org/docs/current/libpq-ssl.html Table 34.1.
-# db_ssl: false
+database:
+  type: sqlite3
```
pallabpain, line 142: Or should this be `type: sqlite`?

config-example.yaml
```diff
@@ -138,24 +138,26 @@ ephemeral_node_inactivity_timeout: 30m
...
+
+  # SQLite config
+  sqlite:
+    path: /var/lib/headscale/db.sqlite
+
+  # # Postgres config
+  # postgres:
+  #   # If using a Unix socket to connect to Postgres, set the socket path in the 'host' field and leave 'port' blank.
+  #   type: postgres
```
pallabpain, line 151: todo: This seems like a typo. We don't need `type` here.
```suggestion

```

pallabpain (changes requested):

config-example.yaml
kradalby: I've swapped to the type can be both sqlite and sqlite3 and kept the setting entry

config-example.yaml
kradalby: Removed

hscontrol/types/config.go
```diff
@@ -46,16 +46,9 @@ type Config struct {
 	Log                            LogConfig
 	DisableUpdateCheck             bool
 
-	DERP DERPConfig
+	Database DatabaseConfig
```
pallabpain, line 49: nit: Can we implement the method to get the connection string on this struct?
```
func (d DatabaseConfig) GetDBConnectionString() (string, error) {
  // Implements the switch case here and returns the DB string for the
  // configured db type, else an error.
}
```
What do you think?

hscontrol/app.go
```diff
@@ -117,32 +117,32 @@ func NewHeadscale(cfg *types.Config) (*Headscale, error) {
 	}
 
 	var dbString string
```
pallabpain, line 119: nit: If we were to implement the method to get the db string on the `DatabaseConfig`, we can just invoke 
```
dbString, err := cfg.Database.GetDBConnectionString()
if err != nil {
  return nil, errors.Wrap(err, "failed to get connection string")
}
```

pallabpain (approved):
Looks good. A few minor comments around refactoring the code for the database config.

hscontrol/types/config.go
kradalby: I ended up doing it a bit from the other side, I moved the dbString logic into the openDb method and then I passed the whole DatabaseConfig to the database module as that is where it is consumed. Let me know what you think, it should making passing the settings for your PR quite trivial.

pallabpain: Yes, that makes things a lot easier. Thanks. 

I tried doing something similar in the PR https://github.com/juanfont/headscale/pull/1583/files#diff-826f8967b23df1209318fba82f94d8ed3eb12136b6cb2d070ddb76d9a059c336R125 but I guess some parts of it are getting refactored in other PRs.

kradalby: Yep, should not be to bad to rebase on top off. I'll try to get @juanfont to review so you can do yours on top.

pallabpain: Sounds good. I'll wait for the merge. Thanks, again.

juanfont (approved):

Issue #1572, opened 2023-10-11
headscale server stopped answering after a day of uptime despite listening of all the ports


axxonadmin:
Went to a local shop  and tried to connect to my remote headscale server v0.23.0-alpha1 I've got working several days ago.

Tailscale client (1.48.2 android) stuck at "connecting"
then I logged out and tried to log in, it stuck at this stage also with no output or warning
I got home, tried to connect from my desktop computer (1.51 windows 11) with no success (it stuck openning listen_addr and port in browser to authenticate)

Then I connected to my server over ssh and ran 'headscale apikeys list' and 'headscale nodes list' with no success, then I restarted headscale server with 'service headscale stop' / 'service headscale start' and everything started working just fine.

Can I do better for you to invetigate this issue if I get it next time?

What I've found so far:

```
# headscale nodes list
2023-10-10T06:50:37Z TRC DNS configuration loaded dns_config={"ExitNodeFilteredS                                                                                                                                                             et":null,"Nameservers":["1.1.1.1"],"Proxied":true,"Resolvers":[{"Addr":"1.1.1.1"                                                                                                                                                             }]}
Cannot get nodes: context deadline exceeded

```
```
# headscale apikeys list
2023-10-10T06:57:51Z TRC DNS configuration loaded dns_config={"ExitNodeFilteredSet":null,"Nameservers":["1.1.1.1"],"Proxied":true,"Resolvers":[{"Addr":"1.1.1.1"}]}
Error getting the list of keys: rpc error: code = DeadlineExceeded desc = context deadline exceeded

```
tcpdump on port 8080 (which is my 'listen_addr') showed smth like that:

```
listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
06:56:01.067801 IP server.http-alt > client.58443: Flags [.], ack 3966225333, win 501, length 0
06:56:01.128148 IP client.58443 > server.http-alt: Flags [.], ack 1, win 508, length 0
06:56:01.323836 IP server.http-alt > client.60098: Flags [.], ack 1688718613, win 501, options [nop,nop,TS val 3603882113 ecr 2274080242], length 0
06:56:01.388132 IP client.60098 > server.http-alt: Flags [.], ack 1, win 501, options [nop,nop,TS val 2274095346 ecr 3603867310], length 0
06:56:03.627793 IP server.http-alt > client.60094: Flags [.], ack 130672375, win 6146, options [nop,nop,TS val 3603884417 ecr 2274082546], length 0
[...]

PR #1701, opened 2024-01-27
Replace database locks with transactions

kradalby:
This commits removes the locks used to guard data integrity for the
database and replaces them with Transactions, turns out that SQL had
a way to deal with this all along.

This reduces the complexity we had with multiple locks that might stack
or recurse (database, nofitifer, mapper). All notifications and state
updates are now triggered _after_ a database change.

### DIFF
--- a/hscontrol/db/node.go
+++ b/hscontrol/db/node.go
@@ -34,22 +34,21 @@ var (
 	)
 )
 
-// ListPeers returns all peers of node, regardless of any Policy or if the node is expired.
 func (hsdb *HSDatabase) ListPeers(node *types.Node) (types.Nodes, error) {
-	hsdb.mu.RLock()
-	defer hsdb.mu.RUnlock()
-
-	return hsdb.listPeers(node)
+	return Read(hsdb.DB, func(rx *gorm.DB) (types.Nodes, error) {
+		return ListPeers(rx, node)
+	})
 }
 
-func (hsdb *HSDatabase) listPeers(node *types.Node) (types.Nodes, error) {
+// ListPeers returns all peers of node, regardless of any Policy or if the node is expired.
+func ListPeers(tx *gorm.DB, node *types.Node) (types.Nodes, error) {
 	log.Trace().
 		Caller().
 		Str("node", node.Hostname).
 		Msg("Finding direct peers")
 
 	nodes := types.Nodes{}
-	if err := hsdb.db.
+	if err := tx.
 		Preload("AuthKey").
 		Preload("AuthKey.User").
 		Preload("User").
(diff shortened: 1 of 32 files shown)


hscontrol/poll.go
```diff
@@ -337,7 +359,7 @@ func (h *Headscale) handlePoll(
 	// to receive a message to make sure we dont block the entire
 	// notifier.
 	// 12 is arbitrarily chosen.
```
dustinblackman, line 361: 🍰 120 now.

hscontrol/db/node.go
```diff
@@ -917,33 +797,12 @@ func (hsdb *HSDatabase) ExpireExpiredNodes(lastCheck time.Time) time.Time {
...
-	}
-
-	// Inform the node itself that it has expired.
-	for _, node := range expiredNodes {
-		stateSelfUpdate := types.StateUpdate{
-			Type:        types.StateSelfUpdate,
-			ChangeNodes: types.Nodes{node},
-		}
-		if stateSelfUpdate.Valid() {
-			hsdb.notifier.NotifyByMachineKey(stateSelfUpdate, node.MachineKey)
```
dustinblackman, line 944: ❓ This logic got moved out of the DB layer and up to the API level, right?

dustinblackman (commented):
I was reviewing the PR in preparation for https://github.com/juanfont/headscale/issues/1572#issuecomment-1932662449, I put down comments I had incase it was useful.

dustinblackman:
@kradalby I had this PR rebased on top of master running for a few hours, and did not have it lock up on me once! I have across two other issues though.

1. New nodes joining the network for the first time are unable to make requests over the network. `tailscale ping IP-ADDRESS` works as expected, but attempting to hit a service on port 80 across the network doesn't work.
2. Ephemeral nodes don't seem to expire properly. I have a node that is marked online (when it's not), expired, and the `expired_at` field continuously increments by 5 seconds every 5 seconds, so the node never goes away from the list in `headscale nodes list`

hscontrol/db/node.go
kradalby: Yep, because both the notifier and db has locks (or lock like things like transactions) which became easy to deadlock and hard to detect.

hscontrol/poll.go
kradalby: wops, fixed

kradalby:
@dustinblackman 

Thanks!

Could you file them as two new issues and tag this PR? It is so big already that I do not really want to dump more stuff onto it, but of  course that needs to be addressed. 

I think the ephemeral tests need to be improved and we need to find a way to replicate it as part of fixing it so we ensure there is no regression in the future.

For the port 80, I think the curling in the ACL tests would cover that, so it is strange that they are passing 🤔

Issue #1712, opened 2024-02-03
Preauthkey table reusable column is incorrect


dustinblackman:
## Bug description

The preauthkeys Reusable column is incorrectly showing `N/A` rather than `true` when a key is reusable. Not quite sure what the original intention was.

I have a patch ready to go. https://github.com/juanfont/headscale/pull/1713

## Environment

- Version of headscale used: v0.23.0-alpha3

## To Reproduce

```bash
headscale user create machine-1
headscale preauthkeys create -u machine-1 -e 1d --ephemeral --reusable
headscale preauthkeys list -u machine-1
# See N/A in reusable column
```

TotoTheDragon:
This seems to be correct with the current implementation of ephemeral keys. However the current implementation is wrong.

According to the following issue https://github.com/tailscale/tailscale/issues/5982
For a ephemeral key to be used multiple times, it needs to be reusable. But in the following code segment, we do not check if the key was used if it is reusable or ephemeral https://github.com/juanfont/headscale/blob/c3257e2146304c52e588c6de2fd28bcc0f13b1ad/hscontrol/db/preauth_keys.go#L199-L201

@kradalby could you check my logic on this?

dustinblackman:
I'm confused. The column is just supposed to expose whether the key was created with `--ephemeral` or not, no? It's just a listing the DB rows.

TotoTheDragon:
> I'm confused. The column is just supposed to expose whether the key was created with `--ephemeral` or not, no? It's just listing the DB rows.

It does, but in the case you described it is N/A with the current implementation. If you want to change the N/A, we should also fix the logic causing it to be N/A. Currently it does not matter if a key is reusable if it already is ephemeral. This is not the same behaviour as tailscale.

kradalby:
@TotoTheDragon I think you are right, 

looks like that line hasnt changed since it was written originally and https://github.com/juanfont/headscale/blob/c3257e2146304c52e588c6de2fd28bcc0f13b1ad/hscontrol/db/preauth_keys.go#L199 is wrong, or, its too simplified.

We need to check ephemeral keys in the same way too.

PR #1713, opened 2024-02-03
fix: Preauth key list reusable column

dustinblackman:
## Overview

Fixes https://github.com/juanfont/headscale/issues/1712

The preauthkeys Reusable column is incorrectly showing N/A rather than true when a key is reusable. Not quite sure what the original intention was, but this makes it more clear.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.mm

### DIFF
--- a/cmd/headscale/cli/preauthkeys.go
+++ b/cmd/headscale/cli/preauthkeys.go
@@ -107,13 +107,6 @@ var listPreAuthKeys = &cobra.Command{
 				expiration = ColourTime(key.GetExpiration().AsTime())
 			}
 
-			var reusable string
-			if key.GetEphemeral() {
-				reusable = "N/A"
-			} else {
-				reusable = fmt.Sprintf("%v", key.GetReusable())
-			}
-
 			aclTags := ""
 
 			for _, tag := range key.GetAclTags() {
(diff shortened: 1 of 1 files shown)


cmd/headscale/cli/preauthkeys.go
```diff
@@ -107,13 +107,6 @@ var listPreAuthKeys = &cobra.Command{
 				expiration = ColourTime(key.GetExpiration().AsTime())
 			}
 
-			var reusable string
-			if key.GetEphemeral() {
-				reusable = "N/A"
-			} else {
-				reusable = fmt.Sprintf("%v", key.GetReusable())
-			}
```
kradalby, line 115: A bit late evening, but I think this `if` is just the wrong way around, it should have the sprintf if it cant get it, and N/A if the node isnt.

I think it should be written like:

```go
reusable := "N/A"
if key.GetEphemeral() {
  reusable = "true" // or "yes"?
}
```

dustinblackman: This function just returns a boolean though. https://github.com/dustinblackman/headscale/blob/b75e537356c5b712dbae6778c64a1b6f9dbf5217/gen/go/headscale/v1/preauthkey.pb.go#L100-L105

Only chance it's nil is if the `key` is not defined. We could handle all three cases by checking if key is nil.

dustinblackman: Should we close this instead until this is solved? https://github.com/juanfont/headscale/issues/1712#issuecomment-1938491046

Issue #1705, opened 2024-01-30
Tailscale client logout with OIDC reconnect does not work (Linux and iOS clients)


felixn-unity:
## Bug description

When doing Tailscale logout reconnect does not work. It only works when node has been deleted and a fresh Tailscale login

## Environment

No docker og reverse-proxy

Headscale running native on debian installed with headscale_0.23.0-alpha3_linux_amd64.deb

Tailscale 1.58.2 on Ubuntu 22.04 (go version: go1.21.5) 
Tailscale 1.58.2 on iOS 17.2.1 iPone 13

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

## To Reproduce

7  | parsec-sto-p01 | parsec-sto-p01     | [eH3Gg]    | [2X1w2] | smilla | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2024-01-30 08:41:21 | 2024-02-04 07:48:16 | online    | no

tailscale logout

7  | parsec-sto-p01 | parsec-sto-p01     | [eH3Gg]    | [2X1w2] | smilla | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2024-01-30 08:42:36 | 2024-01-30 08:42:36 | offline   | yes

(journalctl -f -u headscale.service | grep parsec)

Jan 30 08:42:36 headscale headscale[26922]: 2024-01-30T08:42:36Z INF ../../../home/runner/work/headscale/headscale/hscontrol/poll.go:32 > The client has closed the connection node=parsec-sto-p01 node_key=[2X1w2] omitPeers=false readOnly=false stream=true
Jan 30 08:42:36 headscale headscale[26922]: 2024-01-30T08:42:36Z INF Client requested logout node=parsec-sto-p01
Jan 30 08:42:36 headscale headscale[26922]: 2024-01-30T08:42:36Z INF ../../../home/runner/work/headscale/headscale/hscontrol/auth.go:556 > Successfully logged out node=parsec-sto-p01
**Jan 30 08:42:40 headscale headscale[26922]: 2024-01-30T08:42:40Z ERR 🤮 Cannot expire node error="invalid value, should be pointer to struct or slice" name=parsec-sto-p01 node=parsec-sto-p01**

systemctl restart tailscaled.service && tailscale up --login-server https://headscale.xxxxx.yy --reset --force-reauth

After OIDC login, Tailscale client just hangs and no logs with node name...

[...]

Issue #1675, opened 2024-01-03
OIDC reauthentication hangs


redstonekasi:
## Bug description

When re-authenticating with an OIDC provider, the login never succeeds and instead infinitely hangs.

## Environment

- OS: Arch Linux
- Headscale version: 0.23.0-alpha2
- Tailscale version: 1.56.1

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

## To Reproduce

1. Setup a headscale instance using the mock OIDC server.
2. Login with tailscale.
3. Logout with tailscale.
4. Login with tailscale, watch it break.

Etran-H:
[https://github.com/juanfont/headscale/blob/3b103280efb78f4d705ec5ab6e96a6e37cd30231/hscontrol/oidc.go#L446C3-L446C3](url)
#1609 
Only the Machine key is determined, the Node key may be different

mhahl:
Can confirm. Replicated with version `headscale/headscale:0.23.0-alpha2` running on Fedora with `Tailscale 1.56.1`

PR #1719, opened 2024-02-05
Login with OIDC after having been logged out

TotoTheDragon:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes https://github.com/juanfont/headscale/issues/1675
Fixes https://github.com/juanfont/headscale/issues/1705

These issues started in the v0.23.0 alpha. 
When reauthenticating using oidc, no OldNodeKey is set by the client, the NodeKey set by the client is different from the one headscale has set, but the NodeKey does not get updated to the new one.

 A better fix might be to update the nodekey in the RegisterOIDC stack, but my attempts there did not work.

### DIFF
--- a/hscontrol/auth.go
+++ b/hscontrol/auth.go
@@ -199,6 +199,19 @@ func (h *Headscale) handleRegister(
 			return
 		}
 
+		// When logged out and reauthenticating with OIDC, the OldNodeKey is not passed, but the NodeKey has changed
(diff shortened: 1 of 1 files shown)


kradalby:
Thanks for tracking this down!

> A better fix might be to update the nodekey in the RegisterOIDC stack, but my attempts there did not work.

Would you mind creating a issue to track how this can potentially be one for later?

PR #1723, opened 2024-02-08
Add assert func for verifying status, netmap and netcheck

kradalby:
This PR adds three new integration test assert methods which will do the following for the "general" case:

`assertValidNetmap`: in clients which support `tailscale debug netmap` it will dump the netmap and verify that it is in a known "good" state, this includes checking for endpoints, derp relays, ips, online status and more.

`assertValidStatus`: check the output of the `tailscale status` command, checking derp relay, hostname, os, peer ips, network map status and magic sock status

`assertValidNetcheck`: checks that the client has a valid DERP server using output from `tailscale netcheck`.

The "general" case is any all to all communication, it does not account for things like nodes with ACLs and similar.

This also splits "PingAllByAll" into two, one version which use internal DERP for better use offline or dodgy internet, and one with public DERP.

### DIFF
--- a/hscontrol/poll.go
+++ b/hscontrol/poll.go
@@ -201,9 +201,15 @@ func (h *Headscale) handlePoll(
 			return
 		}
 
+		// TODO(kradalby): Figure out why patch changes does
+		// not show up in output from `tailscale debug netmap`.
+		// stateUpdate := types.StateUpdate{
+		// 	Type:          types.StatePeerChangedPatch,
+		// 	ChangePatches: []*tailcfg.PeerChange{&change},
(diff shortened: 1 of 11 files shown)


hscontrol/poll.go
```diff
@@ -190,9 +190,15 @@ func (h *Headscale) handlePoll(
 			return
 		}
 
+		// TODO(kradalby): Figure out why patch changes does
```
kradalby, line 193: I will work on this in a follow up so it doesnt become even bigger.

juanfont (approved):

PR #1733, opened 2024-02-10
TLS documentation updates

lgrn:
Move "Bring your own certificates" to the top
  since the letsencrypt section is now much longer, it seems wrong to
  keep such a short section way down at the bottom.

Restructure "Challenge types" into separate sections

Add technical description of letsencrypt renewals
  this aims to answer:
  - what can be expected in terms of renewals
  - what logs can be expected (none)
  - how to validate that renewal happened successfully
  - the reason for some of the 'acme/autocert' logs, or at least some best-effort assumptions

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/docs/tls.md
+++ b/docs/tls.md
@@ -1,8 +1,17 @@
 # Running the service via TLS (optional)
 
+## Bring your own certificate
+
+Headscale can be configured to expose its web service via TLS. To configure the certificate and key file manually, set the `tls_cert_path` and `tls_cert_path` configuration parameters. If the path is relative, it will be interpreted as relative to the directory the configuration file was read from.
+
+```yaml
+tls_cert_path: ""
+tls_key_path: ""
+```
+
 ## Let's Encrypt / ACME
 
-To get a certificate automatically via [Let's Encrypt](https://letsencrypt.org/), set `tls_letsencrypt_hostname` to the desired certificate hostname. This name must resolve to the IP address(es) headscale is reachable on (i.e., it must correspond to the `server_url` configuration parameter). The certificate and Let's Encrypt account credentials will be stored in the directory configured in `tls_letsencrypt_cache_dir`. If the path is relative, it will be interpreted as relative to the directory the configuration file was read from. The certificate will automatically be renewed as needed.
+To get a certificate automatically via [Let's Encrypt](https://letsencrypt.org/), set `tls_letsencrypt_hostname` to the desired certificate hostname. This name must resolve to the IP address(es) headscale is reachable on (i.e., it must correspond to the `server_url` configuration parameter). The certificate and Let's Encrypt account credentials will be stored in the directory configured in `tls_letsencrypt_cache_dir`. If the path is relative, it will be interpreted as relative to the directory the configuration file was read from.
 
 ```yaml
 tls_letsencrypt_hostname: ""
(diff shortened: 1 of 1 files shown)


docs/tls.md
```diff
@@ -11,21 +20,57 @@ tls_letsencrypt_cache_dir: ".cache"
...
-headscale can also be configured to expose its web service via TLS. To configure the certificate and key file manually, set the `tls_cert_path` and `tls_cert_path` configuration parameters. If the path is relative, it will be interpreted as relative to the directory the configuration file was read from.
+Headscale uses [autocert](https://pkg.go.dev/golang.org/x/crypto/acme/autocert), a Golang library providing [ACME protocol](https://en.wikipedia.org/wiki/Automatic_Certificate_Management_Environment) verification, to facilitate certificate renewals via [Let's Encrypt](https://letsencrypt.org/about/). Certificates will be renewed automatically, and the following can be expected:
 
-```yaml
-tls_cert_path: ""
-tls_key_path: ""
+* Certificates provided from Let's Encrypt have a validity of 3 months from date issued.
+* Renewals are only attempted by headscale when 30 days or less remains until certificate expiry.
+* Renewal attempts by autocert are triggered at a random interval of 30-60 minutes.
+* No log output is generated for failed or successful renewals.
```
ohdearaugustin, line 44: I'm not 100% sure about that. I think some output can be generated on failed renewals.

docs/tls.md
```diff
@@ -11,21 +20,57 @@ tls_letsencrypt_cache_dir: ".cache"
...
+
+#### Checking certificate expiry
+
+If you want to validate that certificate renewal completed successfully, this can be done either manually, or through external monitoring software. Two examples of doing this manually:
+
+1. Open the URL for your Headscale server in your browser of choice, and manually inspecting the expiry date of the certificate you receive.
+1. Or, check remotely from CLI using `openssl`:
```
ohdearaugustin, line 51: 2.

docs/tls.md
```diff
@@ -11,21 +20,57 @@ tls_letsencrypt_cache_dir: ".cache"
...
+acme/autocert: missing server name
+```
+
+Likely caused by an incoming connection that does not specify a hostname, for example a `curl` request directly against the IP of the server, or an unexpected hostname.
+
+```plaintext
+acme/autocert: host "[foo]" not configured in HostWhitelist
+```
+
+Similarly to the above, this likely indicates an invalid incoming request for an incorrect hostname, commonly just the IP itself.
```
ohdearaugustin, line 74: Similarly to the above, but this time a wrong hostname was used to resolve to your IP address. This will lead to an invalid request to headscale.

ohdearaugustin (commented):
Looks good to me. Thank you. 

Only one small thing I'm not sure about the error messages on failed renewal.

lgrn:
No problem, I can just wipe that section if we're not sure about the cause for them.

docs/tls.md
lgrn: Bad phrasing on my part, I've amended the commit to instead state that no log output is "generated when renewals are skipped, or successful", as this is what I've observed at least.

docs/tls.md
lgrn: I don't think it's strictly true that a "hostname was used", since the log lines I observed actually listed my external IP as the "host".

This is why I chose the general phrasing "incorrect hostname" since I was hoping that could be interpreted as both an incorrect but actual hostname, or that the request uses "ip-as-hostname" which seemed to be the case for my logs.

ohdearaugustin: For my log most of the time it was the case with a different hostname. I guess both cases are valid.
```
2024/02/13 18:02:59 http: TLS handshake error from<IP_reacted>:45250: acme/autocert: host "<ip_reacted>" not configured in HostWhitelist
```
and
```
acme/autocert: host "<subdomain>.your-server.de" not configured in HostWhitelist
```

ohdearaugustin:
BTW: We should run prettier on the file.

lgrn:
> BTW: We should run prettier on the file.

ok, I've never used it before but I installed it in my editor, applied it and committed it as b034d08

kradalby (approved):

PR #1756, opened 2024-02-15
new IP allocator and add postgres to integration tests.

### LINKED ISSUE #1755: LastInsertId is not supported by Postgres
## Bug description

Hi, I've tried to update to the v0.23 release and I received the following error: 

```log
{"level":"fatal","error":"LastInsertId is not supported by this driver","time":1708000343,"message":"Migration failed: LastInsertId is not supported by this driver"}
```

Just to make sure, I've updated Postgres to the latest 16.x version, but I'm still receiving this error message.

I'm not certain what `LastInsertId` should do, I'm guessing it should return the latest used sequential number on `INSERT`? Since it doesn't seem supported by Postgres, I would maybe suggest looking into `RETURNING` (see [PostgreSQL - INSERT](https://www.postgresql.org/docs/16/sql-insert.html)) if there is no simpler workaround.

> The optional `RETURNING` clause causes `INSERT` to compute and return value(s) based on each row actually inserted [...]. This is primarily useful for obtaining values that were supplied by defaults, such as a serial sequence number.

## Environment

- Version of headscale used: `headscale/headscale:0.23.0`; Postgres 15.x and 16.x
- [x] Headscale is behind a (reverse) proxy (`envoyproxy/envoy:v1.29.1`)

## To Reproduce

1. Have an existing (working) installation using PostgreSQL
2. Update the config to accommodate for the breaking changes (`db_*` -> `database.postgres.*`) and set the database to `postgres` again.
3. Update the command in docker from ´headscale serve` to `serve`
4. Deploy the new version
5. See error messages in logs

## Logs and attachments

(Headscale doesn't start)

### Host Headscale

- `Linux acme 5.10.0-27-amd64 #1 SMP Debian 5.10.205-2 (2023-12-31) x86_64 GNU/Linux`
- `Docker version 25.0.1, build 29cf629`

```plain
ID    NAME    CORES   CPU TYPE    ARCHITECTURE   MEMORY     DISK     STORAGE TYPE   TRAFFIC
1     cx11    1       shared      x86            2.0 GB     20 GB    local          20 TB
```

### Host Postgres

- `Linux pg01 6.1.0-17-arm64 #1 SMP Debian 6.1.69-1 (2023-12-30) aarch64 GNU/Linux`
- `Docker version 25.0.1, build 29cf629`

```plain
ID    NAME    CORES   CPU TYPE    ARCHITECTURE   MEMORY     DISK     STORAGE TYPE   TRAFFIC
93    cax21   4       shared      arm            8.0 GB     80 GB    local          20 TB
```

kradalby:
This PR adds a new IP allocator which should be more efficient, stable and less racy. This was necessary to deal with a discovered race condition in the allocation after adding postgresql integration tests.

I believe this is replacing the first "real" code I contributed to the project, and man, that was ver much needed 😆.

In addition it adds a new option to the integration tests to run them against postgresql instead of sqlite to ensure we do not have this regression again. 

The last commit is the generation of new gh actions jobs and can be ignored as generated.

Fixes #1755


hscontrol/grpcv1.go
```diff
@@ -98,6 +99,10 @@ func (api headscaleV1APIServer) ListUsers(
 		response[index] = user.Proto()
 	}
 
+	sort.Slice(response, func(i, j int) bool {
```
juanfont, line 102: good lord, yes.

juanfont (approved):

PR #1759, opened 2024-02-15
fix docker tagging

ohdearaugustin:
This should fix the tags for docker for prereleases. We don't want to push to latest on prerelease.

- [ x] read the [CONTRIBUTING guidelines](README.md#contributing)
- [ x] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -97,7 +97,7 @@ kos:
       - linux/arm64
       - linux/arm/v7
     tags:
-      - latest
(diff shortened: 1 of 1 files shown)


kradalby:
ah thanks, I missed this, I've added some more tags that was mentioned in other issues too in #1763.

We should probably not tag 0, 0.23 and so on with prerelease either.

PR #1763, opened 2024-02-16
rework docker tags

kradalby:
This commit tries to align the new docker tags with the old schema

A prerelease will end up with the following tags:

- unstable
- v0.23.0-alpha3
- 0.23.0-alpha3
- sha-1234adsfg

A release will end up with:

- latest
- stable
- v0.23.0
- v0.23
- v0
- 0.23.0
- 0.23
- 0
- sha-1234adsfg

All of the builds will also have a `-debug` version.

I think this addresses the feedback from discord and https://github.com/juanfont/headscale/issues/1561#issuecomment-1946243866.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -97,13 +97,17 @@ kos:
       - linux/arm64
       - linux/arm/v7
     tags:
-      - latest
-      - "{{ .Tag }}"
-      - "{{ .Major }}.{{ .Minor }}.{{ .Patch }}"
-      - "{{ .Major }}.{{ .Minor }}"
-      - "{{ .Major }}"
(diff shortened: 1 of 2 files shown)


kradalby:
@ohdearaugustin please sanity check.

ohdearaugustin (approved):
Looks sane to me

ohdearaugustin:
I guess we have to manually fix the tags from the last release.

juanfont (approved):

vbrandl:
Currently there is no `latest` tag at https://hub.docker.com/r/headscale/headscale/tags. Will this tag be created with the next proper release?

kradalby:
Likely, but we might avoid it as it is quite bad practise to use it.

PR #1790, opened 2024-02-23
docs/reverse-proxy: use standard map, correct X-Forwarded-Proto varia…

SuperSandro2000:
…ble for nginx

I tested this on NixOS. After switching from postgres back to sqlite and with the standard websocket proxy config, everything started to work.

Initially the none standard map which differs from upstreams docs https://nginx.org/en/docs/http/websocket.html got me very confused and I thought this would be somehow required.

Also $http_x_forwarded_proto doesn't exist, NixOS uses $scheme instead and I am going with that.

proxy_redirect also didn't work for me but that might be on my rather big config.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/docs/reverse-proxy.md
+++ b/docs/reverse-proxy.md
@@ -33,8 +33,7 @@ The following example configuration can be used in your nginx setup, substitutin
 
 ```Nginx
 map $http_upgrade $connection_upgrade {
-    default      keep-alive;
-    'websocket'  upgrade;
+    default      upgrade;
     ''           close;
 }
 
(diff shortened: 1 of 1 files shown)


kradalby:
I think this might be Linux distribution specific and different OS does it differently making it hard for us to say what is the correct way to have this in the doc. While I am a NixOS user, I think the majority use Debian/Ubuntu and it makes problematic to target one distro if they are not the same.

@ohdearaugustin could you have a look?

SuperSandro2000:
> I think this might be Linux distribution specific

Nope, http_x_forwarded_proto is nginx ingress controller specific. https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/advanced-configuration-with-annotations/

I am pretty confident that this config also works on the default Debian installation.

ohdearaugustin (approved):
This was already discussed in the accidentally closed PR

PR #1791, opened 2024-02-23
rework mapsession

kradalby:
This commit restructures the map session in to a struct
holding the state of what is needed during its lifetime.

For streaming sessions, the event loop is structured a
bit differently not hammering the clients with updates
but rather batching them over a short, configurable time
which should significantly improve cpu usage, and potentially
flakyness.

The use of Patch updates has been dialed back a little as
it does not look like its a 100% ready for prime time. Nodes
are now updated with full changes, except for a few things
like online status.

In addition, there is a lot of simplifications, removing internal states in the mappers, 
lots of error cleanups and more tests.


kradalby:
Known TODOs:
- There is a bug with Expiry, its sent very often.
- If a streaming mapsession is open, and a new one is attempted, the old one should be closed.

juanfont (approved):

PR #1792, opened 2024-02-24
feat: implements apis for managing headscale policy

pallabpain:
### Description
Headscale currently lacks APIs for managing ACLs.
The only way to manage ACLs is by loading them 
from a file, and any changes to the policy 
require reloading the Headscale process. This 
limitation makes it difficult to integrate 
Headscale with other systems via APIs, as 
there is no ACL management available.

This commit introduces two APIs allowing you to set the policy.

#### Set ACL Policy
```
PUT /api/v1/policy
```
Payload (Not JSON, so convert to string before invoking the API.)
```
{"policy": "{// This is a comment\n\"groups\":{\"g1\":[\"u1\",\"u2\"]},\"acls\":[]}"}
```
#### Get ACL Policy
```
GET /api/v1/policy
```
Response
```
{"policy": {"groups": {}, "tags": {}}}
```

### CLI
```
→ go run cmd/headscale/headscale.go --config ./config.yaml policy --help
Manage the Headscale ACL Policy

Usage:
  headscale policy [command]

Available Commands:
  get         Print the current ACL Policy
  set         Updates the ACL Policy

Flags:
  -h, --help   help for policy

Global Flags:
  -c, --config string   config file (default is /etc/headscale/config.yaml)
      --force           Disable prompts and forces the execution
  -o, --output string   Output format. Empty for human-readable, 'json', 'json-line' or 'yaml'

Use "headscale policy [command] --help" for more information about a command.
```

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

### Resolves
* https://github.com/juanfont/headscale/issues/582


hscontrol/app.go
```diff
@@ -931,6 +922,53 @@ func notFoundHandler(
...
+
+		pol, err := policy.LoadACLPolicyFromBytes(acl.Policy, "json")
+		if err != nil {
+			return pkgerrors.Wrap(err, "failed to load parse policy")
+		}
+
+		h.ACLPolicy = pol
+
+		log.Info().
+			Msg("ACL policy successfully reloaded, notifying nodes of change")
```
pallabpain, line 955: todo: handle the case when the policy is empty (IsZero)

pallabpain:
@evenh @kradalby Please take a look. 🙇🏼

evenh:
I might have some free time this evening to do a review

pallabpain:
I've attached a screen recording to the PR in case someone wishes to see how the APIs work.

hscontrol/types/acl.go
```diff
@@ -0,0 +1,45 @@
...
+	ErrACLPolicyUpdateIsDisabled = errors.New("update is disabled for modes other than 'db'")
+)
+
+// ACL describes the data model for ACLs used to restrict access to resources.
+type ACL struct {
+	ID     uint64 `gorm:"primary_key"`
+	Policy datatypes.JSON
+
+	CreatedAt  *time.Time
+	Expiration *time.Time
```
evenh, line 26: Is `Expiration` and `LastSeen` copy pasta? :-)

hscontrol/types/config.go
```diff
@@ -180,6 +189,8 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 	viper.AutomaticEnv()
 
+	viper.SetDefault("acl.policy_mode", ACLPolicyModeDB)
```
evenh, line 192: Wouldn't this break existing setups which already uses a file? I'm not a maintainer, but I would be fine with the DB being the default going forward but in that case we should not break existing users setups.

hscontrol/grpcv1.go
```diff
@@ -652,6 +654,64 @@ func (api headscaleV1APIServer) DeleteApiKey(
...
+// SetACL inserts or updates the ACL.
+func (api headscaleV1APIServer) SetACL(
+	_ context.Context,
+	request *v1.SetACLRequest,
+) (*v1.SetACLResponse, error) {
+	if api.h.cfg.ACL.PolicyMode != types.ACLPolicyModeDB {
+		return nil, types.ErrACLPolicyUpdateIsDisabled
+	}
+
+	polBytes, err := request.GetPolicy().MarshalJSON()
```
evenh, line 684: Would this fail if the input is HuJSON (default Tailscale ACL format)? `policy.LoadACLPolicyFromBytes` will use the hujson parser by default when the format is not `yaml`.

hscontrol/grpcv1.go
```diff
@@ -652,6 +654,64 @@ func (api headscaleV1APIServer) DeleteApiKey(
...
+	if err != nil {
+		return nil, err
+	}
+
+	a, err := policy.LoadACLPolicyFromBytes(polBytes, "json")
+	if err != nil {
+		return nil, errors.Wrap(err, types.ErrInvalidACLPolicyFormat.Error())
+	}
+
+	api.h.ACLPolicy = a
```
evenh, line 694: Should this line be moved down to after the ACL is successfully persisted?

hscontrol/app.go
```diff
@@ -761,25 +768,9 @@ func (h *Headscale) Serve() error {
 					Msg("Received SIGHUP, reloading ACL and Config")
 
 				// TODO(kradalby): Reload config on SIGHUP
```
evenh, line 770:
```suggestion
```

hscontrol/app.go
```diff
@@ -931,6 +922,58 @@ func notFoundHandler(
...
+
+		pol, err := policy.LoadACLPolicyFromBytes(acl.Policy, "json")
+		if err != nil {
+			return pkgerrors.Wrap(err, "failed to parse policy")
+		}
+
+		h.ACLPolicy = pol
+
+		log.Info().
+			Msg("ACL policy successfully reloaded, notifying nodes of change")
```
evenh, line 960:
```suggestion
			Msg("ACL policy successfully loaded, notifying nodes of change")
```

This function is both used for the initial loading and for the SIGHUP reloading.

hscontrol/app.go
```diff
@@ -931,6 +922,58 @@ func notFoundHandler(
...
+	default:
+		log.Warn().
+			Str("mode", string(h.cfg.ACL.PolicyMode)).
+			Msg("Unknown ACL policy mode")
+	}
+
+	if h.ACLPolicy != nil {
+		ctx := types.NotifyCtx(context.Background(), "acl-sighup", "na")
```
evenh, line 968: Do we want to notifyCtx + notifyAll if it's not a reload, but a normal startup? 🤔

evenh (commented):
Overall it looks good - I've left some inline comments. One open question is whether to support reading the current ACL via the new APIs/CLI commands even if the policy is backed by a file (for convenience)?

hscontrol/types/acl.go
pallabpain: Yes. 😅 I'll fix this.

hscontrol/types/config.go
pallabpain: Yes, this will be a breaking change. I can set up an alias for existing users. However, their setups may already be broken from the recent database config restructure. I'm of the opinion that we need not keep it backward compatible since users will have to update their configs anyway.

hscontrol/grpcv1.go
pallabpain: Thanks for bringing this up. I will test and verify it.

hscontrol/app.go
pallabpain: We do not. 😅 Thanks, I've updated this bit.

hscontrol/app.go
pallabpain: Moved the log close to the notify call.

hscontrol/grpcv1.go
pallabpain: Yes. I've moved it after we set the ACL in the database.

hscontrol/app.go
pallabpain: This still isn't implemented. I'll keep this for the time being.

pallabpain:
> Overall it looks good - I've left some inline comments. One open question is whether to support reading the current ACL via the new APIs/CLI commands even if the policy is backed by a file (for convenience)?

Enabled it. 👍🏼 Now it will work for both modes.

config-example.yaml
```diff
@@ -200,10 +200,20 @@ log:
...
+## ACL
+acl:
+  # The supported policy modes are "db" and "file".
+  # If "db" is selected, the policies will be stored in the database.
+  # You will be able to get and set the ACL via the ACL APIs.
+  #
+  # If "file" is selected, the policies will be loaded from a file
+  # located at `policy_path`. It will not be stored in the databse
+  # and you will not be able to get or set the ACL via the ACL APIs.
+  policy_mode: "db"
```
kradalby, line 212: I'm open for the example config to be `db`, but the default should still be file.

go.mod
```diff
@@ -96,6 +98,7 @@ require (
 	github.com/glebarez/go-sqlite v1.22.0 // indirect
 	github.com/go-jose/go-jose/v3 v3.0.1 // indirect
 	github.com/go-ole/go-ole v1.3.0 // indirect
+	github.com/go-sql-driver/mysql v1.7.0 // indirect
```
kradalby, line 101: Wonder why we suddenly need a mysql driver?

hscontrol/app.go
kradalby: Should probably not allow "large" TODOs like this without an issue... (including me)

hscontrol/app.go
```diff
@@ -931,6 +922,58 @@ func notFoundHandler(
...
 }
 
+func (h *Headscale) loadACLPolicy() error {
+	switch h.cfg.ACL.PolicyMode {
+	case types.ACLPolicyModeFile:
+		if h.cfg.ACL.PolicyPath != "" {
+			aclPath := util.AbsolutePathFromConfigPath(h.cfg.ACL.PolicyPath)
+			pol, err := policy.LoadACLPolicyFromPath(aclPath)
+			if err != nil {
+				return pkgerrors.Wrap(err, "failed to load ACL policy from file")
```
kradalby, line 942: Whats the difference between this error wrap compared to:

```
fmt.Errorf("failed to load ACL policy from file: %w", err)
```

hscontrol/types/acl.go
kradalby: GORM also  adds created at I believe, so that can also be omitted

hscontrol/types/config.go
kradalby: See comment earlier, I think file should still be the default. I'm open to be convinced.

hscontrol/types/config.go
```diff
@@ -376,10 +387,12 @@ func GetLogTailConfig() LogTailConfig {
 }
 
 func GetACLConfig() ACLConfig {
-	policyPath := viper.GetString("acl_policy_path")
+	policyPath := viper.GetString("acl.policy_path")
+	policyMode := viper.GetString("acl.policy_mode")
```
kradalby, line 391: Maybe we should just use the term policy?

```yaml
policy:
  mode: "" // <-- does "source" or "storage" or "backend" make more sense than mode?
  path: ""
```

hscontrol/types/acl.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+var (
+	ErrACLPolicyNotFound         = errors.New("acl policy not found")
+	ErrInvalidACLPolicyFormat    = errors.New("invalid policy format")
+	ErrACLPolicyUpdateIsDisabled = errors.New("update is disabled for modes other than 'db'")
+)
+
+// ACL describes the data model for ACLs used to restrict access to resources.
+type ACL struct {
```
kradalby, line 21: There was a proposal of having the policy versioned. I do not think we should do that now, but I think we should think of if there is anything about the format that should be addressed to enable that in the future. I do not have any concerns per now, but want to hear what you think.

kradalby (commented):
I think this looks reasonable, some comments and questions and to larger things:

I am tempted to use this opportunity to change the general phrasing from ACL to Policy through out the code, particularly since we are changing the config, adding commands etc.

I think the base format for the policy should be HuJSON, it is a bit more flexible format that allows preserving comments. This probably means using string/byte for the database and not the JSON type, but we dont really use it (or will use SQL to look up in the JSON) so that should not be a problem.
As part of this, I am willing to discuss dropping the YAML support to just streamline what we support.

pallabpain:
> I think this looks reasonable, some comments and questions and to larger things:
> 
> I am tempted to use this opportunity to change the general phrasing from ACL to Policy through out the code, particularly since we are changing the config, adding commands etc.
> 
> I think the base format for the policy should be HuJSON, it is a bit more flexible format that allows preserving comments. This probably means using string/byte for the database and not the JSON type, but we dont really use it (or will use SQL to look up in the JSON) so that should not be a problem. As part of this, I am willing to discuss dropping the YAML support to just streamline what we support.

Thanks for the review. This sounds like a good idea. I wrote this implementation to quickly demonstrate the APIs and commands that we can use as a reference to discuss the feature further.

Keeping the base format as `HuJSON` should be the way to proceed and we can store it as a bytearray or string in the db. Since we don't intend to query it any further. Although I'm not entirely sure how the API request and response with look like over RPC since I do not w[]ant to model the policy structure in the proto definition.

I am also okay with keeping the default mode as `file`. 

Let me update this PR with the suggestions and put it our for a second round of review.

evenh:
Hi @pallabpain. Have you had a chance to update the PR yet?

config-example.yaml
PizzaProgram: > I'm open for the example config to be db, but the default should still be file.

Personally I strongly **disagree**.
 
1. Many people still use docker.  (I know I know, it's not officially supported... still!) 
Manipulating files inside it, especially if the container is **not** running is a very hard task.

2. IMHO it is more **secure** to store ACL data inside a `db`. A separate file can be easier hacked / modified / viewed.
 After all, ACL is the "main security" config of HeadScale, telling who has access to what !
 Please keep in mind: The main reason for using VPNs is security. 
 Offering an easy-to-view / edit file to the open is against it.
 
3. Eventually more people will use this feature using a web-UI, so it would make much easier to setup with that already.

4. For those who like to do things manually (by typing text files, etc) -> changing the config to `file` will not be a big task to do, but for those who are not used to do tasks like this: it would be an "impossible nightmare".

Anyway: is the database password protected by default, so only HS can access it?
- or can other programs access it too?

hscontrol/types/acl.go
PizzaProgram: > having the policy versioned. 

I agree. Having plus one line of a simple `version=2024-03-24` is not a big deal now, 
but can save us lots of trouble later.

_(In my 30y. of experience I've realized: 
 - it is better to use ISO-dates for versions. 

Saves lot's of time and prevents confusion. 
Also better to merge with release-notes / what changed.)_

pallabpain:
> Hi @pallabpain. Have you had a chance to update the PR yet?

I haven't had the chance to update the PR, yet. I have some untested changes locally that I should be able to work on this weekend. I may change the request-response structure based on the newer comments.

IamTaoChen:
Maybe, we should create 2 new tables to store groups and tags. And sometimes, we just want to update groups or tags. So `/api/v1/acl/groups`, `/api/v1/acl/tags` and `/api/v1/acl/all` may be more useful.

I'm thinking about how to update ACL groups automatically, e.g. AD/LDAP or extract from OIDC (OIDC may be a problem because it can only get groups when users authenticate)

kradalby:
@pallabpain there has been quite some changes, do you think that you might have time to pick this up again?

pallabpain:
> @pallabpain there has been quite some changes, do you think that you might have time to pick this up again?

Sure. I'll review the recent changes and update the pull request. I'll try and get a working version ready soon.

gawsoftpl:
Hi, all this pull request is still active? I ask because I plan to create ui editor in package https://github.com/tale/headplane for update acl, and that apiendpoint will be great oportunity.

tale:
[Re this](https://github.com/juanfont/headscale/pull/1792#discussion_r1536788150)

> > I'm open for the example config to be db, but the default should still be file.
> 
> Personally I strongly **disagree**.
> 
> 1. Many people still use docker.  (I know I know, it's not officially supported... still!)
>    Manipulating files inside it, especially if the container is **not** running is a very hard task.
> 2. IMHO it is more **secure** to store ACL data inside a `db`. A separate file can be easier hacked / modified / viewed.
>    After all, ACL is the "main security" config of HeadScale, telling who has access to what !
>    Please keep in mind: The main reason for using VPNs is security.
>    Offering an easy-to-view / edit file to the open is against it.
> 3. Eventually more people will use this feature using a web-UI, so it would make much easier to setup with that already.
> 4. For those who like to do things manually (by typing text files, etc) -> changing the config to `file` will not be a big task to do, but for those who are not used to do tasks like this: it would be an "impossible nightmare".
> 
> Anyway: is the database password protected by default, so only HS can access it?
> 
> * or can other programs access it too?

As someone who develops a web UI for Headscale I don't think that the default should be db mode, *atleast not yet*. I don't understand the security standpoint here, if someone is able to get into a machine where the policy file is stored they can view your config which will contain much more sensitive information (ie. database credentials, secret keys, OIDC credentials, etc).

On the web UI end, I'd expect API requests to the ACL endpoints to just return status 403 or something if file mode is enabled. It's very easy to handle the status code and inform users on a web UI that they need to change their config.

pallabpain:
> Maybe, we should create 2 new tables to store groups and tags. And sometimes, we just want to update groups or tags. So `/api/v1/acl/groups`, `/api/v1/acl/tags` and `/api/v1/acl/all` may be more useful.
> 
> I'm thinking about how to update ACL groups automatically, e.g. AD/LDAP or extract from OIDC (OIDC may be a problem because it can only get groups when users authenticate)

Sounds like a good idea to implement. However, if we get and set the entire policy file contents, it keeps things simple instead of exposing APIs to do specific things on the policy. This shall also keep `headscale` in-line with the Tailscale ACL APIs and someone migrating from headscale to Tailscale SaaS can do it without making a lot of changes.

kradalby:
I would prefer keeping it simple and doing the whole thing now. Further extensions can be discussed later.

pallabpain:
> I think this looks reasonable, some comments and questions and to larger things:
> 
> I am tempted to use this opportunity to change the general phrasing from ACL to Policy through out the code, particularly since we are changing the config, adding commands etc.
> 
> I think the base format for the policy should be HuJSON, it is a bit more flexible format that allows preserving comments. This probably means using string/byte for the database and not the JSON type, but we dont really use it (or will use SQL to look up in the JSON) so that should not be a problem. As part of this, I am willing to discuss dropping the YAML support to just streamline what we support.

@kradalby Revised the PR. I've tried to keep the implementation simple in the first iteration. While I've updated the naming convention from ACL to Policy in some places, I've left some code untouched as of now. The default format will be `HuJSON` and the API will also support the same. 

Please take a look.

config-example.yaml
```diff
@@ -202,7 +202,9 @@ log:
 # Path to a file containing ACL policies.
 # ACLs can be defined as YAML or HUJSON.
 # https://tailscale.com/kb/1018/acls/
-acl_policy_path: ""
+policy:
+  mode: file
+  path: ""
```
kradalby, line 208:
```suggestion
  # mode can be either `file` or `database` which determine where headscale will try to read the 
  # Policy from.
  mode: file
  # If mode is `file`, a file path pointing to the Policy file.
  path: ""

```

hscontrol/app.go
```diff
@@ -1012,3 +1011,46 @@ func readOrCreatePrivateKey(path string) (*key.MachinePrivate, error) {
...
+		}
+
+		pol, err = policy.LoadACLPolicyFromBytes([]byte(p.Data), "hujson")
+		if err != nil {
+			return fmt.Errorf("failed to parse policy: %w", err)
+		}
+	default:
+		log.Warn().
+			Str("mode", string(h.cfg.Policy.Mode)).
+			Msg("Unknown ACL policy mode")
```
kradalby, line 1050: I think this should be a fatal

hscontrol/grpcv1.go
```diff
@@ -671,6 +674,73 @@ func (api headscaleV1APIServer) DeleteApiKey(
...
+	case types.PolicyModeDB:
+		p, err := api.h.db.GetPolicy()
+		if err != nil {
+			return nil, err
+		}
+
+		// We retain the HuJSON format of the policy while storing
+		// it in the database. But should we return the same in the
+		// get policy response? Or should we consider returning the
+		// policy in JSON format?
```
kradalby, line 691: I think HuJSON.

hscontrol/grpcv1.go
```diff
@@ -671,6 +674,73 @@ func (api headscaleV1APIServer) DeleteApiKey(
...
+
+		return &v1.GetPolicyResponse{
+			Policy:    p.Data,
+			UpdatedAt: timestamppb.New(p.UpdatedAt),
+		}, nil
+	case types.PolicyModeFile:
+		b, err := json.Marshal(api.h.ACLPolicy)
```
kradalby, line 698: I suppose if we return the content of the file instead of the loaded content, we could preserve comments.

hscontrol/types/config.go
```diff
@@ -38,6 +39,13 @@ const (
 	IPAllocationStrategyRandom     IPAllocationStrategy = "random"
 )
 
+type PolicyMode string
+
+const (
+	PolicyModeDB   = "db"
```
kradalby, line 45:
```suggestion
	PolicyModeDB   = "database"
```

kradalby (commented):
I think this is more or less there, a couple of small things.

I would also like to see a CLI integration test to validate the roundtrip:

- Load new ACL
- Read it out and compare
- Check that changes have a applied (if possible since we dont have hotreloading)

pallabpain:
> I think this is more or less there, a couple of small things.
> 
> I would also like to see a CLI integration test to validate the roundtrip:
> 
> * Load new ACL
> * Read it out and compare
> * Check that changes have a applied (if possible since we dont have hotreloading)

Thanks for the review. I'll implement the suggestions and write the tests.

kradalby:
Great, I think it would be great to try to get this into 0.23.0, I am looking to release a beta soonish so let me know if looking at this work before that is feasible, if not, there is no problem pushing it for later.

hscontrol/grpcv1.go
pallabpain: Since we support both YAML and HuJSON, the response may differ based on the file format used for the ACLs.

I'm updating this block to return the contents of the file. But we can consider dropping YAML support altogether. What do you think?

config-example.yaml
pallabpain: Done.

hscontrol/app.go
pallabpain: Done.

hscontrol/grpcv1.go
pallabpain: Thanks for confirming.

pallabpain:
> Great, I think it would be great to try to get this into 0.23.0, I am looking to release a beta soonish so let me know if looking at this work before that is feasible, if not, there is no problem pushing it for later.

@kradalby I've addressed the comments. I should be able to write the tests this week. But when is v0.23.0 planned for? Is there a due date or a cut-off date that you're looking at?

Please take a look at this comment as well: https://github.com/juanfont/headscale/pull/1792#discussion_r1649686735

hscontrol/grpcv1.go
kradalby: I'm open for that yes

kradalby:
Sounds good, no hard cutoff, I'm on holiday for two weeks, so staying a bit offline, if you make the test I'll pop in and review and we can do an alpha.

hscontrol/grpcv1.go
pallabpain: Sounds good. I'll go ahead and remove YAML support.

kradalby:
I'm back, and I would love to cut a beta this week, do you think it would be feasible to push some tests before midweek~ so we have time for a review cycle? I'll do a re-review of the code now.

hscontrol/app.go
```diff
@@ -1012,3 +1011,46 @@ func readOrCreatePrivateKey(path string) (*key.MachinePrivate, error) {
 
 	return &machineKey, nil
 }
+
+func (h *Headscale) loadACLPolicy() error {
+	var (
```
kradalby, line 1016: We should probably have a fatal if `acl_policy_path` is set so people are forced to migrate.
Alternative we could have a warn, and if it is set we default to:

```yaml
policy:
   mode: file
   path: <Same as `acl_policy_path`>
```

hscontrol/policy/acls.go
```diff
@@ -108,35 +107,22 @@ func LoadACLPolicyFromPath(path string) (*ACLPolicy, error) {
 		Bytes("file", policyBytes).
 		Msg("Loading ACLs")
 
-	switch filepath.Ext(path) {
-	case ".yml", ".yaml":
-		return LoadACLPolicyFromBytes(policyBytes, "yaml")
```
kradalby, line 113: Maybe we need to emit some sort of warning/fatal here too?

kradalby: I suppose it errors when it cant unmarshal, but maybe we need a good error message.

tale: It would make more sense as a consumer to say that it's no longer supported instead of just spewing out unmarshaling errors.

pallabpain: Thanks. Makes sense to throw a meaningful error. I've moved the errors in `LoadACLPolicyFromBytes`.

hscontrol/app.go
pallabpain: Thanks for adding the required piece of code. 🙇🏼

kradalby:
No problem, I'm available to take a stab at the integration tests now if you have not started. Maybe I could get it in shape to merge today.

kradalby:
@pallabpain I ended up working on another bug today, let me know if you will not have time to look at the integration tests and I'll have a look next week.

pallabpain:
> @pallabpain I ended up working on another bug today, let me know if you will not have time to look at the integration tests and I'll have a look next week.

I managed to find some time to look into the integration tests, tonight. While I understood how I'd go about implementing the tests, I may have to spend some time to figure out the right way to configure the headscale instance with the right policy mode. 😅

kradalby:
Sounds good, have a look at the hsic.With options, there you can configure things, use environment variables instead of file.

pallabpain:
> Sounds good, have a look at the hsic.With options, there you can configure things, use environment variables instead of file.

@kradalby I have managed to implement a simple test that sets a policy reads it back and then verifies the content. I had to expose `WriteFile` via the `ControlServer` interface to create a file for the `headscale policy set --file FILE` command.

Please take a look.

kradalby:
I think that looks great, what would be great is a test in the ACLs that do something like:

1. Start with two nodes and no ACL, using database mode
2. have them ping eachother
3. apply a ACL with the command that makes them unable to connect
4. verify they can no longer connect

I can have a look at that tomorrow if you do not have time to take a stab at it, let me know.

kradalby:
I've added the test as described, but I found a bug I dont know if is new or old,

when a new policy is set, the nodes reload, but one of the nodes gets a packet filter that is wide open "*:*", the node belonging to user1, which should get an empty filter. I will continue investigating.

kradalby:
Ok, I found the bug, it likely hasnt been discovered since _most_ people never reloaded ACLs without restarting the server.

If you set the packetfilter in tailcfg to an empty slice, it will marshall to `nil`, and that will cause it to be ignored, so the client will continue to use the previous filter:
https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L1863-L1865

I wonder if there is a way to get around this without copy and maintain a copy of the whole tailcfg 🤔 .

pallabpain:
> Ok, I found the bug, it likely hasnt been discovered since _most_ people never reloaded ACLs without restarting the server.
> 
> If you set the packetfilter in tailcfg to an empty slice, it will marshall to `nil`, and that will cause it to be ignored, so the client will continue to use the previous filter: https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L1863-L1865
> 
> I wonder if there is a way to get around this without copy and maintain a copy of the whole tailcfg 🤔 .

Thanks for adding the tests and resolving the issue. I can clean up the commits once the tests pass.

kradalby:
> Thanks for adding the tests and resolving the issue. I can clean up the commits once the tests pass.

Thank you, I think the last commit should resolve it, not sure why, but we can leave it for now.

pallabpain:
> > Thanks for adding the tests and resolving the issue. I can clean up the commits once the tests pass.
> 
> Thank you, I think the last commit should resolve it, not sure why, but we can leave it for now.

Sounds good. ✅ Let me clean up the commits, then.

pallabpain:
@kradalby I have cleaned up the commits. Should be good to merge now. Please take a look. 🙇🏼

kradalby (approved):
Thank you!, lots of work, but we got there, I think a lot of people will appreciate this feature!

hscontrol/app.go
```diff
@@ -8,7 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
-	_ "net/http/pprof" //nolint
+	_ "net/http/pprof" // nolint
```
juanfont, line 11: Is this staying?

config-example.yaml
```diff
@@ -203,10 +203,18 @@ log:
...
+# headscale supports Tailscale's ACL policies.
+# Please have a look to their KB to better
+# understand the concepts: https://tailscale.com/kb/1018/acls/
+policy:
+  # The mode can be "file" or "database" that defines
+  # where the ACL policies are stored and read from.
+  mode: file
+  # If the mode is set to "file", the
+  # path to a file containing ACL policies.
+  # The file can be in YAML or HuJSON format.
```
snh, line 216: I note the changelog mentioned that support for ACL policies in YAML format was dropped as part of this PR. Does this comment in `config-example.yaml` need updating to make it clear that only HuJSON is supported for file based now, or is the changelog incorrect?

PR #1799, opened 2024-02-29
simplify integration testing with matrix jobs

vdovhanych:
This PR significantly reduces the number of integration test workflow files. I modified the setup to use the GitHub matrix strategy instead of having 90+ generated workflow files.

I also tried to keep the original go script that generated the workflow files, but now, instead of generating whole workflow files, it adds test names to a matrix strategy. This, in turn, improves readability and updates to the test structure.

This will spawn a job for every combination, so if we fe. have `TestACLHostsInNetMapTable` and `database: [postgres, sqlite]` for the matrix strategy, this will create two jobs 

```
(TestACLHostsInNetMapTable, postgres)
(TestACLHostsInNetMapTable, sqlite)
```

Screenshot 2024-02-29 at 19 05 14

Or see how it will look in the PR for the already run tests.

Caveats to this setup:
 - requires you to have yq installed so you can generate test names for the workflow
 - yq will remove the empty line formatting from the GitHub file, so either fix it manually or commit it with those changes (it does not affect the functionality, but it's not as convenient to read as the current formatting)
 - the naming of the test is not that clear but can be improved to reflect better what it does( especially in regards to 0 or 1 for the usage of postgres)

I have checked that the failing tests are not connected to this change. The same issues are currently present with the previous setup.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


cmd/gh-action-integration-generator/main.go
```diff
@@ -125,61 +26,44 @@ func findTests() []string {
...
-	for _, test := range tests {
-		for _, postgres := range []bool{false, true} {
-			log.Printf("generating workflow for %s", test)
-
-			name := test
-			if postgres {
-				name = test + "-postgres"
-			}
+	yqCommand := fmt.Sprintf(
+		"yq eval '.jobs.integration-test.strategy.matrix.test = %s' ../../.github/workflows/test-integration.yaml -i",
```
kradalby, line 45: can you add `yq` to the flake.nix file so its automatically installed for the dev env? that way we should not run into an issue with needing it.

.github/workflows/test-integration.yaml
```diff
@@ -0,0 +1,114 @@
...
+          - TestSubnetRouteACL
+          - TestHeadscale
+          - TestCreateTailscale
+          - TestTailscaleNodesJoiningHeadcale
+          - TestSSHOneUserToAll
+          - TestSSHMultipleUsersAllToAll
+          - TestSSHNoSSHConfigured
+          - TestSSHIsBlockedInACL
+          - TestSSHUserOnlyIsolation
+        postgres: [0, 1]
```
kradalby, line 62: Would be neat if we could fix this naming right of the bat with `-sqlite` and `-postgres`

kradalby:
In general looks good, thanks for replacing my hacks with some other neater hacks 👍

.github/workflows/test-integration.yaml
vdovhanych: Yeah ill try doing it like this

```yaml
database: [postgres, sqlite]
```
and use the if commands for the variable so if postgres=1 if sqlite postgres=0

But maybe a better solution would be to change what code expects to get from the env.

vdovhanych: Ok i changed the workflow it will have the name `-postgres` or `-sqlite` and i change it to if for the env itself if its 1 or 0 :)

cmd/gh-action-integration-generator/main.go
vdovhanych: done

vdovhanych:
Sorry for one more force push, I realized that the `yq` I'm using is `yq-go` in the nix packages.

cmd/gh-action-integration-generator/main.go
vdovhanych: Little sidenote, is there any particular reason why `gofumpt` is not part of the devdeps in the nix shel?

kradalby:
hmm, thats some weird postgres failure, all at the same time.

kradalby:
Fix in #1802

kradalby:
I'm just throwing this out there since you have made a very nice change, It would be really nice to have a workflow that runs the generate command and if there is a new file that isnt checked in, it fails so people (me mostly) do not forget to add the new tests.

If you are interested in helping more, feel free to see if there are other CI stuff that can be improved, it is mostly something I do the shortest path to success because I really want to spend time on other things, as you can see with the original implementation. The only thing I ask is that its done with Nix so I can have it replicated locally.

vdovhanych:
I'll look at what could be improved over the weekend. 

About the workflow, you're talking about it. Do you mean it will check what tests are in the workflow `test-integration.yaml`, compare it with the tests folder, and fail if not all of the tests are in that workflow?

Happy to help with this project and relieve some of your time towards the features 😄

kradalby:
> I'll look at what could be improved over the weekend.

Thank you!

> About the workflow, you're talking about it. Do you mean it will check what tests are in the workflow `test-integration.yaml`, compare it with the tests folder, and fail if not all of the tests are in that workflow?

I think the easiest way would be to do something like:

- git checkout
- cd cmd/gh-generator
- go generate
- git status <-- fail if there are any new files

> Happy to help with this project and relieve some of your time towards the features 😄

Thank you, I would say this is in the category of "we need it" but not "core business" 😄

vdovhanych:
I added the check you suggested. ✔️  
Also, if you look at the `test-integration` workflow, it is now formatted stupidly, but it is a limitation of yq, and for the check to work, we need to go with how the yq formats it, so we are left with the diff that's actually relevant, for the use case.

kradalby:
Great! Ppstfix is in so a rebase should make it happier

cmd/gh-action-integration-generator/main.go
kradalby: No it should be, I must have missed it, that can be done in a followup.

Currently we are lacking a bit on jobs to verify the formatting, so a job similar to the "is the tests generated" would potentially be beneficial for golines + gofumpt.

.github/workflows/check-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+      - uses: DeterminateSystems/nix-installer-action@main
+        if: steps.changed-files.outputs.any_changed == 'true'
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+        if: steps.changed-files.outputs.any_changed == 'true'
+
+      - name: Generate and check integration tests
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: |
+          nix develop --run "cd cmd/gh-action-integration-generator && go generate"
```
kradalby, line 35: I wonder if this needs to be:

```
nix develop --run -- "cd cmd/gh-action-integration-generator && go generate"
```

vdovhanych: noo its supposed to be `nix develop --command` I'm used to nix-shell where its `--run` sorry will fix right away

vdovhanych: fixed

vdovhanych:
Just noticed, maybe go ahaed and cancell the previous waiting checks for aproovement as it now cancells these latest ones and its now running not the final setup.

vdovhanych:
I dont know what is going on, it keeps me spamming with the notifications i think i managed to confuse GitHub 😆

kradalby (approved):

PR #1801, opened 2024-02-29
fix: keep the update channel from being cleaned up

### LINKED ISSUE #1798: Headscale removes node after it was re-added
## Bug description

It seems there is an issue with nodes that are hopping between different networks. The most common scenario is a mobile device with a mobile network and WiFi connections. So when the node switches network, it sends an update and Headscale adds a new update channel for it. The problem is, that the new update channel rewrites the old one in the notifier instance before the cleanup of the old one takes place. So when it's time to close and delete the old channel, Headscale in fact removes the new one. I believe it's because there is no place for multiple update channels for a single node. If the node is fast enough to send a new update before the cleanup, the old channel will be overwritten, and the deferred remove procedure takes down the new one instead. Thus the node ends up offline and receives no updates from this point.

## Environment

- OS: Ubuntu 22.04 5.15.0-97-generic
- Headscale version: v0.23.0-alpha5
- Tailscale version: v1.60.0, but could be any actually

## To Reproduce

Swap any node between networks, make it send new updates. Observe how update channels change

## Logs and attachments

I've came up with some silly and naive quick-fix and I don't think it's a good one, because it may fix the outcome, but the original problem. Anyway, I haven't tested it yet, it's just an idea.

Firstly, we add a channel counter for each machine. 

```golang
type Notifier struct {
	l         sync.RWMutex
	nodes     map[string]chan<- types.StateUpdate
	connected map[key.MachinePublic]bool
	hits      map[key.MachinePublic]int // channel counter goes here
}

func NewNotifier() *Notifier {
	return &Notifier{
		nodes:     make(map[string]chan<- types.StateUpdate),
		connected: make(map[key.MachinePublic]bool),
		hits:      make(map[key.MachinePublic]int),
	}
}
```

Then we increment this counter in the AddNode function

```golang
func (n *Notifier) AddNode(machineKey key.MachinePublic, c chan<- types.StateUpdate) {
	log.Trace().Caller().Str("key", machineKey.ShortString()).Msg("acquiring lock to add node")
	defer log.Trace().
		Caller().
		Str("key", machineKey.ShortString()).
		Msg("releasing lock to add node")

	n.l.Lock()
	defer n.l.Unlock()

	n.nodes[machineKey.String()] = c
	n.connected[machineKey] = true
	n.hits[machineKey]++  // increment counter

	log.Trace().
		Str("machine_key", machineKey.ShortString()).
		Int("open_chans", len(n.nodes)).
		Int("chan_hits", n.hits[machineKey]).
		Msg("Added new channel")
}
```

And finally, we decrement the counter in the RemoveNode and remove the node completely only if it was the last one

```golang
func (n *Notifier) RemoveNode(machineKey key.MachinePublic) {
	log.Trace().Caller().Str("key", machineKey.ShortString()).Msg("acquiring lock to remove node")
	defer log.Trace().
		Caller().
		Str("key", machineKey.ShortString()).
		Msg("releasing lock to remove node")

	n.l.Lock()
	defer n.l.Unlock()

	if len(n.nodes) == 0 {
		return
	}

        // only decrement counter if there was more than one channel
	if n.hits[machineKey] > 1 {
	    n.hits[machineKey]--

        log.Trace().
            Str("machine_key", machineKey.ShortString()).
                Int("open_chans", len(n.nodes)).
                Int("chan_hits", n.hits[machineKey]).
                Msg("Held channel")

            return;
        }

        // remove the channel if it's the last one indeed
	delete(n.nodes, machineKey.String())
	n.connected[machineKey] = false
	n.hits[machineKey] = 0;

	log.Trace().
		Str("machine_key", machineKey.ShortString()).
		Int("open_chans", len(n.nodes)).
		Msg("Removed channel")
}
```

eNdiD:
- add counter in Notifier for every machineKey
- do not remove channel if counter is more than 1

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #1798


kradalby:
I dont think this is relevant after the last betas, thanks.

PR #1812, opened 2024-03-05
add PKCE verifier for OIDC

Rorical:
https://www.rfc-editor.org/rfc/rfc7636.html
To fix the error "Could not exchange code for the token" when using the PKCE method, a verifier should be generated and used during the authentication process. 

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

## Summary by CodeRabbit

- **New Features**
	- Enhanced OpenID Connect (OIDC) registration and callback process with Proof Key for Code Exchange (PKCE) for improved security.
	- Introduced a new data structure, `RegistrationInfo`, to manage OIDC-related information effectively.

- **Bug Fixes**
	- Improved validation processes to ensure secure handling of authorization code exchanges.

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -364,6 +364,11 @@ unix_socket_permission: "0770"
 #   allowed_users:
 #     - alice@example.com
 #
+#   # Optional: Enable PKCE (Proof Key for Code Exchange) support for enhanced security
+#   # and prevent CSRF attacks.
+#   # See https://datatracker.ietf.org/doc/html/rfc7636
+#   enable_pkce: false
+#
 #   # Map legacy users from pre-0.24.0 versions of headscale to the new OIDC users
 #   # by taking the username from the legacy user and matching it with the username
 #   # provided by the OIDC. This is useful when migrating from legacy users to OIDC
--- a/docs/ref/oidc.md
+++ b/docs/ref/oidc.md
@@ -45,6 +45,17 @@ oidc:
   allowed_users:
     - alice@example.com
 
+  # Optional: PKCE (Proof Key for Code Exchange) configuration
+  # PKCE adds an additional layer of security to the OAuth 2.0 authorization code flow
+  # by preventing authorization code interception attacks
+  pkce:
+    # Enable or disable PKCE support (default: false)
+    enabled: false
+    # PKCE method to use:
+    # - plain: Use plain code verifier
+    # - S256: Use SHA256 hashed code verifier (default, recommended)
+    method: S256
+
   # If `strip_email_domain` is set to `true`, the domain part of the username email address will be removed.
   # This will transform `first-name.last-name@example.com` to the user `first-name.last-name`
   # If `strip_email_domain` is set to `false` the domain part will NOT be removed resulting to the following
--- a/hscontrol/oidc.go
+++ b/hscontrol/oidc.go
@@ -28,12 +28,14 @@ import (
 )
 
 const (
-	randomByteSize = 16
+	randomByteSize           = 16
+	defaultOAuthOptionsCount = 3
 )
 
 var (
 	errEmptyOIDCCallbackParams = errors.New("empty OIDC callback params")
 	errNoOIDCIDToken           = errors.New("could not extract ID Token for OIDC callback")
+	errNoOIDCRegistrationInfo  = errors.New("could not get registration info from cache")
 	errOIDCAllowedDomains      = errors.New(
 		"authenticated principal does not match any allowed domain",
 	)
@@ -47,11 +49,17 @@ var (
 	errOIDCNodeKeyMissing = errors.New("could not get node key from cache")
 )
 
+// RegistrationInfo contains both machine key and verifier information for OIDC validation.
+type RegistrationInfo struct {
+	MachineKey key.MachinePublic
+	Verifier   *string
+}
+
 type AuthProviderOIDC struct {
 	serverURL         string
 	cfg               *types.OIDCConfig
 	db                *db.HSDatabase
-	registrationCache *zcache.Cache[string, key.MachinePublic]
+	registrationCache *zcache.Cache[string, RegistrationInfo]
 	notifier          *notifier.Notifier
 	ipAlloc           *db.IPAllocator
 	polMan            policy.PolicyManager
@@ -87,7 +95,7 @@ func NewAuthProviderOIDC(
 		Scopes: cfg.Scope,
 	}
 
-	registrationCache := zcache.New[string, key.MachinePublic](
+	registrationCache := zcache.New[string, RegistrationInfo](
 		registerCacheExpiration,
 		registerCacheCleanup,
 	)
@@ -157,18 +165,35 @@ func (a *AuthProviderOIDC) RegisterHandler(
 
 	stateStr := hex.EncodeToString(randomBlob)[:32]
 
-	// place the node key into the state cache, so it can be retrieved later
-	a.registrationCache.Set(
-		stateStr,
-		machineKey,
-	)
+	// Initialize registration info with machine key
+	registrationInfo := RegistrationInfo{
+		MachineKey: machineKey,
+	}
 
-	// Add any extra parameter provided in the configuration to the Authorize Endpoint request
-	extras := make([]oauth2.AuthCodeOption, 0, len(a.cfg.ExtraParams))
+	extras := make([]oauth2.AuthCodeOption, 0, len(a.cfg.ExtraParams)+defaultOAuthOptionsCount)
+	// Add PKCE verification if enabled
+	if a.cfg.PKCE.Enabled {
+		verifier := oauth2.GenerateVerifier()
+		registrationInfo.Verifier = &verifier
 
+		extras = append(extras, oauth2.AccessTypeOffline)
+
+		switch a.cfg.PKCE.Method {
+		case types.PKCEMethodS256:
+			extras = append(extras, oauth2.S256ChallengeOption(verifier))
+		case types.PKCEMethodPlain:
+			// oauth2 does not have a plain challenge option, so we add it manually
+			extras = append(extras, oauth2.SetAuthURLParam("code_challenge_method", "plain"), oauth2.SetAuthURLParam("code_challenge", verifier))
+		}
+	}
+
+	// Add any extra parameters from configuration
 	for k, v := range a.cfg.ExtraParams {
 		extras = append(extras, oauth2.SetAuthURLParam(k, v))
 	}
+
+	// Cache the registration info
+	a.registrationCache.Set(stateStr, registrationInfo)
 
 	authURL := a.oauth2Config.AuthCodeURL(stateStr, extras...)
 	log.Debug().Msgf("Redirecting to %s for authentication", authURL)
@@ -203,7 +228,7 @@ func (a *AuthProviderOIDC) OIDCCallbackHandler(
 		return
 	}
 
-	idToken, err := a.extractIDToken(req.Context(), code)
+	idToken, err := a.extractIDToken(req.Context(), code, state)
 	if err != nil {
 		http.Error(writer, err.Error(), http.StatusBadRequest)
 		return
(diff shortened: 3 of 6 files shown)


kradalby:
I imagine this can go in and be rebased on top of #2020, whenever that is ready.

kradalby:
@Rorical if you have time to rebase this in the next few days, then it might be able to make it into the next release.

Rorical:
Testing and documentation will be added soon

config-example.yaml
```diff
@@ -369,6 +369,11 @@ unix_socket_permission: "0770"
 #   allowed_users:
 #     - alice@example.com
 #
+#   # Optional: Enable PKCE (Proof Key for Code Exchange) support for enhanced security
+#   # and prevent CSRF attacks.
+#   # See https://datatracker.ietf.org/doc/html/rfc7636
+#   enable_pkce: false
```
kradalby, line 375: If it is default off, I think we can leave it uncommented.

Are there any requirements to use this on the OIDC provider side? Should this just be on by default? what is the pro/con for having it as a default?

I think also this should be called:
```
pkce:
  enabled: false
```
to align with our config language.

hscontrol/oidc.go
```diff
@@ -157,15 +164,32 @@ func (a *AuthProviderOIDC) RegisterHandler(
...
+
+	// Add PKCE verification if enabled
+	if a.cfg.EnablePKCE {
+		verifier := oauth2.GenerateVerifier()
+		registrationInfo.Verifier = &verifier
+		extras = append(extras,
+			oauth2.AccessTypeOffline,
+			oauth2.S256ChallengeOption(verifier),
+		)
+	}
```
kradalby, line 187: This prealloc logic seems a bit unnecessary from a readability vs actual performance impact, I'm in favour of a simpler implementation:

```suggestion
	var extras []oauth2.AuthCodeOption

	// Add PKCE verification if enabled
	if a.cfg.EnablePKCE {
		verifier := oauth2.GenerateVerifier()
		registrationInfo.Verifier = &verifier
		extras = append(extras,
			oauth2.AccessTypeOffline,
			oauth2.S256ChallengeOption(verifier),
		)
	}
```

hscontrol/oidc.go
```diff
@@ -449,8 +485,8 @@ func (a *AuthProviderOIDC) createOrUpdateUserFromClaim(
 	// look it up by username. This should only be needed once.
 	// This branch will presist for a number of versions after the OIDC migration and
 	// then be removed following a deprecation.
-	// TODO(kradalby): Remove when strip_email_domain and migration is removed
```
kradalby, line 452: This change looks unrelated

integration/auth_oidc_test.go
```diff
@@ -524,6 +524,114 @@ func TestOIDC024UserCreation(t *testing.T) {
...
+
+	sort.Slice(listUsers, func(i, j int) bool {
+		return listUsers[i].Id < listUsers[j].Id
+	})
+
+	if diff := cmp.Diff(want, listUsers, cmpopts.IgnoreUnexported(v1.User{}), cmpopts.IgnoreFields(v1.User{}, "CreatedAt")); diff != "" {
+		t.Fatalf("unexpected users: %s", diff)
+	}
+
+	// Test that PKCE verifier was properly used in authentication
```
kradalby, line 617: How does this test verify that the PKCE works properly? this test looks more or less the same as the normal test, but with the option enabled?

I would expect two test cases:
- One with the option enabled, and where the client is joined correctly
- One with the option enabled, where the verification fails and the client is not joined

Rorical: MockOIDC seems automatically handle code_challenge parameter for pkce, so I expect that normal test should work in this case.

kradalby: Thats fine, but we need to make sure that a node is denied if the pkce is wrong/not valid to ensure that the feature works.

Rorical: PKCE verifier is used by OIDC provider to check the validity of client. E.G. if any man-in-middle attack is performed to sniff the authorization code, then it can be used to get access token. Do you mean that "pkce is wrong/not valid" refers that we fake a pkce verifier in headscale to check if the OIDC provider can deny that request?

docs/ref/oidc.md
```diff
@@ -45,11 +45,16 @@ oidc:
...
   # by preventing authorization code interception attacks
-  # See https://datatracker.ietf.org/doc/html/rfc7636
-  enable_pkce: false
+  pkce:
+    # Enable or disable PKCE support (default: false)
+    enabled: false
+    # PKCE method to use:
+    # - plain: Use plain code verifier
+    # - S256: Use SHA256 hashed code verifier (default, recommended)
+    method: S256
```
kradalby, line 57: Can we call this SHA256? is there a reason why it is S256?

kradalby: Apparently the standard, ignore this.

kradalby:
@Rorical any opportunity to look at the integration tests that are failing? You probably also need to rebase to get a docker fix that broke all the builds.

Rorical:
Apologize for the delay. The failing test proves to be harder to implement for me, as I can not get the nix test working for me.

kradalby:
No worries, can you do:
```
cd cmd/gh-action-integration-generator
go generate
```

and commit the updated workflow? this will run the new tests, I will get to reviewing a bit later, thank you!

Rorical:
The test should works now. Is it possible to still make this change into v0.24.0? I really need this feature to have my instance continue working...

kradalby:
If the tests pass, I can try to review it tomorrow, but I am hesitant adding it this late in the release cycle and adding new features in beta, so I won't guarantee it.

config-example.yaml
kradalby: Good, the example needs to be updated

integration/auth_oidc_test.go
```diff
@@ -534,6 +539,211 @@ func TestOIDC024UserCreation(t *testing.T) {
...
+		log.Printf("RoundTrip error: %v", err)
+
+		return nil, err
+	}
+	log.Printf("Response status: %s", resp.Status)
+
+	return resp, err
+}
+
+func TestOIDCAuthenticationWithPKCEVerifierTampering(t *testing.T) {
```
kradalby, line 685: I think the challenge with this test is that it doesnt verify that the test work, since the positive test uses the tailscale client, while the negative test does something completely different by running a http client. 

I am conflicted, I am not sure if this tests does what is needed, but not sure how we can achieve it with the client.

kradalby (commented):
I am not really sure about the tests, but I did test this with my setup and `kanidm` and it does work.

integration/auth_oidc_test.go
Rorical: Well, actually the negative test does something completely same as the positive test did, however the only difference is that I tried to modify the request from user to oidc server (with a custom http transport). The whole redirection logic is completely same:

1. User request login from tailscale
2. tailscale request headscale, headscale return machinekey `headscale.instance/register/mkey:xxxxxxxxxxx`
3. user request register endpoint, the endpoint redirect user to oidc login `oidc.instance/oidc/authorize?xxxxxx`
4. oidc server process request and redirect to headscale `headscale.instance/oidc/callback?xxxxxx`

What the http client did is that it modify the parameter `code_challenge` for the third step, so the login request from the headscale redirection is invalid. This is what we did to mimic MiTM. The oidc server will still proceed, but the response will not pass the code verifier stored inside headscale instance.

This is an actual log generated by the test:
```
2024/12/19 16:33:53 running tailscale up for user user1
2024/12/19 16:33:53 ts-unstable-dnh7da running tailscale up
2024/12/19 16:34:23 ts-unstable-dnh7da login url: https://172.22.0.5:8080/register/mkey:809f6ad985579a4ccabf7ee7880e20c7efb049c6f952dd174b6c93f617c50d1f
2024/12/19 16:34:23 ts-unstable-dnh7da logging in with url
2024/12/19 16:34:23 RoundTrip: GET https://172.22.0.5:8080/register/mkey:809f6ad985579a4ccabf7ee7880e20c7efb049c6f952dd174b6c93f617c50d1f
2024/12/19 16:34:23 Processing GET request
2024/12/19 16:34:23 No code_challenge found in GET query params
2024/12/19 16:34:23 Response status: 302 Found
2024/12/19 16:34:23 RoundTrip: GET http://172.22.0.3:34663/oidc/authorize?access_type=offline&client_id=superclient&code_challenge=6cg09A-VPrE-wgpCxUtfArY6ammErUF4vP3A2LXsm1M&code_challenge_method=S256&redirect_uri=https%3A%2F%2Fhs-oidcauthpkce-lrbl4l%3A8080%2Foidc%2Fcallback&response_type=code&scope=openid+profile+email&state=dcf4af7abc68da1876db3617094bf63b
2024/12/19 16:34:23 Processing GET request
2024/12/19 16:34:23 Found GET verifier: 6cg09A-VPrE-wgpCxUtfArY6ammErUF4vP3A2LXsm1M
2024/12/19 16:34:23 Modified URL to: http://172.22.0.3:34663/oidc/authorize?access_type=offline&client_id=superclient&code_challenge=6cg09A-VPrE-wgpCxUtfArY6ammErUF4vP3A2LXsm1M_tampered&code_challenge_method=S256&redirect_uri=https%3A%2F%2Fhs-oidcauthpkce-lrbl4l%3A8080%2Foidc%2Fcallback&response_type=code&scope=openid+profile+email&state=dcf4af7abc68da1876db3617094bf63b
2024/12/19 16:34:23 Response status: 302 Found
2024/12/19 16:34:23 RoundTrip: GET https://hs-oidcauthpkce-lrbl4l:8080/oidc/callback?code=bzi7BK5ZeYWCS3J7Es0PTIkKZ5riN4yP&state=dcf4af7abc68da1876db3617094bf63b
2024/12/19 16:34:23 Processing GET request
2024/12/19 16:34:23 No code_challenge found in GET query params
2024/12/19 16:34:23 Response status: 400 Bad Request
2024/12/19 16:34:23 ts-unstable-dnh7da response code of oidc login request was 400 Bad Request
2024/12/19 16:34:23 body: could not exchange code for token: oauth2: "invalid_grant" "Invalid code verifier. Code challenge did not match hashed code verifier."
2024/12/19 16:34:23 auth got error: status code not OK
```

kradalby: Unless I am really missing something, I think what bothers me is that it isnt the tailscale client that is doing the request, but a `httpClient.Do(req)`. I think the "correct" way to do this is to have the request go through a proxy and that could be passed as an option to hsic or tsic. where you can give it a func that will modify the requests.

That way we can have unmodified (and only one version of) `runTailscaleUp`, and have failing and passing nodes in the same test, some just get tampered with and some dont, compared to a dedicated test for each.

This is a lot of code that is very specific to one test, and a proxy in between headscale and tailscale for tampering would be quite useful. 

I dont want to put it on you to have to write that for this test tho, so I wonder if we will do the following: 

- Can you squash up the commits so there is two commits:
  - one with the code change, changelog and doc/config changes
  - one with the tests
- Make a PR with only the code change commit
- Make a draft PR with the tests
- Make an issue about adding a tampering proxy for headscale integration tests and tag the draft PR.

Then we can merge the code change without tests for now, and do it in a generic way so we dont have to maintain a lot of specific code.

config-example.yaml
```diff
@@ -364,6 +364,11 @@ unix_socket_permission: "0770"
 #   allowed_users:
 #     - alice@example.com
 #
+#   # Optional: Enable PKCE (Proof Key for Code Exchange) support for enhanced security
```
kradalby, line 367: This part needs updating to align with the config.go and docs.

kradalby (changes requested):

integration/auth_oidc_test.go
Rorical: Ok, I see your point. by that way we can have a proxy which could also be useful for some other tests.

Rorical:
I suggest to include the positive test for the code change PR, since it does not require the proxy feature

kradalby:
Yes let’s keep the positive one, thanks
 Sent from Proton Mail for iOS 
On Sun, Dec 22, 2024 at 15:39, Rorical ***@***.***&gt; wrote:   I suggest to include the positive test for the code change PR, since it does not require the proxy feature —

Rorical:
I'm trying to open a new PR for this.

Rorical:
Closed due to #2314

PR #1815, opened 2024-03-06
Defaults to deny all rather than accept all if no acl policy defined

### LINKED ISSUE #1814: Default to a FilterDenyAll instead of tailcfg.FilterAllowAll if no acl policy defined
## Why

If no acl policy is defined then headscale defaults to `tailcfg.FilterAllowAll` which allows all devices visibility of all other devices and the ability to communicate with them on any protocol. This complete openness does not appear to fit with a zero trust network design. A more robust default security response that locks down device visibility and communication would be a more appropriate response.

## Description

Proposal is to do following in `hscontrol > policy > acts.go`:
 - add a new `tailcfg.FilterRule` of `FilterDenyAll` (as no appropriate policy appears to be available within tailscale);
 - update the `GenerateFilterAndSSHRules` so that if there is no policy defined headscale now:
    - logs a warning message of `Default deny all ACL rules being applied`;
    - defaults to the `FilterDenyAll` rather than `tailcfg.FilterAllowAll`.

The proposed `FilterDenyAll` would be:
```
var FilterDenyAll = []tailcfg.FilterRule{
	{
		SrcIPs:  []string{},
		SrcBits: nil,
	},
}
```

andy-netltd:
- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

Fixes #1814.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ after improving the test harness as part of adopting [#1460](https://github.com/
   - `/var/lib/headscale` and `/var/run/headscale` is no longer created automatically, see [container docs](./docs/running-headscale-container.md)
 - Prefixes are now defined per v4 and v6 range. [#1756](https://github.com/juanfont/headscale/pull/1756)
   - `ip_prefixes` option is now `prefixes.v4` and `prefixes.v6`
+- If no acl policy defined now defaults to deny all rather than allow all for device visibility and ability to communicate on any protocol. [#1814](https://github.com/juanfont/headscale/pull/1814)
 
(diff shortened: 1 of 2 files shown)


kradalby:
Tailscale has a default allow policy and we will adhere to the same as upstream.

aemaeth2501:
Hello,
Although I fully understand the will to adhere to tail scale concepts:
- there is no "reject" capabilities in ACL ATM which renders complicated access control
- can't this change of policy be configurable through an option in the config file?

Please see no offense in my remark, I love this soft !

PR #1823, opened 2024-03-12
feat: derpmap field in config

gdraganic:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Currently DERP map can be set only from URLs or local yaml config paths. This PR adds a simple and optional `DERPMap` config field which allows setting initial DERP map which is useful when using headscale programatically. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration capabilities with the addition of a `DERPMap` field in the DERP configuration.
	- Improved logic in the DERP map retrieval process to prevent runtime errors by excluding nil values.

### DIFF
--- a/hscontrol/derp/derp.go
+++ b/hscontrol/derp/derp.go
@@ -82,6 +82,9 @@ func mergeDERPMaps(derpMaps []*tailcfg.DERPMap) *tailcfg.DERPMap {
 
 func GetDERPMap(cfg types.DERPConfig) *tailcfg.DERPMap {
 	var derpMaps []*tailcfg.DERPMap
+	if cfg.DERPMap != nil {
+		derpMaps = append(derpMaps, cfg.DERPMap)
+	}
 
 	for _, path := range cfg.Paths {
 		log.Debug().
--- a/hscontrol/types/config.go
+++ b/hscontrol/types/config.go
@@ -179,6 +179,7 @@ type DERPConfig struct {
 	STUNAddr                           string
 	URLs                               []url.URL
 	Paths                              []string
+	DERPMap                            *tailcfg.DERPMap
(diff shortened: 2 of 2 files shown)


kradalby:
Hi sorry, I've missed this, you can already load DERP config via file, so another method seems excessive, I don't think we will include this.

Tpuljak:
@kradalby I understand it can be loaded from a file, but that's a bit awkward when using headscale programmatically (as a lib). It would require the application to first write a file and then load it. Additionally, it provides more overhead if the configuration has to be updated after it was already written.

From the looks of it, the PR wouldn't undermine loading the map from file, it simply allows users to add an initial derp map as an argument of the config.

kradalby:
When you say lib, do you mean as part of a go program? because in that case you can just import it and override yourself. 

I still dont think we will include this, sorry.

Tpuljak:
> When you say lib, do you mean as part of a go program? because in that case you can just import it and override yourself.

[This](https://github.com/daytonaio/daytona/blob/main/pkg/server/headscale/config.go#L23) is how we set the initial config in Daytona and the use it to initialize the server [here](https://github.com/daytonaio/daytona/blob/main/pkg/server/headscale/server.go#L68). It would be great if we could just add DERPMap to the initial cfg as well.

kradalby:
Ok fair, sorry, in my head I thought it would expose it to users via the config, as long as we dont let users set it via config.yaml, it is ok.

kradalby (approved):

Tpuljak:
> Ok fair, sorry, in my head I thought it would expose it to users via the config, as long as we dont let users set it via config.yaml, it is ok.

Thanks!

PR #1854, opened 2024-03-31
docs: Add FAQ question about using the same machine as server and client

### LINKED ISSUE #1852: Add documentation about using the same machine as client and server
## Why

As a user of tailscale that wants to try headscale, I have not been able to find information on whether it is possible to use the same server as both a headscale server and a tailscale client at the same time. (And about whether it is a good/bad idea to do this).

My use case would be having a single VM in some cloud that I run Headscale on, but also setting up Gitea on the same VM and connecting to it via the Tailnet.

## Description

My proposal would be an FAQ question that looks like this:

**Can/Should I use the same machine as both server and client?**
Fantasy answer: _Yes, this is no problem, just install headscale and tailscale according to the instructions and use `--login-server=localhost`._

Riezebos:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #1852

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -53,4 +53,4 @@ For convenience, we also build Docker images with `headscale`. But **please be a
 We don't know. We don't use reverse proxies with `headscale` ourselves, so we don't have any experience with them. We have [community documentation](https://headscale.net/reverse-proxy/) on how to configure various reverse proxies, and a dedicated [Discord channel](https://discord.com/channels/896711691637780480/1070619818346164324) where you can ask for help to the community.
 
 ## Can I use headscale and tailscale on the same machine?
-Yes, this is no problem. Install Headscale and Tailscale according to the instructions and use `tailscale up --login-server localhost`.
+Running headscale on a machine that is also in the tailnet might cause problems with routing and MagicDNS - at very least. It should be extremely discouraged.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -52,5 +52,5 @@ For convenience, we also build Docker images with `headscale`. But **please be a
 
 We don't know. We don't use reverse proxies with `headscale` ourselves, so we don't have any experience with them. We have [community documentation](https://headscale.net/reverse-proxy/) on how to configure various reverse proxies, and a dedicated [Discord channel](https://discord.com/channels/896711691637780480/1070619818346164324) where you can ask for help to the community.
 
-## Can/Should I use the same machine as both server and client?
+## Can I use headscale and tailscale on the same machine?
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -51,3 +51,6 @@ For convenience, we also build Docker images with `headscale`. But **please be a
 ## Why is my reverse proxy not working with Headscale?
 
 We don't know. We don't use reverse proxies with `headscale` ourselves, so we don't have any experience with them. We have [community documentation](https://headscale.net/reverse-proxy/) on how to configure various reverse proxies, and a dedicated [Discord channel](https://discord.com/channels/896711691637780480/1070619818346164324) where you can ask for help to the community.
+
+## Can/Should I use the same machine as both server and client?
```
ohdearaugustin, line 55:
```suggestion
## Can I use headscale and tailscale on the same machine?
```

ohdearaugustin (changes requested):
Rephrased for clearer meaning.

ohdearaugustin:
Closes #1852

juanfont:
I think this is a bad idea. 

Running headscale on a machine that is also in the tailnet might cause problems with routing and MagicDNS - at very least. It should be _extremely_ discouraged.

Riezebos:
@juanfont thanks! That is exactly the kind of answer to this question I was hoping to find in the FAQ. I updated the PR to match what you wrote.

docs/faq.md
```diff
@@ -51,3 +51,6 @@ For convenience, we also build Docker images with `headscale`. But **please be a
 ## Why is my reverse proxy not working with Headscale?
 
 We don't know. We don't use reverse proxies with `headscale` ourselves, so we don't have any experience with them. We have [community documentation](https://headscale.net/reverse-proxy/) on how to configure various reverse proxies, and a dedicated [Discord channel](https://discord.com/channels/896711691637780480/1070619818346164324) where you can ask for help to the community.
+
+## Can I use headscale and tailscale on the same machine?
+Running headscale on a machine that is also in the tailnet might cause problems with, at least, routing and MagicDNS. It should be extremely discouraged.
```
juanfont, line 56:
```suggestion
Running headscale on a machine that is also in the tailnet can cause problems with subnet routers, traffic relay nodes, and MagicDNS. It might work, but it is not supported.
```

ohdearaugustin (approved):

juanfont:
@Riezebos can you run prettier in the faq.md file?

juanfont:
thanks!!

PR #1870, opened 2024-04-13
feat: add config to overwrite grpc certificate 

### LINKED ISSUE #1709: Split TLS options for gRPC and HTTP
## Why
Currently for a remote CLI you have no choice but to setup TLS. As a result, the public HTTP endpoints also end up encrypted, and in some configurations this sacrifices significant observability (currently my linkerd2 sidecar containers are unable to inspect traffic, monitor endpoint performance, and enforce per-route ACLs). While workarounds are possible (for example running headscale without TLS, and a proxy terminating TLS for gRPC), they would be far from ideal.

## Description
One solution is adding split TLS options - to allow a user to enable TLS for gRPC, while not enabling it for other endpoints.

I'm happy to work on this if it's an acceptable change.

K3das:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

This adds two config options to overwrite the TLS certificate for the gRPC API

```yaml
# Use separate a certificate for gRPC, this overwrites
# the global certificate.
grpc_tls_cert_path: ""
grpc_tls_key_path: ""
```

(closes #1709)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1728093190,
-        "narHash": "sha256-CAZF2NRuHmqTtRTNAruWpHA43Gg2UvuCNEIzabP0l6M=",
+        "lastModified": 1729850857,
+        "narHash": "sha256-WvLXzNNnnw+qpFOmgaM3JUlNEH+T4s22b5i2oyyCpXE=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "e2f08f4d8b3ecb5cf5c9fd9cb2d53bb3c71807da",
+        "rev": "41dea55321e5a999b17033296ac05fe8a8b5a257",
         "type": "github"
       },
       "original": {


config-example.yaml
```diff
@@ -40,6 +40,11 @@ grpc_listen_addr: 127.0.0.1:50443
 # are doing.
 grpc_allow_insecure: false
 
+# Use separate a certificate for gRPC, this overwrites
```
ohdearaugustin, line 43:
```suggestion
# Use separate a x509 certificate for gRPC, this overwrites
```

ohdearaugustin (approved):
Looks good to me.

kradalby:
Overwrite sounds incorrect, it's more like using separate certs for grpc.

What is the use case for this, why would you want it?

It will also need some sort of integration tests.

K3das:
I used "overwrite" since that would imply that, by default, it would use certificates defined in the other settings, I'm happy to change this wording.

My use case is described in the attached issue; I don't want to configure TLS for the HTTP API endpoints, but configure it for the gRPC endpoints.

I'll add integration tests if you still consider this an acceptable change.

config-example.yaml
```diff
@@ -40,6 +40,11 @@ grpc_listen_addr: 127.0.0.1:50443
...
+# the global certificate.
+grpc_tls_cert_path: ""
+grpc_tls_key_path: ""
```
kradalby, line 46: At this point, with the growing amount of options it should end up like:
```suggestion
grpc:
  allow_insecure: false
  tls:
    cert_path: ""
    tls_key_path: ""
```

K3das: agreed, though this would create breaking changes, which @ohdearaugustin asked to avoid

kradalby: It should be possible to map the allow insecure to the new value.

K3das:
> I do not know exactly how I want them to look or how to test something like this, so this might be a challenging exercise in it self, but if you want to take a stab at it, you are welcome to do so.

Not sure either, but this could probably involve working on a better test suite for the TLS implementation, since from what I understand, ACME is completely untested. Alternatively, integration tests could generate and trust a certificate for gRPC like they do for the HTTP API, then just check the fingerprint(?).

Outside of CLI tests, are there tests for gRPC at all?

integration/embedded_derp_test.go
```diff
@@ -43,6 +43,7 @@ func TestDERPServerScenario(t *testing.T) {
 		hsic.WithExtraPorts([]string{"3478/udp"}),
 		hsic.WithEmbeddedDERPServerOnly(),
 		hsic.WithTLS(),
+		hsic.WithGrpcTLS(),
```
kradalby, line 46: This looks reasonable, but please add a new test so we dont change already known ones.

K3das:
@kradalby just to confirm, is it fine that this is breaking gRPC config. Would it be a better idea to put this with the TLS settings and break those (by making them a map)?

kradalby:
> @kradalby just to confirm, is it fine that this is breaking gRPC config. Would it be a better idea to put this with the TLS settings and break those (by making them a map)?

It should not require breaking it, you should be able to set up an alias in cobra/viper to handle it.

K3das:
Alright, I'll see what I can do, still, would it be a better idea to keep this with TLS configs instead?

kradalby:
> Alright, I'll see what I can do, still, would it be a better idea to keep this with TLS configs instead?

Sorry, forgot to answer this. Might be, I kind of want this feature to be "available, but not accessible", as I mentioned earlier that I am concerned people would use it in the wrong way. Maybe "hiding" it in tls makes sense.

K3das:
closing this for now, I don't have a significant need for this, and given new changes it would require a bit of effort to update this patch.

PR #1877, opened 2024-04-17
Log available update as warning

casdr:
This will result in the message being logged to stderr. Logging this to stdout may break some scripts that use the output of headscale commands.

- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md


kradalby (approved):

masterwishx:
Updated to alpha10 , this wrn should exist ?

image

casdr:
I think this is an issue in hashicorp/go-version, where 10 is lower than 9 because it starts with a 1.

masterwishx:
> I think this is an issue in hashicorp/go-version, where 10 is lower than 9 because it starts with a 1.

Sorry not checked, just informed of the case, Thanks for your pr

PR #1893, opened 2024-04-18
Add test stage to docs

juanfont:
I reckon we currently have the doc generation broken, add test to see.


kradalby (approved):

ohdearaugustin:
It seems like that the mkdocs-material was too old, updated it to 9.5.18. I guess I will add poetry in another PR, so it is easier to update dependencies and also to build the docs locally easier.

PR #1906, opened 2024-04-22
remove examples/, and kustomize

kradalby:
this directory is unmaintained and not verified, if it should be restored, it should end up under the community docs effort.


ohdearaugustin (approved):
As @kradalby mentioned, this configuration is unmaintained since 2021. We do not have the resources to verify it. Furthermore it would belong into the docs.

Issue #1817, opened 2024-03-07
headscale 0.23.0-alpha5 ignores /32 masks in ACLs


winterheart:
## Bug description

We have tailnet network with node1 as subnet router to `172.16.0.0/24` network.

After updating to 0.23.0-alpha5 some of our users lost access to resources provided by ACL. Here short example of acl.yaml:

```
groups:
  group:access:
    - user1
hosts:
  net.vlan0: 172.16.0.0/24
  net.dns01: 172.16.0.21/32
# ...
acls:
  - action: accept
    src:
      - group:access
    dst:
      - tag:access-servers:* # node1 in it
      - net.dns01:*
```

Users in group `access` (user1) now cannot access to 172.16.0.21. If I change `net.dns01` with `net.vlan0` in ACL, user1 can ping 
172.16.0.21 (and other addresses in 172.16.0/24 too which is undesirable). Previous version 0.22.3 was allowed access to hosts with /32 mask.

## Environment

- OS: Linux
- Headscale version: 0.23.0-alpha5
- Tailscale version: 1.56, 1.60

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

## To Reproduce

Have subnet router and ACLs to networks behind it.

## Logs and attachments

netmap with acl /32 in config:

```
...
                {
                        "SrcIPs": [
                                "100.64.0.3/32",
                                "fd7a:115c:a1e0::b/128"
                        ]
                        "DstPorts": [
                                {
                                        "IP": "100.64.0.2/32",
                                        "Bits": null,
                                        "Ports": {
                                                "First": 0,
                                                "Last": 65535
                                        }
                                },
                                {
                                        "IP": "fd7a:115c:a1e0::2/128",
                                        "Bits": null,
                                        "Ports": {
                                                "First": 0,
                                                "Last": 65535
                                        }
                                },
                        ]
                 }
```

netmap with acl /24 in config:

```
...
                {
                        "SrcIPs": [
                                "100.64.0.3/32",
                                "fd7a:115c:a1e0::b/128"
                        ]
                        "DstPorts": [
                                {
                                        "IP": "100.64.0.2/32",
                                        "Bits": null,
                                        "Ports": {
                                                "First": 0,
                                                "Last": 65535
                                        }
                                },
                                {
                                        "IP": "fd7a:115c:a1e0::2/128",
                                        "Bits": null,
                                        "Ports": {
                                                "First": 0,
                                                "Last": 65535
                                        }
[...]

PR #1914, opened 2024-04-26
fix issue #1817

mangareader:
fix issue #1817


kradalby:
Hi, thank you, this issue will be addressed in #1917 which also contains a lot of tests to make sure it does not regress again.

PR #1920, opened 2024-05-01
fix preauth key logging in as previous user

### LINKED ISSUE #1885: Headscale logs in user A with user B's auth key
## Bug description

Logging in a node with a preauthkey logs in as the previous user that was logged in on the same node, i.e. after performing the following three steps on the same node:

1. Login with user A's key
2. Logout A
3. Login with user B's key

the result is that user _A_ is logged in even though user _B_'s key is used. See full terminal session with the steps to reproduce below. I noticed this with release 0.22.3 when changing the user for some nodes and although the new key was used the same old user was still logged in. The MWE below is with v0.23.0-alpha7.

## Environment

- OS: Ubuntu 22.04
- Headscale version: v0.23.0-alpha7
- Tailscale version: 1.64.0

## To Reproduce

Install headscale on a clean machine (i.e. no existing config or database)
```
$ sudo apt install ./headscale_0.23.0-alpha7_linux_amd64.deb
[...]

$ sudo systemctl enable headscale

$ sudo systemctl start headscale
```

Create users `foo` and `bar` with associated keys:
```
$ sudo headscale users create foo

$ sudo headscale users create bar

$ sudo headscale preauthkeys create --user foo --reusable
73c546e371611141ef442e7aeaff918734708124072d530f

$ sudo headscale preauthkeys create --user bar --reusable
6b00073a57c0828e2cb5dcb4c64e2eef6de0972507f6478c
```

Login using `foo`'s key works as expected:
```
$ sudo tailscale up --reset --login-server=http://localhost:8080 --auth-key=73c546e371611141ef442e7aeaff918734708124072d530f

$ tailscale status
100.64.0.1      myhost               foo          linux   -

$ sudo headscale nodes list
ID | Hostname | Name   | MachineKey | NodeKey | User | IP addresses                  | Ephemeral | Last seen           | Expiration          | Connected | Expired
1  | myhost   | myhost | [Rl+ED]    | [YHvFJ] | foo  | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2024-04-17 14:09:32 | 0001-01-01 00:00:00 | online    | no
```

Logout `foo` and login with `bar`'s key:
```
$ sudo tailscale logout

$ sudo tailscale up --reset --login-server=http://localhost:8080 --auth-key=6b00073a57c0828e2cb5dcb4c64e2eef6de0972507f6478c
```

Even though the key belongs to `bar` we are logged in as `foo` according to both client and server:
```
$ tailscale status
100.64.0.1      myhost               foo          linux   -

$ sudo headscale nodes list
ID | Hostname | Name   | MachineKey | NodeKey | User | IP addresses                  | Ephemeral | Last seen           | Expiration          | Connected | Expired
1  | myhost   | myhost | [Rl+ED]    | [UgilM] | foo  | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2024-04-17 14:10:17 | 0001-01-01 00:00:00 | online    | no
```

Both keys are marked as used:
```
$ sudo headscale preauthkeys list -u foo
ID | Key                                              | Reusable | Ephemeral | Used | Expiration          | Created             | Tags
1  | 73c546e371611141ef442e7aeaff918734708124072d530f | true     | false     | true | 2024-04-17 15:07:54 | 2024-04-17 14:07:54 |

$ sudo headscale preauthkeys list -u bar
ID | Key                                              | Reusable | Ephemeral | Used | Expiration          | Created             | Tags
2  | 6b00073a57c0828e2cb5dcb4c64e2eef6de0972507f6478c | true     | false     | true | 2024-04-17 15:07:58 | 2024-04-17 14:07:58 |
```

## Logs and attachments
- ACL configuration: N/A
- Headscale configuration: Default `/etc/headscale/config.yml` installed with .deb.

kradalby:
Fixes #1885

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -26,6 +26,7 @@ jobs:
           - TestPreAuthKeyCommand
           - TestPreAuthKeyCommandWithoutExpiry
           - TestPreAuthKeyCommandReusableEphemeral
+          - TestPreAuthKeyCorrectUserLoggedInCommand
           - TestApiKeyCommand
           - TestNodeTagCommand
           - TestNodeAdvertiseTagNoACLCommand
--- a/hscontrol/auth.go
+++ b/hscontrol/auth.go
@@ -315,13 +315,16 @@ func (h *Headscale) handleAuthKey(
 
 		node.NodeKey = nodeKey
 		node.AuthKeyID = uint(pak.ID)
-		err := h.db.NodeSetExpiry(node.ID, registerRequest.Expiry)
+		node.Expiry = &registerRequest.Expiry
+		node.User = pak.User
+		node.UserID = pak.UserID
+		err := h.db.DB.Save(node).Error
 		if err != nil {
 			log.Error().
 				Caller().
 				Str("node", node.Hostname).
 				Err(err).
-				Msg("Failed to refresh node")
+				Msg("failed to save node after logging in with auth key")
 
 			return
 		}
@@ -344,7 +347,7 @@ func (h *Headscale) handleAuthKey(
 		}
 
 		ctx := types.NotifyCtx(context.Background(), "handle-authkey", "na")
-		h.nodeNotifier.NotifyWithIgnore(ctx, types.StateUpdateExpire(node.ID, registerRequest.Expiry), node.ID)
+		h.nodeNotifier.NotifyAll(ctx, types.StateUpdate{Type: types.StatePeerChanged, ChangeNodes: []types.NodeID{node.ID}})
 	} else {
 		now := time.Now().UTC()
 
--- a/integration/cli_test.go
+++ b/integration/cli_test.go
@@ -461,8 +461,8 @@ func TestPreAuthKeyCorrectUserLoggedInCommand(t *testing.T) {
 		t.Fatalf("expected node to be logged in, backend state: %s", status.BackendState)
 	}
 
-	if status.Self.UserID.String() != "2" {
-		t.Fatalf("expected node to be logged in as user2, got: %s", status.Self.UserID.String())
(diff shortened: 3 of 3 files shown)


juanfont (approved):

ItsShadowCone:
Could you confirm #1885 was a dupe of #1310 I.e. Both are fixed now?

kradalby:
@ItsShadowCone based on the description it looks like it is, but I find it a bit hard to understand the description, it would be helpful if you test and help me confirm.

ItsShadowCone:
My observation in #1310 was based on fast user switching and logging in via preauth keys, I.e. The purported use case was a "personal" account for admins via OIDC as well as an "admin" account with network access to the entire network via preauth key and fast user switching to limit attack surface when not performing administrative tasks.

Looking through the code, the old version seemed to see that a node (I.e. "instance of tailscaled") was already registered by matching the old node key and refreshing only expiry date and node key, not user key, thereby leading to issues with *any* user switching, fast or not. 

Now I don't know specifically what the tailscale client does when fast user switching, but I would naively assume the node key stays the same (if there was a new node key, we would never have noticed this issue).

Reading naively into the code this fix would then not fully allow fast user switching since the old node-user key combo is swapped out for the new one, effectively logging out the previous user in the process.

I can test it when I get to it, but what was the logic behind reusing the same node entry instead of registering a new one on every login? In my naive understanding, login should add a new node-user pair, logout should expire it immediately as administrative "revoke this login" should (this is iirc only implemented via cli, If anything).

The last point of consideration is what are the differences between preauth login and OIDC? From a naive understanding both should only assert user identity while the rest of the "login" follows a common code path, but this is apparently not the case. Would this be a possible goal for one of the next alphas?

Most of what I wrote are assumptions or vague observations, please correct me where I'm wrong 😉

Issue #1911, opened 2024-04-22
alpha9 takes 90 seconds to stop


mooncaptain108:
## Bug description
After upgrading from alpha8 to alpha9 I found that the stop increased from almost instantly to 90 seconds.
`systemctl stop headscale`
here's the journalctl log

> Apr 22 12:05:09 headscale systemd[1]: Stopping headscale coordination server for Tailscale...
> Apr 22 12:05:09 headscale headscale[4098]: 2024-04-22T12:05:09-05:00 INF Received signal to stop, shutting down gracefully signal=terminated
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: State 'stop-sigterm' timed out. Killing.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4098 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4099 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4100 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4101 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4102 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4103 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4104 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Killing process 4105 (headscale) with signal SIGKILL.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Main process exited, code=killed, status=9/KILL
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Failed with result 'timeout'.
> Apr 22 12:06:39 headscale systemd[1]: Stopped headscale coordination server for Tailscale.
> Apr 22 12:06:39 headscale systemd[1]: headscale.service: Consumed 33.940s CPU time.

## Environment

- OS: ubuntu 22.04
- Headscale version: v0.23.0-alpha9
- Tailscale version:  1.64.0

- [x ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

## To Reproduce

[...]

PR #1922, opened 2024-05-02
trace log and notifier shutdown

kradalby:
Updates #1911


juanfont (approved):

mooncaptain108:
I added a comment in #1911

PR #1934, opened 2024-05-09
re-construct OIDC config and flatten keycloak groups

IamTaoChen:
* use the `claims_map` to extract the designed information by the name which was a fixed map.  the `username` needs to be careful because now OIDC uses `claim.username` as USERNAME, if you want to keep same as before (Email), please set  `username=email` .
* combine `allowed_domains`,`allowed_groups`,`allowed_userd` to `allowed`
* add `misc` to save the random thing. if you set `misc.flatten_groups=true`, it will try to flatten the groups. this is for keycloak which group format is "/group/subgroup". 
* The `misc.strip_email_domain` only works when the `username` is email format, e.g `claims_map.usename=email`

new OIDC config
```yaml
oidc:
  only_start_if_oidc_is_available: true
  issuer: "https://auth.example.com/auth/realms/master"
  client_id: "YOUR_CLIENT_ID"
  client_secret: "YOUR_SECRET"
  #   # Alternatively, set `client_secret_path` to read the secret from the file.
  #   # It resolves environment variables, making integration to systemd's
  #   # `LoadCredential` straightforward:
  #   client_secret_path: "${CREDENTIALS_DIRECTORY}/oidc_client_secret"
  #   # client_secret and client_secret_path are mutually exclusive.
  #
  #   # Customize the scopes used in the OIDC flow, defaults to "openid", "profile" and "email" and add custom query
  #   # parameters to the Authorize Endpoint request. Scopes default to "openid", "profile" and "email".
  # scope: ["openid", "profile", "email"]

  expiry:
    #
    #   # Use the expiry from the token received from OpenID when the user logged
    #   # in, this will typically lead to frequent need to reauthenticate and should
    #   # only been enabled if you know what you are doing.
    #   # Note: enabling this will cause `oidc.expiry.fixed_time` to be ignored.
    from_token: false
    #
    #   # The amount of time from a node is authenticated with OpenID until it
    #   # expires and needs to reauthenticate.
    #   # Setting the value to "0" will mean no expiry.
    fixed_time: 180d

  #   extra_params:
  #     domain_hint: example.com

  # allowd:
  #   domains:
  #     # List allowed principal domains and/or users. If an authenticated user's domain is not in this list, the
  #     # authentication request will be rejected.
  #     - example.com
  #   groups:
  #     # List allowed groups. 
  #     - admins
  #   users:
  #     - admin@example.com

  #  Map claims from the OIDC token to the user object
  claims_map:
    name: name
    username: preferred_username
    email: email
    groups: groups
    

  #  some random configuration
  misc:
    # if the username is set to `email` then `strip_email_domain` is valid
    # If `strip_email_domain` is set to `true`, the domain part of the username email address will be removed.
    # This will transform `first-name.last-name@example.com` to the user `first-name.last-name`
    # If `strip_email_domain` is set to `false` the domain part will NOT be removed resulting to the following
    # user: `first-name.last-name.example.com`
    strip_email_domain: true
    # If `flatten_groups` is set to `true`, the groups claim will be flattened to a single level.
    # this is used for keycloak where the groups are nested. the groups format from keycloak is `group1/subgroup1/subgroup2`
    flatten_groups: true
    # If `flatten_splitter` is set to a string, the groups claim will be split by the string and flattened to a single level.
    flatten_splitter: "/"
```

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


IamTaoChen:
I notice there is PR which supports Google, maybe, we can add a key named `provider = keycloak | google` to distinguish different IdP

config-example.yaml
```diff
@@ -284,44 +284,60 @@ unix_socket_permission: "0770"
...
+#     from_token: false
+#     #
+#     # The amount of time from a node is authenticated with OpenID until it
+#     # expires and needs to reauthenticate.
+#     # Setting the value to "0" will mean no expiry.
+#     fixed_time: 180d
+
+#   # # List allowed principal domains and/or users. If an authenticated user's domain is not in this list, the
+#   # # authentication request will be rejected.
+#   # allowd:
```
ChibangLW, line 311: Just a small typo.

IamTaoChen: I corrected it

Issue #1205, opened 2023-02-06
Expose more machine info via api


TymanWasTaken:
**Feature request**

Currently, the headscale http api allows for programs to list and get info on machines connected to the headscale network, such as IP, name, last seen date, and routes. However, it is still missing a lot of information that would be nice to have exposed, such as tailscale SSH support, tailscale client version, or network info (such as DERP latency and connection info such as hairpinning or UPnP/PCP/PMP support). 

This information would be beneficial to be exposed in order to both allow for better display of devices from for example a web browser or another API client. My specific use case was that I recently setup [home assistant](https://home-assistant.io), and noticed that the builtin tailscale integration did not support headscale. I would like to make a headscale integration, however many of the features of the builtin tailscale integration are not possible due to the lack of exposed information in the headscale API.

Additionally, this information would likely be easily exposed, as a quick inspection of the headscale database shows that this information is already stored in the `host_info` column of the `machines` table. For example, ssh support can be detected from the `sshHostKeys` field, and all of the network info such as UPnP and DERP info is seen in the `NetInfo` field.

juanfont:
Hi @TymanWasTaken, @joshuataylor

There has been a PR on this by @joshuataylor https://github.com/juanfont/headscale/pull/942.

After discussing it with @kradalby, we reckon it would be safer/saner for Headscale if you just marshal the Machine.HostInfo as a string into gRPC, rather than transforming it into a proto object in Headscale. This way, if something changes (and it might from Tailscale side) Headscale would not be affected.

PR #1938, opened 2024-05-14
Added HostInfo output to Node response

hopleus:
- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

# Usecase
As an administrator, I want to see the information that is available only in HostInfo, for example, the version of the client's TailScale, the OS name, its version, etc.


hopleus:
@juanfont, I saw your comment https://github.com/juanfont/headscale/issues/1205#issuecomment-1542204670 but I think that adding the base fields from HostInfo will not affect the HeadScale in any way, because it is unlikely that TailScale will change the fields used in any way

kradalby:
Quoting the issue referred above:
> After discussing it with @kradalby, we reckon it would be safer/saner for Headscale if you just marshal the Machine.HostInfo as a string into gRPC, rather than transforming it into a proto object in Headscale. This way, if something changes (and it might from Tailscale side) Headscale would not be affected.

We cannot maintain all these field hand typed into the API, however just marshalling it as JSON and then unmarshalling it on the other side using `tailcfg.Hostinfo` would be fine. 

If this is implemented, it should also be part of the JSON output of `headscale nodes list --json` and it should have integration tests to verify the output. 

As mentioned in our contribution guidelines, all contributions needs tests and there are a few more tickboxes that has not been addressed.

hopleus:
> Quoting the issue referred above:
> 
> > After discussing it with @kradalby, we reckon it would be safer/saner for Headscale if you just marshal the Machine.HostInfo as a string into gRPC, rather than transforming it into a proto object in Headscale. This way, if something changes (and it might from Tailscale side) Headscale would not be affected.
> 
> We cannot maintain all these field hand typed into the API, however just marshalling it as JSON and then unmarshalling it on the other side using `tailcfg.Hostinfo` would be fine.
> 
> If this is implemented, it should also be part of the JSON output of `headscale nodes list --json` and it should have integration tests to verify the output.
> 
> As mentioned in our contribution guidelines, all contributions needs tests and there are a few more tickboxes that has not been addressed.

Replaced with marshaling in JSON (protobuf.Struct).
Now, in the api, as well as through the cli, what is stored in the database will be output, without any processing

hscontrol/types/node.go
```diff
@@ -190,6 +192,33 @@ func (node *Node) IPsAsString() []string {
 	return ret
 }
 
+func (node *Node) HostInfoAsProtoStruct() *structpb.Struct {
+	if node.HostinfoDatabaseField == "" {
+		return nil
+	}
+
+	var mapper map[string]interface{}
+	if err := json.Unmarshal([]byte(node.HostinfoDatabaseField), &mapper); err != nil {
```
kradalby, line 201: You should not rely on this field as it is only set when we save, and not when things change. `.Hostinfo` must be used, can you pass that straight to `structpb.NewStruct()`?

integration/cli_test.go
```diff
@@ -999,6 +999,9 @@ func TestNodeCommand(t *testing.T) {
 	assert.Equal(t, "node-4", listAll[3].GetName())
 	assert.Equal(t, "node-5", listAll[4].GetName())
 
+	assert.Equal(t, "TestOS", listAll[0].GetHostInfo().AsMap()["OS"])
```
kradalby, line 1002: we should unmarshall it into the appropriate `tailcfg.HostInfo` type.

proto/headscale/v1/node.proto
```diff
@@ -35,11 +36,12 @@ message Node {
 
     RegisterMethod register_method = 13;
 
-    reserved 14 to 17;
+    reserved 14;
     // google.protobuf.Timestamp updated_at = 14;
-    // google.protobuf.Timestamp deleted_at = 15;
 
-    // bytes host_info      = 15;
+    google.protobuf.Struct host_info = 15;
```
kradalby, line 42: I do not think you can reuse 15, you need to use 18, but I am not sure.

kradalby (changes requested):

hscontrol/types/node.go
```diff
@@ -192,31 +191,21 @@ func (node *Node) IPsAsString() []string {
 	return ret
 }
 
-func (node *Node) HostInfoAsProtoStruct() *structpb.Struct {
-	if node.HostinfoDatabaseField == "" {
-		return nil
+func (node *Node) HostInfoAsString() string {
```
kradalby, line 194: maybe this could just be `[]byte` ? not sure if gRPC/protobuf covers that?

hscontrol/types/node.go
```diff
@@ -397,12 +386,15 @@ func (node *Node) Proto() *v1.Node {
...
 		User:        node.User.Proto(),
 		ForcedTags:  node.ForcedTags,
-		HostInfo:    node.HostInfoAsProtoStruct(),
 
 		// TODO(kradalby): Implement register method enum converter
 		// RegisterMethod: ,
 
 		CreatedAt: timestamppb.New(node.CreatedAt),
+
+		TailCfg: &v1.TailCfg{
```
kradalby, line 395: This intermediate type seem unnecessary, just put HostInfo straight in the Node object.

integration/cli_test.go
```diff
@@ -999,8 +1000,14 @@ func TestNodeCommand(t *testing.T) {
 	assert.Equal(t, "node-4", listAll[3].GetName())
 	assert.Equal(t, "node-5", listAll[4].GetName())
 
-	assert.Equal(t, "TestOS", listAll[0].GetHostInfo().AsMap()["OS"])
-	assert.Equal(t, "DebugTestNode", listAll[0].GetHostInfo().AsMap()["Hostname"])
+	assert.NotNil(t, listAll[0].TailCfg)
+
+	var hi *tailcfg.Hostinfo
+	err = json.Unmarshal([]byte(listAll[0].TailCfg.HostInfo), &hi)
```
kradalby, line 1006: By this point we should make sure its already unmarshalled and part of the JSON that is presented to the user, not an escaped string.

hscontrol/types/node.go
hopleus: > maybe this could just be `[]byte` ? not sure if gRPC/protobuf covers that?

[]byte can be returned, but the output will be a base64 string. This will lead to the fact that on the client side you will first have to convert base64 to string, and only then parse it

integration/cli_test.go
hopleus: > By this point we should make sure its already unmarshalled and part of the JSON that is presented to the user, not an escaped string.

The user will be provided with tailcfg.HostInfo serialized into a string that will be passed to gRPC, resulting in JSON inside JSON. And I understand that this is exactly what you assumed, and if so, then I don't quite understand what your phrase means. The user himself will have to parse the JSON from the string.
> its already unmarshalled and part of the JSON that is presented to the user

tale:
@kradalby as someone who works on a web interface for Headscale I'm interested in seeing this PR merged. I'd be willing to also take it on and move it over the finish line if the original author is no longer available 😄

kradalby:
@tale happy for you to give it a go.

I am not really sure what I want done atm, I really dont like having it as a string and then having the user having to unmarshal it after, it makes it all ver unergonomic. 

I dont have a good solution, but I am more conservative to adding it than adding it for the sake of it. If you want to have a look and come up with a proposal I am willing to hear it.

ArcticLampyrid:
> I dont have a good solution, but I am more conservative to adding it than adding it for the sake of it. If you want to have a look and come up with a proposal I am willing to hear it.

I respectfully disagree with this perspective. Regardless of its limitations, a unified mechanism to check the client version is essential as it helps mitigate security vulnerabilities. While avoiding action may reduce the risk of introducing errors, it unfortunately does not address the underlying issues.


> We cannot maintain all these field hand typed into the API

May I ask why this is a concern? These fields are also utilized by Tailscale’s console, which is almost certainly stable. Following this approach would not introduce more compatibility issues than those present in Tailscale’s official instance.

kradalby:
> I respectfully disagree with this perspective. Regardless of its limitations, a unified mechanism to check the client version is essential as it helps mitigate security vulnerabilities. While avoiding action may reduce the risk of introducing errors, it unfortunately does not address the underlying issues.

I disagree with this perspective. If the version field specifically is needed, then I propose that being added.

> May I ask why this is a concern? These fields are also utilized by Tailscale’s console, which is almost certainly stable. Following this approach would not introduce more compatibility issues than those present in Tailscale’s official instance.

The alternative is this: https://github.com/juanfont/headscale/pull/942/files#diff-ac94c91f8ea84762d55b6932ef1466a2165984591d3e646360c5bb31b2130900R248

The fields do change, manually managing drift is not an option.

ArcticLampyrid:
> The fields do change, manually managing drift is not an option.

I briefly checked HostInfo (not very thoroughly), and it seems there have been no breaking changes in recent years (I mean deleting or renaming a field). It does change, but only with the addition of new fields. Following general principles, writing a manual transformation function for this doesn’t seem to add much overhead. As for the fields that might be added in the future, we can simply ignore them. If they become useful to us, a random user who needs them will likely file an issue, and we can address it then. I don’t see anything wrong with that. Users who need to use potential new features will have to wait for a new version to adapt, which is very common.

> I disagree with this perspective. If the version field specifically is needed, then I propose that being added.

This is just my personal requirement, and I'm not sure how many others share the same need. However, recently, Headscale has been frequently removing support for backward compatibility with older client versions (which makes sense for maintainability), and I believe we need a convenient way to determine which clients will be broken by updating Headscale (in my case, these are usually clients running on OpenWRT), so we can take preventive measures before updating Headscale.

> I am more conservative to adding it than adding it for the sake of it.

I still don't think being conservative is the best approach. Perhaps it would reduce the maintenance cost of Headscale itself, but it may not be the case for the entire ecosystem. Headplane (one of the popular control panels) is actually considering running a Tailscale client and then using "localapi" to fetch information (tale/headplane#65), but in my view, this would add enough burden to deployment and usage.

ArcticLampyrid:
Considering that I’ve discovered some new clues, I will request this feature once again and propose providing information equivalent to the following JSON:

```json
{
  "addresses": [
    "100.87.74.78",
    "fd7a:115c:a1e0:ac82:4843:ca90:697d:c36e"
  ],
  "id": "92960230385",
  "nodeId": "n292kg92CNTRL",
  "user": "amelie@example.com",
  "name": "pangolin.tailfe8c.ts.net",
  "hostname": "pangolin",
  "clientVersion": "v1.36.0",
  "updateAvailable": false,
  "os": "linux",
  "created": "2022-12-01T05:23:30Z",
  "connectedToControl": true,
  "lastSeen": "2022-12-01T05:23:30Z",
  "keyExpiryDisabled": false,
  "expires": "2023-05-30T04:44:05Z",
  "authorized": false,
  "isExternal": false,
  "multipleConnections": true,
  "machineKey": "",
  "nodeKey": "nodekey:01234567890abcdef",
  "blocksIncomingConnections": false,
  "enabledRoutes": [
    "10.0.0.0/16",
    "192.168.1.0/24"
  ],
  "advertisedRoutes": [
    "10.0.0.0/16",
    "192.168.1.0/24"
  ],
  "clientConnectivity": {
    "endpoints": [
      "199.9.14.201:59128",
      "192.68.0.21:59128"
    ],
    "latency": {
      "Dallas": {
        "latencyMs": 60.463043
      },
      "New York City": {
        "preferred": true,
        "latencyMs": 31.323811
      }
    },
    "mappingVariesByDestIP": false,
    "clientSupports": {
      "hairPinning": false,
      "ipv6": false,
      "pcp": false,
      "pmp": false,
      "udp": false,
      "upnp": false
    }
  },
  "tags": [
    "tag:golink"
  ],
  "tailnetLockError": "",
  "tailnetLockKey": "",
  "sshEnabled": false,
  "postureIdentity": {
    "serialNumbers": [
      "CP74LFQJXM"
    ]
  },
  "isEphemeral": false
}
```

These fields are provided by the official Tailscale API, and it is almost certain (by design) that they are stable. It would be best for us to offer capabilities similar to those of the official API.

See also:
[Tailscale API: Get a device​](https://tailscale.com/api#tag/devices/get/device/{deviceId})

kradalby:
With the idea of throwing out gRPC, this should be much easier to achieve, which will likely just implement the same api, see https://github.com/juanfont/headscale/issues/2305

PR #1949, opened 2024-05-17
add download page

IamTaoChen:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Add a download page to indicate where can download Tailscale and a link to know how to use it.
image

### DIFF
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -474,6 +474,7 @@ func (h *Headscale) createRouter(grpcMux *grpcRuntime.ServeMux) *mux.Router {
 	router.HandleFunc("/oidc/register/{mkey}", h.RegisterOIDC).Methods(http.MethodGet)
 	router.HandleFunc("/oidc/callback", h.OIDCCallback).Methods(http.MethodGet)
 	router.HandleFunc("/apple", h.AppleConfigMessage).Methods(http.MethodGet)
+	router.HandleFunc("/download", h.DownloadMessage).Methods(http.MethodGet)
 	router.HandleFunc("/apple/{platform}", h.ApplePlatformConfig).
 		Methods(http.MethodGet)
 	router.HandleFunc("/windows", h.WindowsConfigMessage).Methods(http.MethodGet)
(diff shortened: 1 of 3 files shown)


ohdearaugustin:
I think this is a feature. 

First of all it is not 100% clear that this will download the client and not headscale itself. 

As the clients are not maintained by this project, we would add an external dependency we have to track. Furthermore, I don't see the value of adding this extra download page. Maybe you can clarify the use case of it, as I don't see one.

Furthermore no documentation and unittests for the endpoint.

IamTaoChen:
> I think this is a feature.
> 
> First of all it is not 100% clear that this will download the client and not headscale itself.
> 
> As the clients are not maintained by this project, we would add an external dependency we have to track. Furthermore, I don't see the value of adding this extra download page. Maybe you can clarify the use case of it, as I don't see one.
> 
> Furthermore no documentation and unittests for the endpoint.

This page just want to tell my group members how to use headsacle. before, I need to ask them which OS they are using and give them the corresponding link. And now,  I only need send this one page. Here is a demo [https://vpn.demo8.org/download](https://vpn.demo8.org/download). I would say it copy from [Tailscale](https://tailscale.com/download), I ask chatgpt to write this html.

IamTaoChen:
I also have another solution. change the the file name to `download.html` and put it at `/var/www/headscale` the add the 
```conf
    RewriteEngine On
    RewriteCond %{REQUEST_URI} ^/download$
    RewriteRule ^/download$ /var/www/headscale/download.html [L]
    Alias /download /var/www/headscale/download.html
```
to apache2 config file.


[download.html.txt](https://github.com/juanfont/headscale/files/15369819/download.html.txt)

kradalby:
> As the clients are not maintained by this project, we would add an external dependency we have to track. Furthermore, I don't see the value of adding this extra download page. 

I agree with this, it is not great linking to external sources and the URL might change which means that people with old headscales will have download pages with broken links. I think the benefit of this vs going to pkgs.tailscale.com or tailscale.com/download is not outweighing the potential maintenance.

PR #1957, opened 2024-05-25
feat: support client verify for derp

### LINKED ISSUE #1953: [Feature] Support for derp's verify-client-url
### Use case

When I deploy derp myself and don't want it to be used by other unauthorized clients, the traditional approach is to have derp access tailscaled to verify that the clientKey is in the list via derp's `verify-clients` parameter.

But I don't want to deploy tailscale on derp's nodes, and derp provides the `verify-client-url` parameter to determine if the clientKey is in the list via HTTP. I want Headscale to support this HTTP interface, so I can set derp's `verify-client-url` to the Headscale interface.

### Description

See <https://github.com/tailscale/tailscale/blob/964282d34f06ecc06ce644769c66b0b31d118340/derp/derp_server.go#L1159>.

Derp sent a POST request to `verifyClientsURL` with the following JSON

```json
{
  "NodePublic": "clientKey",
  "Source": "clientIP"
}
```

The expected return is

```json.
{
    "Allow": true
}
```

### Contribution

In Headscale, it could be to provide an HTTP interface that receives an authentication request, checks if the clientKey is in the list of nodes, and returns Allow.

117503445:
- [ ] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

I have initially implemented the verification function. Derp can verify whether the client is in the node list of headscale by specifying the `--verify-client-url` parameter to `/verify` in headscale.

I know there is still a lot of work to be done, such as comments, testing, documentation, etc. I am also willing to participate in the subsequent work.

Fixes #1953

### DIFF
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -484,6 +484,8 @@ func (h *Headscale) createRouter(grpcMux *grpcRuntime.ServeMux) *mux.Router {
 	router.HandleFunc("/swagger/v1/openapiv2.json", headscale.SwaggerAPIv1).
 		Methods(http.MethodGet)
 
+	router.HandleFunc("/verify", h.VerifyHandler).Methods(http.MethodPost)
+
 	if h.cfg.DERP.ServerEnabled {
 		router.HandleFunc("/derp", h.DERPServer.DERPHandler)
 		router.HandleFunc("/derp/probe", derpServer.DERPProbeHandler)
--- a/hscontrol/handlers.go
+++ b/hscontrol/handlers.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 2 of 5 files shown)


kradalby:
Nothing looks outrageous, but we need a way to test this, is there a way you can run a `tailscale debug` command or something and verify that it works or not? that should make it into a integration test.

hscontrol/handlers.go
```diff
@@ -53,6 +54,63 @@ func parseCabailityVersion(req *http.Request) (tailcfg.CapabilityVersion, error)
...
+	if req.Method != http.MethodPost {
+		http.Error(writer, "Wrong method", http.StatusMethodNotAllowed)
+		return
+	}
+	log.Debug().
+		Str("handler", "/verify").
+		Msg("verify client")
+
+	body, _ := io.ReadAll(req.Body)
+	derpAdmitClientRequest := tailcfg.DERPAdmitClientRequest{}
```
kradalby, line 71:
```suggestion
	var derpAdmitClientRequest tailcfg.DERPAdmitClientRequest
```

hscontrol/app.go
```diff
@@ -485,6 +485,8 @@ func (h *Headscale) createRouter(grpcMux *grpcRuntime.ServeMux) *mux.Router {
 	router.HandleFunc("/swagger/v1/openapiv2.json", headscale.SwaggerAPIv1).
 		Methods(http.MethodGet)
 
+	router.HandleFunc("/verify", h.VerifyHandler).Methods(http.MethodPost)
```
kradalby, line 488: maybe this should be under the derp router further down since it is only relevant if we run a derp?

117503445: No, helping other derps verify and running headscale build-in derp are two different things. Even without activating the built-in derp, headscale can still help other derps verify whether the client is in the nodelist.

In my own use case, my cloud server is more stable but has less bandwidth, while my homelab has more bandwidth but is often unavailable. Therefore, I install headscale on the cloud server but do not enable the built-in derp, and install derp in my homelab. At this point, I want the homelab's derp to access the cloud server's headscale via a `verify-url` to avoid unauthorized access to the derp.

117503445:
I encountered an unresolved issue.

In my testing, I created a DerpClient and DerpServer, setting the VerifyClientURL of the DerpServer to the Verify interface of Headscale. When a DerpClient created with a new privateKey attempts to connect to the DerpServer, the connection is rejected, which is the expected behavior.

However, I am unable to obtain the privateKey of the connected node, so I cannot create a legitimate DerpClient, resulting in a failure to pass the verification.

kradalby:
Does that mean there is an issue in derp, tailscale client or headscale?

117503445:
Now that I've solved the problem of testing, PR Ready

badsmoke:
any news here? it would be fantastic if this feature made it into the next version

ArcticLampyrid:
I found this feature useful to me. Is there something I can help to push the process?

117503445:
> I found this feature useful to me. Is there something I can help to push the process?

I compiled and used it on my own branch and had a good experience using it. I also hope this change will be merged.

kradalby:
Im sorry, we are completely overrun, I have not had time to look carefully but it looks like this is still missing an integration test for the client.

The test implemented checks the web endpoint from a http request, which is also good, but for us to merge this, it needs to be verified using the Tailscale client, not our code.

I don't know how that should look, maybe a tailscale client not joined to the Headscale client (and therefore not authorised to use the derp) that fails.

We are not able to take on the burden of untested code.

117503445:
merged by #2046

openapphub:
How to Use? It seems that the 0.24.0 document does not provide any information on this.

117503445:
> How to Use? It seems that the 0.24.0 document does not provide any information on this.

Assuming the headscale address is `https://headscale.mydomain.com`, you need to pass the `--verify-client-url` parameter to derp.

For example:

```sh
# remove --verify-clients flag
derper --verify-client-url https://headscale.mydomain.com/verify # you need to add other parameters as well
```

PR #1972, opened 2024-06-11
feat: HTML template to register a node

longregen:
I was trying to get a somewhat non-technical user to access my network. When using the tailscale app on iOS, we got presented with an HTML for registering the node. This was a little shocking in comparison with the rest of the experience.

So, I went on to https://v0.dev; asked it to craft a little QR code registration website, and exported it as a static HTML page. I connected it with https://github.com/davidshimjs/qrcodejs/. I then added some js code to add a copy button. I made sure it all works without javascript. I copied the "embed" approach from the OIDC routes.

This can be a controversial PR; but I think it adds some value. Now, due to how important security is in this project, I'm not sure you will be inclined at all to accept this approach; particularly I'm not happy that:

* This embeds too much CSS and makes the download pretty heavy (currently sitting at 93kb)
* It clutters the repository with generated code and makes checkouts heavier
* It's dubious whether our use case is also important for other headscale users

Here's how it looks like (on dark-mode):
Screenshot 2024-06-11 at 12-05-53 Headscale Registration

Without js:
Screenshot 2024-06-11 at 12-06-05 Headscale Registration

I would like to know your opinions, and I'm happy to contribute a solution with a different approach that satisfies you. One approach that I thought of involved allowing for the configuration of a templates folder; so that it can be modified at runtime and not compile time.

Looking forward to your comments!

## Summary by CodeRabbit

- **New Features**
	- Introduced a new registration interface for Headscale nodes with a responsive design.
	- Added functionality for dynamic QR code generation to simplify machine registration.
	- Enabled users to copy registration commands easily from the interface.

- **Improvements**
	- Enhanced maintainability of the registration web API by separating HTML content from source code.

### DIFF
--- a/hscontrol/assets/register_template.html
+++ b/hscontrol/assets/register_template.html
@@ -1,15 +1,2885 @@
-<html>
-  )
+<!doctype html>
+<html lang="en">
   <head>
-    <title>Registration - Headscale</title>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+
+    <style>
+      /*
+! tailwindcss v3.4.3 | MIT License | https://tailwindcss.com
+*/
+      *,
+      :after,
+      :before {
+        box-sizing: border-box;
+        border: 0 solid #e5e7eb;
+      }
+      :after,
+      :before {
+        --tw-content: "";
+      }
+      :host,
+      html {
+        line-height: 1.5;
+        -webkit-text-size-adjust: 100%;
+        -moz-tab-size: 4;
+        tab-size: 4;
+        font-family:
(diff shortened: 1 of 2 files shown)


kradalby:
> This was a little shocking in comparison with the rest of the experience

Considering Headscale no UI and CLI approach I am not sure if I would call it "shocking" :P

I am failing to see how having the command encoded into a QR code will help?

In any case, CSS is fine, but embedding tailwindcss is too much, both as an external dependency or embedding. 

I am sure someone (or an llm) can vipp up something that we can comfortably add to make it prettier if that is desired.

kradalby:
Closing due to no response

PR #1985, opened 2024-06-20
Make write-ahead-log default and configurable for sqlite

### LINKED ISSUE #1983: [Feature] Use SQLite in WAL mode by default
### Use case

SQLite is a very capable database that should easily be able to handle the use cases headscale aim to tackle. We support both SQLite and PostgreSQL and while we recommend SQLite, some users needed to switch to Postgres because 0.22.3 and older scaled poorly.

One of the reasons for this is that we do not run SQLite with the appropriate settings for making it scale better, like setting the journaling mode to write-ahead-log (WAL).

### Description

We should investigate, and make headscale use "the best" defaults for SQLite for 0.23.0 and onwards, making it the database of choice.

### Contribution

Review the current SQLite default settings, and set better defaults. 

We also need to investigate that using WAL is not breaking current setups, and there should be an opt out.

kradalby:
this commit makes headscale correctly enable write-ahead-log for sqlite and adds an option to turn it on and off.

WAL is enabled by default and should make sqlite perform a lot better, even further eliminating the need to use postgres.

It also adds a couple of other useful defaults.

Note, this PR removes WAL from the sqlite connection string, as that did not have any effect with the sqlite driver we use.

Fixes #1983

### DIFF
--- a/hscontrol/db/db.go
+++ b/hscontrol/db/db.go
@@ -434,13 +434,29 @@ func openDB(cfg types.DatabaseConfig) (*gorm.DB, error) {
 			Msg("Opening database")
 
 		db, err := gorm.Open(
-			sqlite.Open(cfg.Sqlite.Path+"?_synchronous=1&_journal_mode=WAL"),
+			sqlite.Open(cfg.Sqlite.Path),
 			&gorm.Config{
 				Logger: dbLogger,
 			},
 		)
 
-		db.Exec("PRAGMA foreign_keys=ON")
+		if err := db.Exec(`
+			PRAGMA foreign_keys=ON;
+			PRAGMA busy_timeout=10000;
+			PRAGMA auto_vacuum=INCREMENTAL;
+			PRAGMA synchronous=NORMAL;
+			`).Error; err != nil {
+			return nil, fmt.Errorf("enabling foreign keys: %w", err)
+		}
+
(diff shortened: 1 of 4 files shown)


evenh (approved):

hscontrol/db/db.go
```diff
@@ -434,13 +434,29 @@ func openDB(cfg types.DatabaseConfig) (*gorm.DB, error) {
...
+		if err := db.Exec(`
+			PRAGMA foreign_keys=ON;
+			PRAGMA busy_timeout=10000;
+			PRAGMA auto_vacuum=INCREMENTAL;
+			PRAGMA synchronous=NORMAL;
+			`).Error; err != nil {
+			return nil, fmt.Errorf("enabling foreign keys: %w", err)
+		}
+
+		if cfg.Sqlite.WriteAheadLog {
```
juanfont, line 452: This means that if someone reverts it to WriteAheadLog=false it will not be converted back right?

kradalby: Yes, default is journal mode delete which will bring it back

juanfont (approved):

hscontrol/db/db.go
zraexy: https://www.sqlite.org/wal.html#persistence_of_wal_mode

PR #1989, opened 2024-06-20
Remove allocations of lists before use

kradalby:
Remove the allocation of lists before they are used.


ohdearaugustin (approved):
Looks good

PR #2005, opened 2024-07-11
chore: configure some actions to be skipped for forks

vdovhanych:
This is just a simple change. 

This will configure some actions to be skipped for forks, as the actions that are changed here depend on the main repository or the configured secrets, so it makes no sense to run them for forks. This will also remove failed cron jobs from the forks, rest of the actions that don't depend on anything are left untouched.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -11,6 +11,9 @@ on:
     tags:
       # Stable release tags
       - v[0-9]+.[0-9]+.[0-9]+
+    paths:
+      - 'docs/**'
+      - 'mkdocs.yml'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/gh-actions-updater.yaml
+++ b/.github/workflows/gh-actions-updater.yaml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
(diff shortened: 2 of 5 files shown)


kradalby (approved):

kradalby:
Sorry I completely missed this, can you rebase?

vdovhanych:
I rebased it, but I'm not sure about the Docs deployment now. I see it's building it whenever you push to main, but I still feel that it is not really necessary for the forked repositories.
LMK, I can revert it there if you think it should still run on forks.

vdovhanych:
> > I rebased it, but I'm not sure about the Docs deployment now. I see it's building it whenever you push to main, but I still feel that it is not really necessary for the forked repositories. LMK, I can revert it there if you think it should still run on forks.
> 
> You're right. Docs should probably only get built/pushed when either `mkdocs.yml` or something in `docs/` changes.
> 
> But building and pushing should still work in a fork; I use this to render previews of proposed doc changes.

Okay, let's keep it then. I removed the if condition from the docs and added that we should only run the build on push if the paths relevant to the docs change.

kradalby:
Looks good to me, I'll let @nblock has the final say about the docs pipeline

nblock (approved):

PR #2018, opened 2024-07-18
actually lint file on CI

greizgh:
CI was producing this kind of messages:
> [config_reader] The output format `github-actions` is deprecated, please use `colored-line-number`

(see https://github.com/juanfont/headscale/actions/runs/9994119669/job/27623152138#step:6:206)

This also restore golangci-lint action which was not properly linting files due to the extra `.`  passed (we can expect it to fail if there are some non compliant files).

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: golangci-lint
         if: steps.changed-files.outputs.files == 'true'
-        run: nix develop --command -- golangci-lint run --new-from-rev=${{github.event.pull_request.base.sha}} --out-format=colored-line-number .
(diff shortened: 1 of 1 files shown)


kradalby:
hmm, I would have thought the `github-actions` to magically work nicely in github actions...

greizgh:
@kradalby if that's the disappearing `.`  which troubles you, this is because directories are NOT analyzed recursively (cf [quickstart](https://golangci-lint.run/welcome/quick-start/)). So we can either _not_ specify the directory (as I did in this PR) or explicitly pass `./...` for recursive check.
This single dot was the reason why the check never reported anything.

kradalby (approved):

PR #1997, opened 2024-06-27
adding oidc desired username as option in config

Lite5h4dow:
This PR allows people to use their desired username from their oidc provider instead of their email address. this is a user-facing feature and doesn't change any underlying processes

- [x] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [x] added unit tests
- [x] added integration tests
- [x] updated documentation if needed
- [x] updated CHANGELOG.md

PR #1594, opened 2023-11-08
Add OIDC claim names options

fen4o:
Some identity providers (auth0 for example) do not allow to set the groups claims (https://auth0.com/docs/secure/tokens/json-web-tokens/create-custom-claims) and administrators must use custom claims names and add them in the id token.

This commit adds the following configuration options:

- `oidc.groups_claim` to set the groups claim name
- `oidc.email_claim` to set the email claim name

All claims default to the previous values for backwards compatibility.

The groups claim can now also accept `[]string` or `string` as some providers might return only a string response instead of array.

- [x] read the [CONTRIBUTING guidelines](README.md#contributing)
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [x] added unit tests
- [ ] added integration tests
- [x] updated documentation if needed
- [ ] updated CHANGELOG.md

Fixes #1114

Issue #1531, opened 2023-08-21
Enhance OIDC Authentication with refresh_token Support


vsychov:
Hello,

First and foremost, I'd like to express my appreciation for the excellent product you have built.

Currently, `headscale` offers OIDC authorization, but its functionality appears to be somewhat limited. As it stands, it's challenging to provide both a smooth user experience and maintain security due to `headscale`'s inability to utilize `refresh_token`. This limitation presents us with a dilemma: either we grant the user an extended lifespan, during which their access might be revoked by IdAM, or we ask the user to re-authenticate periodically (e.g., daily).

A feasible solution would be to use both `access` and `refresh` tokens. Implementing this would primarily involve storing the `refresh_token` in the database and periodically updating the existing `access_token`. It's worth noting that the current OIDC implementation (from `coreos/go-oidc`) supports renewing the `access_token` via the `refresh_token`, as demonstrated [here](https://github.com/coreos/go-oidc/issues/199). Should an attempt to refresh the `access_token` fail, nodes authenticated by it should be logged out (set expire to `now`).

@juanfont , @kradalby I'm curious to hear your thoughts on this proposal. If this aligns with the project's direction, I would be more than happy to submit a pull request with the necessary modifications.

Thank you for your time and consideration.

almereyda:
PRs are set to wait until #1473 is resolved. There it was announced end of May that no contributions are taken for the time being.

mitchellkellett:
@kradalby did this ever get resolved? It looks like it went stale a few times. 

I’m having a similar issue where it’s not using the refresh token. Not sure if it should be to start with.

baznikin:
Please, implement `refresh_token` support

kradalby:
I've tried to read up on this and understand the usecase, and how the oauth/oidc part plays in and I am still a bit unsure about somethings, so I will lay it out how I understand it and I need someone to elaborate on the things I have misunderstood and usecases.

First I want to establish that from what I understand, Tailscale SaaS does not implement neither using the access token nor refresh tokens for expiry. They use their own key expiry time, which we also do as default.

In addition, headscale has a config option for making the expiry being set to the access token's expiry time.

About OIDC (or I suppose OAuth2 technically), access tokens are a part of OAuth2 to let **authorise** access to a service, typically from a program or another service. In headscale, access tokens are not used for anything.

[...]

PR #2020, opened 2024-07-19
Redo OIDC configuration

### LINKED ISSUE #938: Option to use Username instead of email when logging in via oidc

### LINKED ISSUE #1980: [Feature] OIDC DisplayName and ProfilePicURL support
### Use case

Mostly improves appearance, but probably also usability for users with multiple tailnets.

### Description

`DisplayName`, `ProfilePicURL` and `Email` (optional) are used and pushed to clients if user is logged in from OIDC

### Contribution

After OIDC login, if ProfilePicUrl and/or DisplayName is provided, it is set in DB in case it differs from saved state, then pushed to clients.

Currently `DisplayName` on macOS and Windows is only displayed in IPN settings, while `LoginName` (username@domain.tld) is used as primary identifier for Fast User Switching UI.

`LoginName` in `UserProfile` in `tailcfg` is expected to be an email in SaaS, which is not the case how it works in Headscale now, if username is stripped from email domain after login, it is just a username. However, if username email domain stripping is enabled, MagicDNS is not working correctly ('@' appears in FQDN), but I should probably create another issue for this.

I have implemented and tested experimental support for this (except `Email`) in my [fork](https://github.com/adipierro/headscale/tree/feat-oidc), but I don't have good programming skills, and probably might need help with design doc, as I'm sure my implementation is not ideal.

Draft PR is here https://github.com/juanfont/headscale/pull/1981.

| macOS | Windows |
|--------|--------|
| macOS | Windows |

### LINKED ISSUE #1990: [Feature] OIDC with permanent ID
### Use case

Currently, if user account in external system might have an email or username changed, OIDC authentication in Headscale won't match an existing user in DB, and another user will be created instead.

### Use OIDC `sub` claim as a permanent identifier for a user

If we use [`sub`](https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims) claim as a permanent unique ID for a user, we can match OIDC authenticated user with it instead of a username, and update a username (email) in DB if it differs. We should make updating optional as ACLs might stop applying to affected users.

### Use and save OIDC `email` claim regardless of email domain stripping

_A discussion is probably needed._
Screenshot 2024-06-22 at 5 21 29 PM
`email`, if available, could be used to display as `LoginName` in Tailscale clients. Or, it could be another way to identify users in ACLs if `strip_email_domain` is turned on, particularly, to avoid username collisions if multiple domains are allowed to login. 

But considering https://github.com/juanfont/headscale/pull/1987, we might not need to strip email domains anymore.

### Contribution

kradalby:
[thoughts](https://docs.google.com/document/d/1X85PMxIaVWDF6T_UPji3OeeUqVBcGj_uHRM5CI-AwlY)

Copy of Changelog:

- Remove `dns.use_username_in_magic_dns` configuration option [#2020](https://github.com/juanfont/headscale/pull/2020)
  - Having usernames in magic DNS is no longer possible.
- Redo OpenID Connect configuration [#2020](https://github.com/juanfont/headscale/pull/2020)
  - `strip_email_domain` has been removed, domain is _always_ part of the username for OIDC.
  - Users are now identified by `sub` claim in the ID token instead of username, allowing the username, name and email to be updated.
  - User has been extended to store username, display name, profile picture url and email.
    - These fields are forwarded to the client, and shows up nicely in the user switcher.
    - These fields can be made available via the API/CLI for non-OIDC users in the future.

Related issues:

Closes #1990 
Closes #1980
Closes #1981
Closes #1997
Closes #1594
Closes #938

These are closed as we will not support custom features outside of the OIDC standard.
Closes #1858 
Closes #1934

## Summary by CodeRabbit

- **New Features**
	- Enhanced user management with additional fields for display names, email addresses, and profile pictures.
	- Significant updates to OpenID Connect configuration, including the removal of `strip_email_domain` and a shift to using the `sub` claim for user identification.

- **Bug Fixes**
	- Improved error handling throughout authentication and user management processes.

- **Refactor**
	- Simplified authentication mechanisms by introducing an `AuthProvider` interface.
	- Streamlined user identification methods and removed deprecated configurations for clarity.
	- Adjusted DNS configuration handling by eliminating the use of usernames in MagicDNS.

- **Documentation**
	- Updated `CHANGELOG.md` to reflect changes in configuration and user identification mechanisms.


hscontrol/types/node_test.go
```diff
@@ -232,46 +160,16 @@ func TestNodeFQDN(t *testing.T) {
...
-				DNSConfig: &tailcfg.DNSConfig{
-					Proxied: false,
-				},
-				DNSUserNameInMagicDNS: false,
+				GivenName: "useruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruser11111111111111111111111111111111111111113444444444444444444444444444444444444444444444444444444444444444444444441111111111111111111111111111111111111111111111111111111111111111111111",
 			},
-			domain: "example.com",
-			want:   "test.example.com",
+			domain:  "example.com",
+			wantErr: "failed to create valid FQDN (useruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruseruser11111111111111111111111111111111111111113444444444444444444444444444444444444444444444444444444444444444444444441111111111111111111111111111111111111111111111111111111111111111111111.example.com): hostname too long, cannot except 255 ASCII chars",
```
kradalby, line 172: replace with strings.Repeat

mitchellkellett:
@kradalby would it be possible to also implement using a refresh token to allow for automatic node reauthentication as part of this oidc piece? Looks like it went stale in issue #1531

kradalby:
@mitchellkellett ill have a look.

IamTaoChen:
There could be an issue if we use the `sub` claim to identify the user. For example, if we switch the IdP from Keycloak to Authelia, the `sub` might change even though the user remains the same.

Another scenario is when I use Keycloak, but we switch the backend from OpenLDAP to AD. In this case, the `sub` will also change.

Therefore, I suggest we define the claim ourselves to avoid these potential issues.

kradalby:
Tagging a couple of people that have shown interest, might be helpful reviewing this:
@chriswiggins @vsychov @adipierro @ChibangLW @IamTaoChen

hscontrol/types/users.go
```diff
@@ -16,19 +18,57 @@ import (
 // that contain our machines.
 type User struct {
 	gorm.Model
+
+	// Username for the user, is used if email is empty
+	// Should not be used, please use Username().
 	Name string `gorm:"unique"`
```
juanfont, line 25: Same here, `Name`  is the user full name (given + family)... not necesarily unique.

https://openid.net/specs/openid-connect-basic-1_0-22.html#id_res

kradalby: This is the username, or, in OIDC terms: `preferred_username`, I suppose there is a risk that the upstream does not enforce uniqueness here, but also, we are kind of at the mercy of the non-oidc setup where we have had this unique all the time.

The Name you are referred to is going under DisplayName.

kradalby: https://github.com/juanfont/headscale/pull/2020/files#diff-7c0d50dc95e3b04d35f8bfbb30d4aee47101503fa703ea61747cb1a97c778d28R128 

was actually pointing at the wrong field, so fixed that.

juanfont (approved):

kradalby:
The failing test is breaking because of https://github.com/tailscale/tailscale/commit/1eaad7d3deb0815e8932e913ca1a862afa34db38, I will merge this and file a separate issue as it is only HEAD that is failing.

micolous:
> If you change your OIDC provider, a reasonable migration path is to write your own script mapping the new `sub` in place of the old ones based on the email etc. At the moment it could be done via the database, but I would expect it could be exposed via the API.

Yup. :)

The other issue here is if you key on things like `email` or `username`, if a provider allows users to change those attributes on their own without some sort of validation process (either verifying the email address, or a username uniqueness constraint), then you open up the service to account take-over issues).

Here is an example of a similar bug in Mastodon: https://github.com/mastodon/mastodon/security/advisories/GHSA-vm39-j3vx-pch3

It looks like *before* this PR, Headscale keyed on the username alone, so has a similar vulnerability (and IMHO, there should be an advisory about it).

It also looks like **the vulnerability is still there, even when an account is migrated to using `sub`,** because of fallback behaviour:

https://github.com/juanfont/headscale/blob/95150401612ec5143f5d9d44bcfa8730bcf94bba/hscontrol/oidc.go#L444-L456

There should be another check here to ensure the returned `user` does not have an `provider_identifier` already set.

Because otherwise, a *new* malicious user (whose `sub` is unknown to Headscale) who can change their preferred username on the IdP can use that to take over *any* account, even if they're now properly identified by `sub`.

PS: Yes, I'm aware I'm reporting a security bug publicly – but the bug is already in the public domain because of this PR... and it is not actually fixed. 😉

kradalby:
@micolous oh good catch, that didnt occur to me, I have created #2170 to address it with your proposal, could I get you to have a look at that for me?

jirutka:
ACLs still use the email address instead of the username as the user identifier. :/

kradalby:
That is the desired outcome.

jirutka:
Is it the desired outcome of this PR only or in general?

kradalby:
In general, we are moving towards being aligned with tailscale

Issue #2024, opened 2024-07-22
[Bug] ACL policy not working after update to v23.0 beta1 


masterwishx:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

SSH not working after updated to beta1 and changed config for it: 

![image](https://github.com/user-attachments/assets/32290ee2-2e53-4861-b5a6-6cf5348bbb04)

![image](https://github.com/user-attachments/assets/64eddc05-fa6f-4481-b387-6838780a65fe)

### Expected Behavior

wokring in versions befor

### Steps To Reproduce

...

### Environment

```markdown
- OS: Ubuntu
- Headscale version: 23.0 beta1
- Tailscale version:
```

### Runtime environment

- [X] Headscale is behind a (reverse) proxy
- [X] Headscale runs in a container

masterwishx:
adding text:  
```
{
  "groups": {
    "group:admin": ["masterwishx"],
    "group:family": ["user1", "user2", "user3"]
  },

  "tagOwners": {
    "tag:cloud-server": ["group:admin"],
    "tag:home-pc": ["group:admin", "group:family"],
    "tag:home-pc-vm": ["group:admin"],
    "tag:home-server": ["group:admin"],
    "tag:home-server-vm": ["group:admin"],
    "tag:home-mobile": ["group:admin", "group:family"],
    "tag:home-mobile-vm": ["group:admin", "group:family"]
  },

  "acls": [
    {
      // admin have access to all servers
      "action": "accept",
      "src": ["group:admin"],
      "dst": ["*:*"]
    },

    {
      // family have access to all home pcs,Speedtest Tracker
      "action": "accept",
      "src": ["group:family"],
      "dst": ["tag:home-pc:*", "tag:home-server:9443", "tag:home-server:8180"]
    }

    // We still have to allow internal users communications since nothing guarantees that each user have
    // their own users.
    //{ "action": "accept", "src": ["admin"], "dst": ["admin:*"] },
    //{ "action": "accept", "src": ["family"], "dst": ["family:*"] }
  ],

  "ssh": [
    {
      "action": "accept",
      //"src": ["tag:cloud-server", "tag:home-server", "tag:home-pc"],
      "src": ["group:admin"],
      "dst": ["tag:cloud-server", "tag:home-server"],
      "users": ["root", "ubuntu", "abc"]
    }
  ]
}
```

kradalby:
[...]

Issue #1963, opened 2024-05-31
[Bug] dns_config.domains only works if override_local_dns is enabled


ajs124:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Setting `dns_config.domains` without setting `dns_config.override_local_dns = true` leads to this warning and `dns_config.domains` being not applied:
`Warning: dns_config.domains is set, but no nameservers are configured. Ignoring domains.` (https://github.com/juanfont/headscale/blob/main/hscontrol/types/config.go#L545)

This warning is quite misleading, because `nameservers` actually refers to `resolvers`, which is taken from the `nameservers` field, but only if `override_local_dns = true` (https://github.com/juanfont/headscale/blob/main/hscontrol/types/config.go#L503).

### Expected Behavior

If there are technical reasons, keep this behavior and emit a less misleading error message.

If there aren't any technical reasons for why `domains` needs `override_local_dns`, change the check and apply this config.

### Steps To Reproduce

Set `domains` to a non-empty list while setting `override_local_dns` to `false`.

### Environment

```markdown
- OS: NixOS 24.05
- Headscale version: master, e.g. 5f9c26930ce796284a8fc7ad167f76e187813d64
- Tailscale version: 1.66.4
```

### Runtime environment

- [X] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

### Anything else?

I'm open to implementing either option (better error message or changing the check) and opening a PR, if you want me to.

kradalby:
I think this is a longstanding bug, I dont think the error makes sense based on Tailscales docs, can you give https://github.com/juanfont/headscale/pull/1987 a go?

ajs124:
Thanks! I will give that PR a go, but it might take some days/weeks.

ajs124:
I gave it a try and it seems to solve my problem. Didn't do any extensive tests though.

PR #2034, opened 2024-07-26
Redo DNS configuration

### LINKED ISSUE #1936: [Feature] Restricted nameservers without search domains
### Use case

I use the `restricted_nameservers` option for my own implementation of App Connectors (the DNS server adds routes to the subnet router) so I can route services like 1Password over the tailnet. The issue with the current implementation is that 1password.com will also be added as a search domain, which I don't want.

### Description

It would be nice if it was possible to add a restricted nameserver without the domain being added as a searchdomain.

### Contribution

I've been digging in the Tailscale client sourcecode, and I think this can be implemented by adding the domain to `DNSConfig.MatchDomains` instead of `DNSConfig.SearchDomains`, but I'm not really sure about this.

### LINKED ISSUE #2025: [Bug] v0.23.0-beta1 breaks built-in DERP

### Current Behavior

```
$ tailscale status
100.64.0.1      pioneer              chris        linux   -
100.64.0.2      dj-magic-laserbrain  chris        linux   -
100.64.0.5      joeryzen             chris        linux   offline
100.64.0.4      meterpeter           chris        linux   -
100.64.0.6      scotty-the-fifth     chris        linux   idle, tx 4884 rx 0

# Health check:
#     - Tailscale could not connect to the 'Headscale Embedded DERP' relay server. Your Internet connection might be down, or the server might be temporarily unavailable.
#     - Tailscale could not connect to any relay server. Check your Internet connection.
```

### Expected Behavior

The builtin DERP keeps on working with the update, I have also configured and used this setup for a long time now.

### Steps To Reproduce

1. update `headscale` to version v0.23.0-beta1
2. observe that the builtin DERP stops working
3. revert back to v0.23.0-alpha12
4. observe that the DERP works again

I hope that I did not miss anything in the changelogs, but to me it looks like there was no config changes etc. required to keep this working between the two relevant versions.

### Environment

```markdown
- OS: Debian GNU/Linux trixie/sid 
- Headscale version: v0.23.0-beta1
- Tailscale version: 1.70.0
```

### Runtime environment

- [X] Headscale runs in a container

Although both of the above are the case the DERP server is just publicly accesible:
```yaml
ports:
  - 0.0.0.0:3478:3478
```

### Anything else?

The startup log claims that I do not have any DERP's configured:
```
headscale  | 2024-07-23T00:58:26Z INF Opening database database=sqlite3 path=/var/lib/headscale/db.sqlite
headscale  | 2024-07-23T00:58:26Z WRN DERP map is empty, not a single DERP map datasource was loaded correctly or contained a region
headscale  | 2024-07-23T00:58:26Z INF home/runner/work/headscale/headscale/hscontrol/derp/server/derp_server.go:103 > DERP region: {RegionID:999 RegionCode:christian-derp RegionName:Headscale Embedded DERP Latitude:0 Longitude:0 Avoid:false Nodes:[0xc00034b7a0]}
headscale  | 2024-07-23T00:58:26Z INF home/runner/work/headscale/headscale/hscontrol/derp/server/derp_server.go:104 > DERP Nodes[0]: &{Name:999 RegionID:999 HostName:vpn.heusel.eu CertName: IPv4: IPv6: STUNPort:3478 STUNOnly:false DERPPort:443 InsecureForTests:false STUNTestIP: CanPort80:false}
headscale  | 2024-07-23T00:58:26Z INF STUN server started at [::]:3478
headscale  | 2024-07-23T00:58:26Z INF Setting up a DERPMap update worker frequency=86400000
```

and yet this is my derp config (snippet), which used to work with the previous versions:

```yaml
# DERP is a relay system that Tailscale uses when a direct
# connection cannot be established.
# https://tailscale.com/blog/how-tailscale-works/#encrypted-tcp-relays-derp
#
# headscale needs a list of DERP servers that can be presented
# to the clients.
derp:
  server:
    # If enabled, runs the embedded DERP server and merges it into the rest of the DERP config
    # The Headscale server_url defined above MUST be using https, DERP requires TLS to be in place
    enabled: true

    # Region ID to use for the embedded DERP server.
    # The local DERP prevails if the region ID collides with other region ID coming from
    # the regular DERP config.
    region_id: 999

    # Region code and name are displayed in the Tailscale UI to identify a DERP region
    region_code: "christian-derp"

    # Listens over UDP at the configured address for STUN connections - to help with NAT traversal.
    # When the embedded DERP server is enabled stun_listen_addr MUST be defined.
    #
    # For more details on how this works, check this great article: https://tailscale.com/blog/how-tailscale-works/
    stun_listen_addr: "0.0.0.0:3478"

    private_key_path: /etc/headscale/derp_server_private.key

  # List of externally available DERP maps encoded in JSON
  # urls:
  #   - https://controlplane.tailscale.com/derpmap/default

  # Locally available DERP map files encoded in YAML
  #
  # This option is mostly interesting for people hosting
  # their own DERP servers:
  # https://tailscale.com/kb/1118/custom-derp-servers/
  #
  # paths:
  #   - /etc/headscale/derp-example.yaml
  paths: []

  # If enabled, a worker will be set up to periodically
  # refresh the given sources and update the derpmap
  # will be set up.
  auto_update_enabled: true

  # How often should we check for DERP updates?
  update_frequency: 24h
```

### LINKED ISSUE #2026: [Bug] 0.23.0-beta1 breaks DNS custom nameservers

### Current Behavior

After upgrading from 0.23.0-alpha12 to 0.23.0-beta1, with no other changes at all, queries made to `100.100.100.100` which should use a listed `nameserver` give `status: NXDOMAIN`. 

image

### Expected Behavior

Queries should return results from the `nameservers`

image

### Steps To Reproduce

I use the following DNS config:

```
dns_config:
  override_local_dns: true
  nameservers:
    - 100.64.0.6
  restricted_nameservers:
    foo.bar: # Redacted
      - 100.64.0.2
    baz.foo: # Redacted
      - 100.64.0.2
  domains: []
  magic_dns: true
  base_domain: lorem.ipsum # Redacted
```

On alpha9, 10, 11, and 12, it works fine. On beta1, it fails.

### Environment

```markdown
- OS: Unraid 6.12.10, running Docker 24.0.9
- Headscale version: 0.23.0-beta1
- Tailscale version: 1.70.0 on MacOS Sonoma
```

### Runtime environment

- [X] Headscale runs in a container

### Anything else?

_No response_

### LINKED ISSUE #2029: [Bug] 0.23.0-beta1 wipes resolv.conf on clients regardless of dns_config

### Current Behavior

Possibly related to #2026, I'm noticing that with 0.23.0-beta1, the /etc/resolv.conf on all clients is getting wiped out:

0.23.0-alpha14:
on one client with networkmanager
```
$ cat /etc/resolv.conf
# Generated by NetworkManager
nameserver 172.25.0.1
```
or on a different client with resolvconf
```
$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "resolvectl status" to see details about the actual nameservers.

nameserver 67.207.67.3
nameserver 67.207.67.2
```

0.23.0-beta1 on all clients:
```
$ cat /etc/resolv.conf 
# resolv.conf(5) file generated by tailscale
# For more info, see https://tailscale.com/s/resolvconf-overwrite
# DO NOT EDIT THIS FILE BY HAND -- CHANGES WILL BE OVERWRITTEN

search headscale.net
```

Originally seen with a headscale configuration where `dns_config` is completely omitted. With beta1 still installed, I tried specifying a `dns_config` with `override_local_dns: false`. I also tried adding `nameservers: ["172.25.0.1"]`. In all cases the issue persisted (`/etc/resolv.conf` continuing to be wiped)

I reverted back to 0.23.0-alpha14 and the clients immediately went back to their original `resolv.conf` configs

### Expected Behavior

see above

### Steps To Reproduce

Installing 0.23.0-beta1 immediately caused the issue (resolv.conf wiped). Reverting to 0.23.0-alpha14 immediately fixed the issue (resolv.conf restored).

### Environment

```markdown
- OS: Arch Linux and Debian Linux
- Headscale version: 0.23.0-beta1
- Tailscale version: 1.68.2
```

### Runtime environment

- [X] Headscale runs in a container

### Anything else?

_No response_

### LINKED ISSUE #2047: [Bug] Tailscale Exit Node DNS Doesnt Provide DNS

### Current Behavior

When connected to an exit node that is routing all traffic through said node, I cannot access the public internet as DNS does not work.

- I can run nslookups with 1.1.1.1 and google.com and they succeed
- I cannot ping 1.1.1.1 as it responds with a dns lookup error
- I can access local devices and internet-connected servers via their ip address

If I disable Tailscale DNS
- DNS works as expected, but is not under the control of headscale.

Note I am only able to test this with the exit node on windows and the client running android.

### Expected Behavior

Using the tailscale dns, my headscale server should route traffic to 1.1.1.1 as is in the config.

### Steps To Reproduce

1. Set up basic headscale server
2. set the dns servers in the config
3. Run an exit node
4. Connect a device to the exit node with tailscale dns enabled
5. DNS no worky.

### Environment

```markdown
- OS: Windows 10 & Android
- Headscale version: 0.23.0-beta
- Tailscale version: 1.70.0
```

### Runtime environment

- [X] Headscale runs in a container

### Anything else?

_No response_

kradalby:
This PR aims to redo (simplify and correct) DNS configuration to match what is possible with Tailscale. 

TLDR: currently we have several settings that are either incompatible together, dont do what they sound like, or some dont work the way we thought at the time we, or a contributor implemented them.

DNS has been one of these things that _mostly worked_ and had not been revisited until I got #1963 which I looked at, thought, I can easily knock this one out, and then I _messed up everything_. DNS config was very much a patchwork that had been changed over time and a lot of this work was from before we had integration tests. 
It is embarrassing that it made it into the beta, but I suppose its a very efficient way to prioritise work...

This screwup has caused me to revisit the DNS config and I discovered that:
- A lot of the settings is hard to understand what does
- Some of them does not make sense in a Tailscale context
- Some combination might not work at all.

**Notable changes for people coming here from the CHANGELOG**:

`dns_config` is now `dns`.

`override_local_dns`: has been removed, if magic_dns or the other dns options are used, Tailscale needs to take over the DNS configuration. Clients can opt out by passing `--accept-dns=false` (https://tailscale.com/kb/1235/resolv-conf)

`domains` has been renamed to `search_domains` to describe what it actually is.

`nameservers` has become a root key, where:
old `nameservers` is now `nameservers.global` and 
`restricted_nameservers` is now `nameservers.split` to describe the fact that it was configuring Split DNS.

These settings now reflect what Tailscale (https://login.tailscale.com/admin/dns) allows you to do, but note, we still have some additional features like `extra_records` which has been preserved under the same key ;).

TODO:
- [x] Cobra can't read ENV vars for new config, breaking integration tests
- [x] more unit tests
- [x] more integration tests
- [x] updated CHANGELOG.md

Reverts DNS search path part of #1987

Fixes #2026
Fixes #2047
Fixes #1936
Fixes #2025
Fixes #2029
Updates #2024 
Updates #1963


kradalby:
This PR is in progress, I think the DNS part is working as it should but I ran into some issues with Viper and loading config from Environment variables which prevented me from writing all the tests that was missing and is needed. 

If someone takes it for a spin, that would be great, I will continue next week.

config-example.yaml
```diff
@@ -224,54 +224,63 @@ policy:
...
-  #     - 1.1.1.1
-  #     - 8.8.8.8
-
-  # Search domains to inject.
-  domains: []
+    global:
+      - 1.1.1.1
+      - 1.0.0.1
+      - 2606:4700:4700::1111
+      - 2606:4700:4700::1001
```
nickbp, line 253: For this cloudflare example it looks like they have a DoH endpoint at: `https://cloudflare-dns.com/dns-query` (from https://developers.cloudflare.com/1.1.1.1/encryption/dns-over-https/make-api-requests/)

nickbp (commented):
I'm new here but the new structure looks good to me, just had nitpicks with the example config

config-example.yaml
kradalby: I would think that it would work, but I have not actually tested it, happy to include it if someone can test it, but that can be a follow up PR too.

Leseratte10:
I found this PR linked from bug 2029. After updating to beta1, tailscale started messing with my /etc/resolv.conf and broke internet connectivity under Linux. 

I was able to get around that by using `--accept-dns=false` but that requires manual config on every Linux machine. Will there be a setting in this PR that allows me to be like "don't mess with DNS at all and leave the resolv.conf alone" that doesn't require client-side settings, or do I need to add that switch to every single client?

kradalby:
> Will there be a setting in this PR that allows me to be like "don't mess with DNS at all and leave the resolv.conf alone" that doesn't require client-side settings, or do I need to add that switch to every single client?

You will be able to set a null/empty value to all settings under the `dns` key. If any of the settings is set that requires Tailscale to take control of DNS, the client will do that as it sees fit, and the way to opt out of that is `--accept-dns=false`. 

I would like to add that the breakage in beta1 was unfortunate, but when the fix is in, I would not expect it to fail in that way as you likely have had Tailscale take over your connection up until this point.

masterwishx:
It's a little confusing new setup, was:
`server_url: https://headscale.site.com `
`base_domain: headscale.site.com`
Now should be?

kradalby:
> It's a little confusing new setup, was:
> 
> `server_url: https://headscale.site.com `
> 
> `base_domain: headscale.site.com`
> 
> Now should be? 

They need to be different, so you could do clients.headscale.site.com for example

masterwishx:
If you still need to test it please tell me if this OK?
```
git clone --branch kradalby/redo-dns-config https://github.com/kradalby/headscale
cd headscale
docker build -f Dockerfile.debug  -t headscale/headscale:redo-dns-config
```

kradalby:
@masterwishx how did this work for you?

config-example.yaml
```diff
@@ -224,54 +224,67 @@ policy:
...
+#
+# If you want stop Headscale from managing the DNS configuration
+# all the fields under `dns` should be set to empty values.
+dns:
+  # Whether to use [MagicDNS](https://tailscale.com/kb/1081/magicdns/).
+  # Only works if there is at least a nameserver defined.
+  magic_dns: true
+
+  # Defines the base domain to create the hostnames for MagicDNS.
+  # This domain _must_ be different from the server_url domain.
```
SuperSandro2000, line 245: Different as in different subdomains or different TLDs?

Would headscale.example.com and scale.example.com work?

config-example.yaml
SuperSandro2000: Another question: by default resolv.conf only takes 3 name servers. How does tailscale handle more?

config-example.yaml
```diff
@@ -224,54 +224,67 @@ policy:
...
+      - 2606:4700:4700::1001
+
+      # NextDNS (see https://tailscale.com/kb/1218/nextdns/).
+      # "abc123" is example NextDNS ID, replace with yours.
+      # - https://dns.nextdns.io/abc123
+
+    # Split DNS (see https://tailscale.com/kb/1054/dns/),
+    # a map of domains and which DNS server to use for each.
+    split:
+      {}
```
SuperSandro2000, line 266:
```suggestion
    split: {}
```

hscontrol/types/config.go
```diff
@@ -201,7 +216,10 @@ func LoadConfig(path string, isFile bool) error {
 		}
 	}
 
-	viper.SetEnvPrefix("headscale")
+	// viper.ExperimentalBindStruct()
```
SuperSandro2000, line 218: Is this a left over from testing?

config-example.yaml
kradalby: different (sub)domains, your example works.

Could also do 
headscale.net as the server
clients.headscale.net as clients


Technically could do 
server.headscale.net as the server
headscale.net as clients (as long as you dont have a node called server)

hscontrol/types/config.go
```diff
@@ -485,123 +520,117 @@ func GetDatabaseConfig() DatabaseConfig {
...
+
+	return resolvers
+}
 
-		if viper.IsSet("dns_config.extra_records") {
-			var extraRecords []tailcfg.DNSRecord
+// SplitResolvers returns a map of domain to DNS resolvers and a list of search domains.
+// If a nameserver is a valid IP, it will be used as a regular resolver.
+// If a nameserver is a valid URL, it will be used as a DoH resolver.
+// If a nameserver is neither a valid URL nor a valid IP, it will be ignored.
```
SuperSandro2000, line 584: We should at least log a warning here to prevent hard to debug silently ignored errors.

hscontrol/types/config.go
```diff
@@ -805,19 +853,70 @@ func IsCLIConfigured() bool {
...
+
+// warnWithAlias will register an alias between the newKey and the oldKey,
 // and log a deprecation warning if the oldKey is set.
-func registerAliasAndDeprecate(newKey, oldKey string) {
+func (d *deprecator) warnWithAlias(newKey, oldKey string) {
 	// NOTE: RegisterAlias is called with NEW KEY -> OLD KEY
 	viper.RegisterAlias(newKey, oldKey)
 	if viper.IsSet(oldKey) {
-		log.Warn().Msgf("The %q configuration key is deprecated. Please use %q instead. %q will be removed in the future.", oldKey, newKey, oldKey)
+		d.warns = append(d.warns, fmt.Sprintf("The %q configuration key is deprecated. Please use %q instead. %q will be removed in the future.", oldKey, newKey, oldKey))
```
SuperSandro2000, line 862: The message should probably be deduped.

hscontrol/types/config_test.go
```diff
@@ -0,0 +1,272 @@
...
+			name: "unmarshal-dns-full-config",
+			configEnv: map[string]string{
+				"HEADSCALE_DNS_MAGIC_DNS":                 "true",
+				"HEADSCALE_DNS_BASE_DOMAIN":               "example.com",
+				"HEADSCALE_DNS_NAMESERVERS_GLOBAL":        `1.1.1.1 8.8.8.8`,
+				"HEADSCALE_DNS_SEARCH_DOMAINS":            "test.com bar.com",
+				"HEADSCALE_DNS_USE_USERNAME_IN_MAGIC_DNS": "true",
+
+				// TODO(kradalby): Figure out how to pass these as env vars
+				// "HEADSCALE_DNS_NAMESERVERS_SPLIT":  `{foo.bar.com: ["1.1.1.1"]}`,
```
SuperSandro2000, line 223: maybe something like?
```
HEADSCALE_DNS_NAMESERVERS_SPLIT=foo.bar.com=[1.1.1.1,1.0.0.1],buzz.bar.com=[8.8.8.8,8.4.4.8]
```

config-example.yaml
kradalby: resolv.conf is a "fallback" for tailscale, [it will try a number of ways before using resolv.conf](https://tailscale.com/blog/sisyphean-dns-client-linux).

In this case, when you override DNS, the DNS server in resolv.conf will be 100.100.100.100, which lives inside the tailscale daemon, and it will proxy everything.

kradalby: you can see this in the integration tests https://github.com/juanfont/headscale/pull/2034/files#diff-6534f1d1f61ff53e7fe7de90f7736900614802c5bdaa10a63f485316b95351c2R126

integration/dns_test.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+	}
+}
+
+// TestValidateResolvConf validates that the resolv.conf file
+// ends up as expected in our Tailscale containers.
+// All the containers are based on Alpine, meaning Tailscale
+// will overwrite the resolv.conf file.
+// On other platform, Tailscale will integrate with a dns manager
+// if available (like Systemd).
```
SuperSandro2000, line 89:
```suggestion
// if available (like Systemd-Resolved).
```
It is common to use systemd without resolved on eg NixOS

config-example.yaml
kradalby: Prettier prefers this, so this will stay tho it looks a bit weird.

kradalby:
Thanks @SuperSandro2000, think I addressed all comments.

masterwishx:
> @masterwishx how did this work for you?

Sorry not checked yet , i see you added more changes here ?

config-example.yaml
masterwishx: we dont need any changes for Nginx proxy Manger also ? 
if was : 
`server_url: https://headscale.site.com `
`base_domain: headscale.site.com`

and will be : 

`server_url: https://headscale.site.com `
`base_domain: clients.headscale.site.com`

kradalby: I dont think so, but I dont know anything about nginx proxy manager.

juanfont (approved):

config-example.yaml
Minion3665: This is a real shame - is there a technical reason why this can't be allowed, or is it a case of being too difficult/risky to support over the rewrite?

I guess what I'm really asking is: is there a possibility of me (or someone else) making a patch to change this down the line?

kradalby: No problem, It is one of these things that "we dont want to add this", but it is how upstream works, and we dont really have any choice, and I understand why it happens upstream too. We just initially implemented it wrongly.

hscontrol/types/config.go
```diff
@@ -711,8 +759,23 @@ func GetHeadscaleConfig() (*Config, error) {
...
+	// BaseDomain cannot be the same as the server URL.
+	// This is because Tailscale takes over the domain in BaseDomain,
+	// causing the headscale server and DERP to be unreachable.
+	// For Tailscale upstream, the following is true:
+	// - DERP run on their own domains
+	// - Control plane runs on login.tailscale.com/controlplane.tailscale.com
+	// - MagicDNS (BaseDomain) for users is on a *.ts.net domain per tailnet (e.g. tail-scale.ts.net)
+	//
+	// TODO(kradalby): remove dnsConfig.UserNameInMagicDNS check when removed.
+	if !dnsConfig.UserNameInMagicDNS && dnsConfig.BaseDomain != "" && strings.Contains(serverURL, dnsConfig.BaseDomain) {
```
mtoohey31, line 773: I could be wrong, but I wonder if the `string.Contains` check might be too strict of a test. I want to use (for example) `https://bar.foo.com` as my `server_url`, and just `foo` as my `dns.base_domain`, but this check rejects that. I've manually removed the check and this seems to work without issue.

motiejus:
Thanks for #2210 @quite!

Since the change got more restrictive, I also posted a question whether it's safe to change `base_domain` on an existing HS instance with connected clients: https://github.com/juanfont/headscale/issues/2210#issuecomment-2480130747

PR #2046, opened 2024-08-07
feat: support client verify for derp (add integration tests)

### LINKED ISSUE #1953: [Feature] Support for derp's verify-client-url

ArcticLampyrid:
* [x]  have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
* [x]  raised a GitHub issue or discussed it on the projects chat beforehand
* [ ]  added unit tests
* [x]  added integration tests
* [ ]  updated documentation if needed
* [ ]  updated CHANGELOG.md

Fixes #1953

By implementing `/verify` endpoint, derper can verify whether the client is in the node list of Headscale by specifying the `--verify-client-url` parameter to `/verify` in Headscale.

This is a continued work on PR #1957, which includes integration testing to meet Headscale's merge requirements.

I am willing to participate in updating the related documentation and change logs, if needed. (Of course, I believe @117503445 is also willing)

### DIFF
--- a/hscontrol/handlers.go
+++ b/hscontrol/handlers.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
@@ -54,6 +55,76 @@ func parseCabailityVersion(req *http.Request) (tailcfg.CapabilityVersion, error)
 	}
 
 	return tailcfg.CapabilityVersion(clientCapabilityVersion), nil
+}
+
+// see https://github.com/tailscale/tailscale/blob/964282d34f06ecc06ce644769c66b0b31d118340/derp/derp_server.go#L1159, Derp use verifyClientsURL to verify whether a client is allowed to connect to the DERP server.
+func (h *Headscale) VerifyHandler(
+	writer http.ResponseWriter,
+	req *http.Request,
+) {
+	if req.Method != http.MethodPost {
+		http.Error(writer, "Wrong method", http.StatusMethodNotAllowed)
+		return
+	}
+	log.Debug().
+		Str("handler", "/verify").
+		Msg("verify client")
+
+	body, err := io.ReadAll(req.Body)
+	if err != nil {
+		log.Error().
+			Str("handler", "/verify").
+			Err(err).
+			Msg("Cannot read request body")
+		http.Error(writer, "Internal error", http.StatusInternalServerError)
+		return
+	}
+
+	var derpAdmitClientRequest tailcfg.DERPAdmitClientRequest
+	if err := json.Unmarshal(body, &derpAdmitClientRequest); err != nil {
+		log.Error().
+			Caller().
+			Err(err).
+			Msg("Cannot parse derpAdmitClientRequest")
+		http.Error(writer, "Internal error", http.StatusInternalServerError)
+		return
+	}
+
+	nodes, err := h.db.ListNodes()
+	if err != nil {
+		log.Error().
+			Caller().
+			Err(err).
+			Msg("Cannot list nodes")
+		http.Error(writer, "Internal error", http.StatusInternalServerError)
+	}
+
+	for _, node := range nodes {
+		log.Debug().Str("node", node.NodeKey.String()).Msg("Node")
+	}
+
+	allow := false
+	// Check if the node is in the list of nodes
+	for _, node := range nodes {
+		if node.NodeKey == derpAdmitClientRequest.NodePublic {
+			allow = true
+			break
+		}
+	}
+
+	resp := tailcfg.DERPAdmitClientResponse{
+		Allow: allow,
+	}
+
+	writer.Header().Set("Content-Type", "application/json")
+	writer.WriteHeader(http.StatusOK)
+	err = json.NewEncoder(writer).Encode(resp)
+	if err != nil {
+		log.Error().
+			Caller().
+			Err(err).
+			Msg("Failed to write response")
+	}
 }
 
 // KeyHandler provides the Headscale pub key
--- a/hscontrol/handlers.go
+++ b/hscontrol/handlers.go
@@ -99,18 +99,7 @@ func (h *Headscale) VerifyHandler(
 		http.Error(writer, "Internal error", http.StatusInternalServerError)
 	}
 
-	for _, node := range nodes {
-		log.Debug().Str("node", node.NodeKey.String()).Msg("Node")
-	}
-
-	allow := false
-	// Check if the node is in the list of nodes
-	for _, node := range nodes {
-		if node.NodeKey == derpAdmitClientRequest.NodePublic {
-			allow = true
-			break
-		}
-	}
+	allow := nodes.ContainsNodeKey(derpAdmitClientRequest.NodePublic)
 
 	resp := tailcfg.DERPAdmitClientResponse{
 		Allow: allow,
--- a/hscontrol/handlers.go
+++ b/hscontrol/handlers.go
@@ -64,6 +64,7 @@ func (h *Headscale) VerifyHandler(
 ) {
 	if req.Method != http.MethodPost {
 		http.Error(writer, "Wrong method", http.StatusMethodNotAllowed)
+
 		return
 	}
 	log.Debug().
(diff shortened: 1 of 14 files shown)


ArcticLampyrid:
Rebased onto main.

Can you give a review?

kradalby:
Im away traveling, but ill get back to it in a few weeks when im back.

yqs112358:
Very very useful future 👍 Eliminates the need to create Tailscale clients for every derper server I build.

hscontrol/handlers.go
```diff
@@ -56,6 +57,76 @@ func parseCabailityVersion(req *http.Request) (tailcfg.CapabilityVersion, error)
...
+	}
+
+	allow := false
+	// Check if the node is in the list of nodes
+	for _, node := range nodes {
+		if node.NodeKey == derpAdmitClientRequest.NodePublic {
+			allow = true
+			break
+		}
+	}
```
kradalby, line 113: This could be a function on `types.Nodes` like `ContainsNodeKey`

hscontrol/handlers.go
```diff
@@ -56,6 +57,76 @@ func parseCabailityVersion(req *http.Request) (tailcfg.CapabilityVersion, error)
...
+		log.Error().
+			Caller().
+			Err(err).
+			Msg("Cannot list nodes")
+		http.Error(writer, "Internal error", http.StatusInternalServerError)
+	}
+
+	for _, node := range nodes {
+		log.Debug().Str("node", node.NodeKey.String()).Msg("Node")
+	}
```
kradalby, line 104: This seem very verbose, left over from debugging?

hscontrol/handlers.go
```diff
@@ -56,6 +57,76 @@ func parseCabailityVersion(req *http.Request) (tailcfg.CapabilityVersion, error)
...
+
+	nodes, err := h.db.ListNodes()
+	if err != nil {
+		log.Error().
+			Caller().
+			Err(err).
+			Msg("Cannot list nodes")
+		http.Error(writer, "Internal error", http.StatusInternalServerError)
+	}
```
kradalby, line 100: Can all of this be a function so we can return a single error and not have to do the whole `log.Error` + `http.Error` so many times?

integration/derp_verify_endpoint_test.go
```diff
@@ -0,0 +1,111 @@
...
+	derpConfig += "    regionid: 900\n"
+	derpConfig += "    regioncode: test-derpverify\n"
+	derpConfig += "    regionname: TestDerpVerify\n"
+	derpConfig += "    nodes:\n"
+	derpConfig += "      - name: TestDerpVerify\n"
+	derpConfig += "        regionid: 900\n"
+	derpConfig += "        hostname: " + derper.GetHostname() + "\n"
+	derpConfig += "        stunport: " + derper.GetSTUNPort() + "\n"
+	derpConfig += "        stunonly: false\n"
+	derpConfig += "        derpport: " + derper.GetDERPPort() + "\n"
```
kradalby, line 58: Can you use `tailcfg.DERPMap` (https://github.com/tailscale/tailscale/blob/main/tailcfg/derpmap.go#L14) and then serialise it in the `hsic.WithCustomDERPServerOnly` method? JSON is valid YAML, so you can just use the JSON marshaller.

integration/derp_verify_endpoint_test.go
```diff
@@ -0,0 +1,111 @@
...
+	headscale, err := scenario.Headscale(
+		hsic.WithHostname(hostname),
+		hsic.WithPort(headscalePort),
+		hsic.WithCustomTLS(certHeadscale, keyHeadscale),
+		hsic.WithHostnameAsServerURL(),
+		hsic.WithCustomDERPServerOnly([]byte(derpConfig)),
+	)
+	assertNoErrHeadscaleEnv(t, err)
+
+	for userName, clientCount := range spec {
```
kradalby, line 69: Is there a reason why you cannot use `scenario.CreateHeadscaleEnv` here? you can pass tsic options with it.

integration/hsic/hsic.go
```diff
@@ -203,6 +216,27 @@ func WithEmbeddedDERPServerOnly() Option {
 	}
 }
 
+// WithCustomDERPServerOnly configures Headscale use a custom
+// DERP server only.
+func WithCustomDERPServerOnly(contents []byte) Option {
```
kradalby, line 221: Maybe this could be named `WithDERPConfig`, since it is more general dumping a file and using that?

integration/dsic/dsic.go
```diff
@@ -0,0 +1,318 @@
...
+
+	cmdArgs := "--hostname=" + hostname
+	cmdArgs += " --certmode=manual"
+	cmdArgs += " --certdir=" + DERPerCertRoot
+	cmdArgs += " --a=:" + strconv.Itoa(dsic.derpPort)
+	cmdArgs += " --stun=true"
+	cmdArgs += " --stun-port=" + strconv.Itoa(dsic.stunPort)
+	if dsic.withVerifyClientURL != "" {
+		cmdArgs += " --verify-client-url=" + dsic.withVerifyClientURL
+	}
```
kradalby, line 146: can you make this a `strings.Builder` and `fmt.Fprintf` for setting the settings with format?

kradalby:
This looks in general good, some comments, and a couple of failing tests to check out. Thanks!

hscontrol/handlers.go
```diff
@@ -71,37 +71,34 @@ func (h *Headscale) VerifyHandler(
...
-	body, err := io.ReadAll(req.Body)
-	if err != nil {
-		log.Error().
-			Str("handler", "/verify").
-			Err(err).
-			Msg("Cannot read request body")
-		http.Error(writer, "Internal error", http.StatusInternalServerError)
-		return
-	}
+	doVerify := func() (bool, error) {
```
kradalby, line 74: Can you make it a func so it doesnt have to be instantiated every time the verify handler is called?

ArcticLampyrid: I don't think it is to be instantiated every time, this function should be inline, which is the most basic optimization. It's hard to believe Golang wouldn't do it.

But if you request to do so, then I will fix it.

kradalby:
Thank you, given passing tests this should be good to go

ArcticLampyrid:
I believe this PR is ready for review and merging. The failures in the checks seem unrelated to the changes introduced in this PR. Specifically:

- `TestDERPServerWebsocketScenario` failure appears to be caused by tailscale/tailscale@020cacbe702463f14a5d2d5427819c491c7e6578, which no longer links websockets by default other than `GOOS=js`.
- The warnings reported by `golangci-lint` are from files that were not modified in this PR.

Please let me know if there's anything further needed.

enoperm:
Okay... at glance I'm not sure how to fix it without either a custom client (compiled with `GOOS=js`) or custom client builds, both of which I believe we'd prefer to avoid, since the point is to ensure we are compatible with the official client, right?

On the other hand, I *specifically* put work into being able to detect this exact scenario (behaviour of the client-under-test changing) in spite of the interface (env var). Seeing it turn red, I feel like it was worth the effort. :)

kradalby:
moved discussion to: https://github.com/juanfont/headscale/issues/2241

kradalby (approved):

PR #2053, opened 2024-08-12
build docker images on PR

kradalby:
Sometimes we want people to test features in PRs and not everyone is used to using git, build go and docker.

This commit builds docker containers and pushes them to GHCR (not dockerhub) for testing on pushes to branches that has open pull requests to main using Ko.
This is configured to mimic the debug images produced by goreleaser.

### DIFF
--- a/.github/workflows/build-docker-pr.yml
+++ b/.github/workflows/build-docker-pr.yml
@@ -1,0 +1,71 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+concurrency:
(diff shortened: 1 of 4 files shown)


kradalby:
I think we need more permissions to the `GITHUB_TOKEN` to allow for pushes from PRs, sensibly disabled by default since it might be code executed by random people.

I think it is fine to enable this because we have "approve to run workflows" enabled, so we can do a quick review to check if something looks suspicious before allowing a pr container to be published.

@juanfont @ohdearaugustin

kradalby:
Think we also need a read access to look up PRs too so we can get the PR number for tag name.

PR #2057, opened 2024-08-12
Don't overwrite previously filled dnsConfig.Domains

SuperSandro2000:
Before we threw away the value from https://github.com/juanfont/headscale/blob/8b3af45091b07169699c7879f3e050e312e42cf1/hscontrol/types/config.go#L561

I am not sure if this was more correct. If so, we should drop the domains code in the loop.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


kradalby:
Thank you!, I have addressed this issue in #2034.

Issue #1923, opened 2024-05-02
[Bug] latest upstream version is wrongly detected


christian-heusel:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
2024-05-02T11:56:17Z WRN An updated version of Headscale has been found (0.23.0-alpha9 vs. your current v0.23.0-alpha11). Check it out https://github.com/juanfont/headscale/releases
```

### Expected Behavior

No log message when on the latest version

### Steps To Reproduce

1. start up any headscale version later than 9
2. observe the above warning

### Environment

```markdown
- OS: Debian trixie/sid with Headscale running in Docker
- Headscale version: v0.23.0-alpha11
- Tailscale version: -
```

### Runtime environment

- [ ] Headscale is behind a (reverse) proxy
- [X] Headscale runs in a container

### Anything else?

With the relevant code inside of this project that is causing this behaviour being: 
https://github.com/juanfont/headscale/blob/a9c568c801a514855396c7dcec031b3598457f20/cmd/headscale/cli/root.go#L72-L88

I have dug into the issue/code a little and found that the error seems to propagate from `github.com/hashicorp/go-version` or alteast the version that the `tcnksm/go-latest` dependency is using. I have come to this conclusion with this little test program:

```golang
package main

import (
	"fmt"

	"github.com/hashicorp/go-version"
	"github.com/rs/zerolog/log"
	"github.com/tcnksm/go-latest"
)

func main() {
	Version := "v0.23.0-alpha10"
	githubTag := &latest.GithubTag{
		Owner:      "juanfont",
		Repository: "headscale",
	}

	fetch, err := githubTag.Fetch()
	fmt.Println("Valid")
	for _, v := range fetch.Versions {
		fmt.Printf("%s\n", v)
	}

	fmt.Println("Malformed")
	for _, v := range fetch.Malformeds {
		fmt.Printf("%s\n", v)
	}

	fmt.Println("Versioncmp")
	ourV, err := version.NewVersion(Version)
	theirV, err := version.NewVersion("0.23.0-alpha9")

	fmt.Printf("versioncompare %s < %s == %t\n", ourV, theirV, ourV.LessThan(theirV))

	// Original code
	res, err := latest.Check(githubTag, Version)
	if err == nil && res.Outdated {
		log.Warn().Msgf(
[...]

PR #2058, opened 2024-08-13
build docker images on PR

kradalby:
Take 2, not from fork, supersedes #2053

Sometimes we want people to test features in PRs and
not everyone is used to using git, build go and docker.

This commit builds docker containers and pushes them to
GHCR (not dockerhub) for testing on pushes to branches
that has open pull requests to main using Ko.
This is configured to mimic the debug images produced
by goreleaser.

### DIFF
--- a/.github/workflows/build-docker-pr.yml
+++ b/.github/workflows/build-docker-pr.yml
@@ -1,0 +1,71 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+      - name: Get changed files
+        id: changed-files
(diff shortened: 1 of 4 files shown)


kradalby:
Looks like this isnt possible,

`pull_request` does not have secrets access, so cannot push.

`pull_request_target` has secret access, but runs on the base branch, not the new commits, so the changes would not be reflected.

chriswiggins:
@kradalby this should work, You need to make sure that the permissions are set for the `GITHUB_TOKEN` so that it is allowed to write to the package registry:

```yaml
permissions:
  pull-requests: write
  packages: write
```

You also need to authenticate to the registry first:
```yaml
        name: Login to GitHub Container Registry
        uses: docker/login-action@v3
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

masterwishx:
> Collaborator
> ###

last beta4 builded `...beta.4` instead of `...beta4`

kradalby:
@masterwishx this is in response to https://github.com/juanfont/headscale/issues/1923

kradalby:
@chriswiggins, I think I could not get it to work for PRs at all? and from what I read in the gh docs secrets are not available, but I might have the motivation to fight and try this again at another time...

kradalby:
@masterwishx oh dear, I'll post it to the #1923 issue.

Issue #2074, opened 2024-08-22
[Bug] Upgrade from version 0.22.3 to 0.23-beta2 failed database migration


mpoindexter:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Upgrading from 0.22.3 to 0.23-beta2 failed on database migration with this error:
```
{"level":"fatal","error":"ERROR: insert or update on table \"routes\" violates foreign key constraint \"fk_nodes_routes\" (SQLSTATE 23503)","time":1724359938,"message":"Migration failed: ERROR: insert or update on table \"routes\" violates foreign key constraint \"fk_nodes_routes\" (SQLSTATE 23503)"}
```

### Expected Behavior

Migration completes successfully

### Steps To Reproduce

Have an existing database where there are rows in the `routes` table that reference nodes that no longer exist in the `nodes` table. Not sure how it happened, but seems like I had a number of these, seems like older versions of the code could cause this to happen.

Running `delete from routes where node_id not in (select id from nodes);` and then attempting to upgrade again solved the issue, but ideally this would be done as part of the migration scripts.

### Environment

```markdown
- OS: any
- Headscale version: 0.23-beta2
- Tailscale version: any
```

### Runtime environment

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

kradalby:
I think I have fixed this in #2076, do you have the ability to test it?

It is some tech debts from the early days of non structured migrations that keep biting us.

mpoindexter:
In some sense the change in #2076 works in that migration does not fail. It does mean that there is now a difference in the structure of the `routes` table between a fresh database and my upgraded one - specifically, the foreign key between `routes` and `node` does not get created. This doesn't seem ideal, IMO it is best to have the migration scripts remove any invalid data vs continuing to have data that doesn't conform to the intended schema still around and creating problems.

mpoindexter:
I tried as an alternate fix the following code vs only running automigrate if the table does not exist:
```
+                                       if tx.Migrator().HasTable(&types.Route{}) && tx.Migrator().HasTable(&types.Node{}) {
+                                               err := tx.Exec("delete from routes where node_id not in (select id from nodes)").Error
+                                               if err != nil {
+                                                       return err
+                                               }
+                                       }
+
                                        err = tx.AutoMigrate(&types.Route{})
                                        if err != nil {
                                                return err
```

[...]

Issue #2063, opened 2024-08-20
[Bug] Upgrade from 0.22.3 to 0.23.0-beta2 loses/forgets routes


RichardMore:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Routes that are advertised and approved in instance running version 0.22.3 are lost/forgotten after upgrading the config to 0.23.0-beta2. Also re-advertising the routes on the nodes that already shared them do not show up at all, restarting device or tailscale service doesn't seem to have any effect.

### Expected Behavior

Routes migrated, show up under routes and other nodes can access them.

### Steps To Reproduce

docker-compose.yml
```
services:
  headscale:
    image: headscale/headscale:0.22.3
    restart: unless-stopped
    networks:
      - traefik
    volumes:
      - ./headscale-config:/etc/headscale
      - ./headscale-data:/var/lib/headscale
    command: headscale serve
    labels:
      - "traefik.enable=true"
      - "traefik.docker.network=traefik"
      - "traefik.http.routers.headscale.entrypoints=https"
      - "traefik.http.routers.headscale.rule=Host(`X.Y`)"
      - "traefik.http.routers.headscale.tls.certResolver=ledns"
      - "traefik.http.services.headscale.loadbalancer.server.port=8080"
networks:
  traefik:
    external: true
```
Advertise routes, check that the show up
Change compose file `image: headscale/headscale:v0.23.0-beta2`, `command: serve`
Check routes, and re-advertise

### Environment

```markdown
- OS: Ubuntu 20.04.6 LTS
- Headscale version: 0.22.3 to 0.23.0-beta2
- Tailscale version:  1.70.0
```

### Runtime environment

- [X] Headscale is behind a (reverse) proxy
- [X] Headscale runs in a container

kradalby:
Could you query in the database to see if they are present? Are they dropped or just not migrated/readable?

RichardMore:
The routes table is empty, advertising new route is not populating it either.

kradalby:
do you have a copy of the database I could have?

RichardMore:
I've emailed the upgraded DB to you

kradalby:
Do you also have the 0.22.3 backup? that is what I would need for testing.

kradalby:
[...]

PR #2083, opened 2024-08-28
Fix FKs on sqllite migrations

mpoindexter:
My take on the right way to solve #2074 and #2063 that leaves the DB in a consistent state whether it's new or migrated

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

## Summary by CodeRabbit

- **New Features**
  - Improved database migration process, ensuring more reliable management of routes during updates.
  - Added functionality to temporarily disable foreign key constraints during migrations, enhancing data integrity.

- **Bug Fixes**
  - Addressed potential data loss by removing invalid routes associated with non-existent nodes before migration.

### DIFF
--- a/hscontrol/db/db.go
+++ b/hscontrol/db/db.go
@@ -123,21 +123,16 @@ func NewHeadscaleDatabase(
 						}
 					}
 
-					// Only run automigrate Route table if it does not exist. It has only been
-					// changed once, when machines where renamed to nodes, which is covered
-					// further up. This whole initial integration is a mess and if AutoMigrate
-					// is ran on a 0.22 to 0.23 update, it will wipe all the routes.
+					// Remove any invalid routes associated with a node that does not exist.
 					if tx.Migrator().HasTable(&types.Route{}) && tx.Migrator().HasTable(&types.Node{}) {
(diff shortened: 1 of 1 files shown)


kradalby:
Much appreciated, thank you, I'm sorry for not getting to this quicker, a lot of things is taking my time so very happy with the help. Is there any way we can add a test that will fail without this change so we ensure changes to this code does not regress? Like using one of the databases added in the previous PR and validating that the FK is there?

I'm happy to have a look when I next have time, but if you have an opportunity that would also be helpful.

mpoindexter:
I guess the existing tests you've added will cover the regression case: without the proper FK handling, the tests will fail with an empty routes table now that the conditions on the AutoMigrate are removed. I suppose I could write a test that a particular constraint exists in a migrated database, but that seems like an overly specific test for a general fix. I guess the only proper test would be to ensure that the schema of a migrated DB and new DB are the same, but that seems like a lot of work.

kradalby:
Sure, make sense, thank you. I will review this later today/tomorrow properly and get it in as the tests pass.

kradalby (approved):

PR #2089, opened 2024-08-30
validate policy against nodes, error if not valid

### LINKED ISSUE #2044: [Feature] Headscale policy set validate ACL before applying?
### Use case

Is it possible to have the headscale policy set command run validation before applying the policy and only load the policy if it passed syntax checking.

### Description

I accidentally put autogroup:internet instead of autogroup:internet:* and it really didn't like that.

Basically all the clients started throwing back 

`Could not get the create map update error="strconv.ParseUint: parsing \"internet\": invalid syntax" `  and I'm pretty sure they all dropped off the tailnet.

### Contribution

_No response_

### LINKED ISSUE #2073: [Bug] "Could not get the create map update"

### Current Behavior

When trying to register a node I get the following error from headscale:
`ERR ../../../home/runner/work/headscale/headscale/hscontrol/poll.go:719 > Could not get the create map update error="invalid action" node=net.example.com node.id=1 omitPeers=false readOnly=false stream=true`

The command I used: 
`tailscale up --login-server https://net.example.com --advertise-exit-node=true --hostname=net.example.com --accept-dns=true`

However `headscale node ls` outputs:
```
ID | Hostname     | Name         | MachineKey | NodeKey | User  | IP addresses | Ephemeral | Last seen           | Expiration          | Connected | Expired
1  | net.example.com | net.example.com | [n04ss]    | [AWCgQ] | example | 100.64.0.1,  | false     | 2024-08-22 18:18:24 | 0001-01-01 00:00:00 | offline   | no
```

### Expected Behavior

The node should register and connect to the tailnet (or headnet?) correctly

### Steps To Reproduce

Unsure. Just try to register a node.  

### Environment

```markdown
- OS: Debian 12
- Headscale version: v0.23.0-beta2
- Tailscale version: 1.72.1
```

### Runtime environment

### Anything else?

I'm using caddy (version: 2.6.2) as a reverse proxy

kradalby:
this commit aims to improve the feedback of "runtime" policy errors which would only manifest when the rules are compiled to filter rules with nodes.

this change will in;

file-based mode load the nodes from the db and try to compile the rules on start up and return an error if they would not work as intended.

database-based mode prevent a new ACL being written to the database if it does not compile with the current set of node.

Fixes #2073
Fixes #2044

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation for Access Control List (ACL) policies to prevent runtime errors from invalid configurations.
	- Improved error reporting with additional context when setting policies, aiding in better issue traceability.

- **Bug Fixes**
	- Resolved potential issues related to policy application by ensuring policies are validated against the current database state.
	- Added tests to validate error handling for invalid policy configurations, ensuring system integrity.

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -37,6 +37,7 @@ jobs:
           - TestNodeRenameCommand
           - TestNodeMoveCommand
           - TestPolicyCommand
+          - TestPolicyBrokenConfigCommand
           - TestResolveMagicDNS
           - TestValidateResolvConf
           - TestDERPServerScenario
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -1001,6 +1001,32 @@ func (h *Headscale) loadACLPolicy() error {
 		if err != nil {
 			return fmt.Errorf("failed to load ACL policy from file: %w", err)
 		}
+
+		// Validate and reject configuration that would error when applied
+		// when creating a map response. This requires nodes, so there is still
+		// a scenario where they might be allowed if the server has no nodes
+		// yet, but it should help for the general case and for hot reloading
+		// configurations.
+		// Note that this check is only done for file-based policies in this function
+		// as the database-based policies are checked in the gRPC API where it is not
+		// allowed to be written to the database.
+		nodes, err := h.db.ListNodes()
+		if err != nil {
+			return fmt.Errorf("loading nodes from database to validate policy: %w", err)
+		}
+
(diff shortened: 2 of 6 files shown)


kradalby:
@pallabpain What do you think about this?

juanfont (approved):

pallabpain:
> @pallabpain What do you think about this?

Sorry, I did not get a chance to look at this. I was thinking more along the lines of validating the incoming policy bytes at the time of unmarshaling it into the ACL type. But this looks good. 👍🏼

PR #2093, opened 2024-08-31
Fix 764

### LINKED ISSUE #764: Headscale fails to activate clients with postgresql backend

mpoindexter:
When a zero time value is loaded from JSON or a DB in a way that assigns it the local timezone, it does not roudtrip in JSON as a value for which IsZero returns true. This causes KeyExpiry to be treated as a far past value instead of a nilish value.

See https://github.com/golang/go/issues/57040

- [x] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #764 

## Summary by CodeRabbit

- **New Features**
	- Introduced new test cases to enhance testing for time-sensitive operations and PostgreSQL interactions.
	- Added a function to allow users to specify timezones for container configurations.
	- Enhanced test configurability with new options for flexible testing scenarios.
  
- **Bug Fixes**
	- Resolved issues related to node registration in non-UTC timezones for PostgreSQL databases.

- **Documentation**
	- Updated CHANGELOG to reflect recent updates and fixes.

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -45,6 +45,7 @@ jobs:
           - TestPingAllByIPPublicDERP
           - TestAuthKeyLogoutAndRelogin
           - TestEphemeral
+          - TestEphemeralInAlternateTimezone
           - TestEphemeral2006DeletedTooQuickly
           - TestPingAllByHostname
           - TestTaildrop
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,8 @@ after improving the test harness as part of adopting [#1460](https://github.com/
 - Make registration page easier to use on mobile devices
 - Make write-ahead-log default on and configurable for SQLite [#1985](https://github.com/juanfont/headscale/pull/1985)
 - Add APIs for managing headscale policy. [#1792](https://github.com/juanfont/headscale/pull/1792)
+- Fix for registering nodes using preauthkeys when running on a postgres database in a non-UTC timezone. [#764](https://github.com/juanfont/headscale/issues/764)
+- Make sure integration tests cover postgres for all scenarios
 
 ## 0.22.3 (2023-05-12)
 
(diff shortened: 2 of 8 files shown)


kradalby:
Thank you, much appreciated, for this change I would need a integration test that first verify the problem, so something that when ran in sqlite/postgres mode returns pass/fail respectively, and then when this change is applied change to pass/pass.

If you have time to look at it would be great, or else I can see what I have time to do next week.

I recon that a fix for this issue also would merit a changelog entry.

mpoindexter:
The unit test I added would fail prior to this change, and now passes. It really actually has nothing to do with the database, but rather the `KeyExpiry` time object having a non-UTC timezone. Just the postgresql driver being used (pgx) happens to put all times into the local timezone, and the sqllite driver in use does not - but I guess that could change on the part of either driver since the the time represents the same instant of the zero value, so I think the test I added is the appropriate verification of this. If you disagree I can look at adding an integration test. I'll update the PR to add a changelog entry.

kradalby:
Integration tests to ensure full roundtrip via both drivers would be necessary. Both for this change and to ensure regressions, we have had libraries introduce silent failures before.

mpoindexter:
I guess it really depends on the timezone the tests run in - the existing integration tests should trip it as long as the timezone the headscale server runs in is non-UTC. I suspect on the integration tests the timezone is set to UTC and hence doesn't happen. I can take a look next week and see how hard it would be to introduce that, I'm not familiar with your integration test setup.

mpoindexter:
OK, added an integration test for this, as I suspected it was enough to ensure that the headscale container was in an alternate timezone using existing test scenarios. I added a test method that forces the container into an alternate TZ and runs it that fails without this change.  Also found that not all integration tests respected the env variable to use postgresql instead of sqlite, so fixed that as well, and added changelog entries.

kradalby (approved):

PR #2116, opened 2024-09-09
dont override golangci go

kradalby:


juanfont (approved):

flake.nix
```diff
@@ -57,9 +57,11 @@
           subPackages = ["protoc-gen-grpc-gateway" "protoc-gen-openapiv2"];
         };
 
-        golangci-lint = prev.golangci-lint.override {
-          buildGoModule = buildGo;
-        };
+        # Upstream does not override buildGoModule properly,
```
SuperSandro2000, line 60: We want to use the latest go for golangci-lint anyway regardless of what version headscale is using.

kradalby: Yes, it was building with 1.22 upstream, then I set it to 1.23, but upstream changed to take `buildGo123Module` instead which made it a bit awkward.

golangci-lint gets sad when you use the wrong go version to build it, so it was to make it use the same as ours.

SuperSandro2000: It gets sad, if you use a newer go version than it was built with. Building golangci-lint with a newer version than later is being used ,is mostly fine.

PR #2120, opened 2024-09-09
changed the html string in the hscontrol/handlers.go file using go-elem

amha-mersha:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

## Summary by CodeRabbit

- **New Features**
	- Enhanced the registration web API with a modular HTML structure for improved maintainability and readability.
	- Introduced dynamic rendering of HTML elements using the `elem-go` library, allowing for easier updates to the registration page.
	- Added comprehensive HTML guides for configuring Tailscale on Apple and Windows devices, improving user experience.

- **Bug Fixes**
	- Resolved issues related to inline HTML by implementing a more structured approach, reducing potential errors in the registration process.


hscontrol/handlers.go
```diff
@@ -138,34 +141,39 @@ type registerWebAPITemplateConfig struct {
...
+		elem.Body(attrs.Props{
+			attrs.Style: styles.Props{
+				styles.FontFamily: "sans",
+			}.ToInline(),
+		},
+			elem.H1(nil, elem.Text("headscale")),
+			elem.H2(nil, elem.Text("Machine registration")),
+			elem.P(nil, elem.Text("Run the command below in the headscale server to add this machine to your network:")),
+			elem.Code(attrs.Props{attrs.Style: codeStyleRegisterWebAPI.ToInline()},
+				elem.Raw(`headscale nodes register --user USERNAME --key {{.Key}}`)),
```
kradalby, line 172: Lets drop the use of the templates, this is the only thing needing it, so you can make this whole thing a function, and then pass a `key string` argument, which is then used here
I think it would be something like:
```suggestion
				elem.Text(fmt.Sprintf("headscale nodes register --user USERNAME --key %s", key)),
```

hscontrol/handlers.go
```diff
@@ -138,34 +141,39 @@ type registerWebAPITemplateConfig struct {
...
+		)))
+
 var registerWebAPITemplate = template.Must(
```
kradalby, line 175: then this can be removed and the places that calls it will call the function you create above with the render method.

amha-mersha: I made it a function that returns a *elem.Element type and the caller calls Render() on it. I [`defined the function](https://github.com/juanfont/headscale/pull/2120/commits/4d4964d1d13f224855f7c9eb62f2cddfed72b2ac#diff-b623bbee13a6f205d5f50b7d972d994547a961f40e25cbd5fd96478ce7325d96R149-R151). I didn't add error for the return I didn't find any place that needed it except if check for the config.Key is needed. Should I
[here](https://github.com/juanfont/headscale/pull/2120/commits/4d4964d1d13f224855f7c9eb62f2cddfed72b2ac#diff-b623bbee13a6f205d5f50b7d972d994547a961f40e25cbd5fd96478ce7325d96R209-R215) is the where the function is called

hscontrol/handlers.go
```diff
@@ -148,17 +146,15 @@ var codeStyleRegisterWebAPI = styles.Props{
 	styles.BackgroundColor: "#eee",
 }
 
-var htmlRegisterWebAPI = elem.Html(nil,
-	elem.Head(
-		nil,
-		elem.Title(
+func registerWebAPITemplate(config registerWebAPITemplateConfig) *elem.Element {
```
kradalby, line 149: you can drop the whole config and just make this a `string`

hscontrol/handlers.go
```diff
@@ -169,11 +165,11 @@ var htmlRegisterWebAPI = elem.Html(nil,
 			elem.H2(nil, elem.Text("Machine registration")),
 			elem.P(nil, elem.Text("Run the command below in the headscale server to add this machine to your network:")),
 			elem.Code(attrs.Props{attrs.Style: codeStyleRegisterWebAPI.ToInline()},
-				elem.Raw(`headscale nodes register --user USERNAME --key {{.Key}}`)),
-		)))
-
-var registerWebAPITemplate = template.Must(
-	template.New("registerweb").Parse(htmlRegisterWebAPI.Render()))
+				elem.Text(fmt.Sprintf("headscale nodes register --user USERNAME --key %s", config.Key)),
```
kradalby, line 168: so this is just `key`

hscontrol/handlers.go
kradalby: then the config struct can be deleted.

kradalby, line 149: This can be renamed to something like `registerWebHTML`

hscontrol/handlers.go
```diff
@@ -210,34 +206,17 @@ func (h *Headscale) RegisterWebAPI(
 		return
 	}
 
-	var content bytes.Buffer
-	if err := registerWebAPITemplate.Execute(&content, registerWebAPITemplateConfig{
+	var htmlContent = registerWebAPITemplate(registerWebAPITemplateConfig{
```
kradalby, line 209: remove this whole var

hscontrol/handlers.go
```diff
@@ -210,34 +206,17 @@ func (h *Headscale) RegisterWebAPI(
...
-
-		return
-	}
+	})
 
 	writer.Header().Set("Content-Type", "text/html; charset=utf-8")
 	writer.WriteHeader(http.StatusOK)
-	_, err = writer.Write(content.Bytes())
-	if err != nil {
+	if _, err := writer.Write([]byte(htmlContent.Render())); err != nil {
```
kradalby, line 215: and then you can call it here like:

`if _, err := writer.Write([]byte(registerWebHTML(machineKey.String()).Render())); err != nil {`

amha-mersha:
I have implemented the suggestions and have also replaced the windows.html and apple.html files with windows.go and apple.go files.

hscontrol/handlers.go
```diff
@@ -134,38 +135,37 @@ func (h *Headscale) HealthHandler(
 	respond(nil)
 }
 
-type registerWebAPITemplateConfig struct {
-	Key string
+var codeStyleRegisterWebAPI = styles.Props{
```
kradalby, line 138: Lets move this to `templates/`, making as much as possible reusable

hscontrol/handlers.go
```diff
@@ -134,38 +135,37 @@ func (h *Headscale) HealthHandler(
...
-		<h1>headscale</h1>
-		<h2>Machine registration</h2>
-		<p>
-			Run the command below in the headscale server to add this machine to your network:
-		</p>
-		<code>headscale nodes register --user USERNAME --key {{.Key}}</code>
-	</body>
-</html>
-`))
+func registerWebHTML(key string) *elem.Element {
```
kradalby, line 145: Lets make a templates package, `templates`, rename it to something like `RegisterWeb` and then we can make more of the generic HTML reusable.

hscontrol/handlers.go
```diff
@@ -134,38 +135,37 @@ func (h *Headscale) HealthHandler(
...
+	return elem.Html(nil,
+		elem.Head(
```
kradalby, line 147: `elem.Html`,`elem.Head`, `elem.Body` can be made reusable between `RegisterWeb`, `Windows` and `Apple`.

hscontrol/templates/apple.go
```diff
@@ -0,0 +1,240 @@
+package templates
+
+import (
+	"fmt"
+
+	"github.com/chasefleming/elem-go"
+	"github.com/chasefleming/elem-go/attrs"
+)
+
+func HeadscaleApple(url string) *elem.Element {
```
kradalby, line 10: This can just be called `Apple`

hscontrol/templates/apple.go
```diff
@@ -0,0 +1,240 @@
...
+			}),
+			elem.Meta(attrs.Props{
+				attrs.Name:    "viewport",
+				attrs.Content: "width=device-width, initial-scale=1.0",
+			}),
+			elem.Title(nil,
+				elem.Text("headscale - Apple")),
+			elem.Style(nil,
+				elem.CSS(`
+      body {
```
kradalby, line 30: Lets make this use the elem-go style syntax and reuse it between `Windows` and `Apple`.

hscontrol/templates/apple.go
```diff
@@ -0,0 +1,240 @@
...
+import (
+	"fmt"
+
+	"github.com/chasefleming/elem-go"
+	"github.com/chasefleming/elem-go/attrs"
+)
+
+func HeadscaleApple(url string) *elem.Element {
+	return elem.Html(nil,
+		elem.Head(attrs.Props{
```
kradalby, line 12: This whole section could be made generic and reused.

hscontrol/templates/apple.go
```diff
@@ -0,0 +1,240 @@
...
+				),
+			),
+			elem.H3(nil,
+				elem.Text("GUI"),
+			),
+			elem.Ol(nil,
+				elem.Li(nil,
+					elem.Text("ALT + Click the Tailscale icon in the menu and hover over the Debug menu")),
+				elem.Li(nil,
+					elem.Text("Under \"Custom Login Server\", select \"Add Account...\"")),
```
kradalby, line 76: When having a string in go that uses `"`, you should do:

```suggestion
					elem.Text(`Under "Custom Login Server", select "Add Account..."`)),
```

to not have to escape everything.

kradalby (commented):
This looks great, here are some more tips for making it a bit more DRY and reusable. And one Go tip.

Note that at least apple.html (maybe also windows.html) was updated and you need to have a look to reflect the latest state of that.

kradalby:
@amha-mersha How is this going? sorry for hammering you with comments :)

amha-mersha:
Hello @kradalby , sorry I had some other work and forgot about it. I will correct them today and push another commit. Thank you for the constant feedbacks and the follow up, It really helps me learn software production

hscontrol/templates/apple.go
amha-mersha: I thought if I do that, It will be redundant for some elements, like the style props should be passed to each header element. Let me do it the way you suggested and I'll let you know

amha-mersha:
@kradalby hopefully I have made it more DRY now.

hscontrol/templates/apple.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/chasefleming/elem-go"
+	"github.com/chasefleming/elem-go/attrs"
+)
+
+func Apple(url string) *elem.Element {
+	return HtmlStructure(
+		elem.Title(nil,
+			elem.Text("headscale - Apple")),
+		elem.Body(attrs.Props{
+			attrs.Style: bodySyle.ToInline(),
```
kradalby, line 15: Typo
```suggestion
			attrs.Style: bodyStyle.ToInline(),
```

kradalby (commented):
Only a small nit, I think it looks good now. Could you try to rebase this so it does not contain all the other commits since you started?

I would like to have @nblock have a quick look over the html pages so we are sure we didnt loose any info.

amha-mersha:
@kradalby  @nblock  I accidentally closed this PR when trying to clean up the commits, so I created a new PR.

PR #2131, opened 2024-09-13
Fix nodes not getting set to offline when disconnects

### LINKED ISSUE #2129: [Bug] online status doesn't change if connection is interrupted

### Current Behavior

If the connection of a client is interrupted (pull the cable, disconnect from wifi) headscale never changes its status to offline.
The state changes to offline when I restart headscale or caddy. (which terminates all connections)

### Expected Behavior

The node should go offline after some time.

### Steps To Reproduce

Connect a device and inerrupt it's internet connection.

### Environment

```markdown
- OS: Flatcar Linux
- Headscale version: v0.23.0-beta.5
- Tailscale version: 1.72.1

I tested it with an Android Phone and my Linux machine. (Ubuntu 22.04)
The reverse proxy is Caddy with one line standard configuration reverse proxy setup. Headscale runs in a Docker container with sqlite DB.
```

### Runtime environment

- [X] Headscale runs in a container

### Anything else?

I can provide further logs and dumps if this problem does not appear in another setup.

kradalby:
Fixes #2129

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -58,6 +58,7 @@ jobs:
           - TestEnableDisableAutoApprovedRoute
           - TestAutoApprovedSubRoute2068
           - TestSubnetRouteACL
+          - TestHASubnetRouterFailoverWhenNodeDisconnects2129
           - TestHeadscale
           - TestCreateTailscale
           - TestTailscaleNodesJoiningHeadcale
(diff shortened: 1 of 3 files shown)


kradalby:
The work missing here is that we cannot in go reliable determine when a node disconnects, this is handled further down the stack.

To resolve the linked issue, and have this test passing, we need to implement something that routinely "pings" any router, or at least routers that are determined to be HA.

This can probably be implemented with something like PingRequest(https://github.com/tailscale/tailscale/blob/main/tailcfg/tailcfg.go#L1717)

PR #2132, opened 2024-09-13
Add compatibility with only websocket-capable clients

enoperm:
- [ ] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

I have not touched `CHANGELOG.md`, as I can see Headscale already entered a release candidate phase and I do not expect anything other than bugfixes would be accepted at this point. I'll do so if/when the changes are deemed acceptable and it is clear what version of Headscale they may become a part of.

This is not a new feature, nor does it change the configuration, the patches merely improve compatibility with existing Tailscale client code, hence no documentation changes (although I'm considering that maybe it could save someone a bunch of debugging if the docs mentioned that the Tailscale DERP-over-websocket code ignores the port number in the configured DERP map).

I'd refrain from calling these changes a *bugfix*, though, because not supporting a (currently entirely optional) client capability does not quite sound like a bug.

## Release Notes

- **New Features**
  - Enhanced WebSocket support for the DERP server, allowing better communication and flexibility.
  - Added new logging capabilities for Tailscale clients and containers, improving debugging and monitoring.
  - Introduced a configuration option to enable or disable WebSocket DERP connections.
  - New test scenarios for validating client configurations with WebSocket and plain connections.
  - Added a new test case specifically for WebSocket interactions with the DERP server.

- **Bug Fixes**
  - Improved error handling in integration tests to enhance robustness.

- **Tests**
  - Expanded testing coverage for WebSocket interactions and DERP server scenarios.

### DIFF
--- a/integration/dockertestutil/logs.go
+++ b/integration/dockertestutil/logs.go
@@ -46,12 +46,7 @@ func SaveLog(
 		return "", "", err
 	}
 
-	// Wouldn't it be simpler to
-	// open and wrap the destination files in a
-	// bufio.Writer, and pass those in docker.LogsOptions?
-	var stdout bytes.Buffer
-	var stderr bytes.Buffer
-
+	var stdout, stderr bytes.Buffer
 	err = WriteLog(pool, resource, &stdout, &stderr)
 	if err != nil {
 		return "", "", err
--- a/integration/dockertestutil/logs.go
+++ b/integration/dockertestutil/logs.go
@@ -3,6 +3,7 @@ package dockertestutil
 import (
 	"bytes"
 	"context"
+	"io"
 	"log"
 	"os"
 	"path"
@@ -12,6 +13,28 @@ import (
 )
 
 const filePerm = 0o644
+
+func WriteLog(
+	pool *dockertest.Pool,
+	resource *dockertest.Resource,
+	stdout io.Writer,
+	stderr io.Writer,
+) error {
+	return pool.Client.Logs(
+		docker.LogsOptions{
+			Context:      context.TODO(),
+			Container:    resource.Container.ID,
+			OutputStream: stdout,
+			ErrorStream:  stderr,
+			Tail:         "all",
+			RawTerminal:  false,
+			Stdout:       true,
+			Stderr:       true,
+			Follow:       false,
+			Timestamps:   false,
+		},
+	)
+}
 
 func SaveLog(
 	pool *dockertest.Pool,
@@ -23,23 +46,13 @@ func SaveLog(
 		return "", "", err
 	}
 
+	// Wouldn't it be simpler to
+	// open and wrap the destination files in a
+	// bufio.Writer, and pass those in docker.LogsOptions?
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 
-	err = pool.Client.Logs(
-		docker.LogsOptions{
-			Context:      context.TODO(),
-			Container:    resource.Container.ID,
-			OutputStream: &stdout,
-			ErrorStream:  &stderr,
-			Tail:         "all",
-			RawTerminal:  false,
-			Stdout:       true,
-			Stderr:       true,
-			Follow:       false,
-			Timestamps:   false,
-		},
-	)
+	err = WriteLog(pool, resource, &stdout, &stderr)
 	if err != nil {
 		return "", "", err
 	}
--- a/integration/embedded_derp_test.go
+++ b/integration/embedded_derp_test.go
@@ -81,7 +81,7 @@ func TestDERPServerWebsocketScenario(t *testing.T) {
 // so marking it as a test helper would be counterproductive.
 //
 //nolint:thelper
-func DERPServerScenario(
+func derpServerScenario(
 	t *testing.T,
 	spec map[string]ClientsSpec,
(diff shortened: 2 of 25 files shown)


enoperm:
I do not remember touching `integration/general_test.go`. I guess there may have been upstream patches I forgot to merge since yesterday evening. Let me see if a merge makes it go away.

enoperm:
`git fetch` did retrieve any new commits, so I guess I messed up a rebase earlier.

integration/dockertestutil/logs.go
```diff
@@ -40,6 +34,25 @@ func SaveLog(
...
+	err := os.MkdirAll(basePath, os.ModePerm)
+	if err != nil {
+		return "", "", err
+	}
+
+	// Wouldn't it be simpler to
+	// open and wrap the destination files in a
+	// bufio.Writer, and pass those in docker.LogsOptions?
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
```
kradalby, line 53: Potentially, not tested or was aware of it, happy for that to be changed, maybe you can test in a followup pr?

integration/embedded_derp_test.go
```diff
@@ -15,13 +15,77 @@ import (
...
+		}
+	})
+}
+
+// This function implements the common parts of a DERP scenario,
+// we *want* it to show up in stacktraces,
+// so marking it as a test helper would be counterproductive.
+//
+//nolint:thelper
+func DERPServerScenario(
```
kradalby, line 84: Can you make this `derpServerScenario`?

kradalby (commented):
Nothing really jumps out at me as problematic and unmaintainable, so I'm happy for this to move along as long as the tests pass.

integration/dockertestutil/logs.go
enoperm: I haven't tested it either, I just thought maybe doing the equivalent of `> output.file` is simpler than doing an extra lap around the same thing. I don't expect any significant improvements or breakages from it either, other than the code becoming shorter. I don't mind doing an extra PR, but should it include only this spot, or all other file operations that may involve redundant `io.Writer`s?

integration/embedded_derp_test.go
enoperm: Done. Also merged current `main` in the hopes that it makes `golangci-lint` only look at the changes introduced in this branch again.

enoperm: Actually wait, I commited a version before I followed up the change at call sites.

enoperm: Okay, should be fine now.

integration/dockertestutil/logs.go
kradalby: If you have time, all would of course be useful. Since this is in the integration tests, its not really that critical, but always nice to have it neater.

enoperm: I tried to look for other manually buffered disk writes with `git grep`, but as far as I can see, this was the only spot where it happened.

I can submit an MR with the changes, but... the code did not become much shorter. The tally comes to -5 lines, three of which were the comment I added. I could *still* see some value in it, because unlike `bytes.Buffer`, which (at least if I remember correctly) by default extends however many times it needs to fit things into memory, `bufio.Writer` has a defined buffer size and automatically flushes when it is about to cross it. So, in theory it *could* make it so that longer logs can be dumped without issue. I'm still waiting for the [integration tests to run](https://github.com/enoperm/headscale/pull/4) on github, but I do not expect problems with it.

(My machine does not seem to like the idea of running the entire test suite at once too much - best I could do was 34 to go before it timed out, and I think I might have raised the timeout before reaching that result).

enoperm:
> @enoperm looks good now, could you add a changelog entry?

Sure - would prefer that I add a specific version number, or use a placeholder like **NEXT**? The current state of `CHANGELOG.md` does not reflect the next intended release yet. I think it also needs to be amended a bit, because it claims that yesterday's release was made in 2023. :)

integration/dockertestutil/logs.go
enoperm: Aaaand it actually failed ina few tests - must have forgot about a flag.

kradalby:
Next is good like I did here https://github.com/juanfont/headscale/pull/2020/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR3, the release cycle should not be that long this time ™️

integration/dockertestutil/logs.go
kradalby: fair, this isnt really critical, if it isnt much simpler or neater, it can be dropped. I'm happy either way as it currently does work, good suggestion tho. You can submit it if it seem to work, and we can take if from there but you dont have to spend that much time on it.

enoperm: Removed the comment for now. I *may* take another look at it, but at this point I would not reallly mind leaving it as it is either.

enoperm:
Added new changelog entry, updated the timestamp on the `0.23` release to align better with reality.

kradalby:
> Added new changelog entry, updated the timestamp on the `0.23` release to align more with reality.

Haha thanks, I must have been a bit quick there, well, technically I was really slow 🫠

kradalby (approved):

PR #2149, opened 2024-09-24
remove versions older than 1.56

kradalby:
From 0.23.0, headscale has a goal of supporting the 10 latest versions of tailscale, this PR removes versions older than 1.56.

This allows us to remove a bunch of code, some duplicate! 🎉 


ohdearaugustin (approved):
Looks good.

Issue #2164, opened 2024-10-02
[Bug] Tailscale HEAD (1eaad7d3d) is broken in integration tests


kradalby:
### Current Behavior

Reauthentication is broken in integration tests for Tailscale clients built from HEAD due to https://github.com/tailscale/tailscale/commit/1eaad7d3deb0815e8932e913ca1a862afa34db38.

It looks like the commit make fast reconnects force them to use 443, so it should not affect users as long as they use HTTPS.

The solution is likely to document that you _more or less_ need to run HTTPS and to make integration tests that reauth is served over 443.

kradalby:
The error at hand will only occur if you have an headscale running on http, not answering on 443, and the client tries to reconnect/reauth within 2 minutes. If it reconnects after 2 minutes, it should be fine.

PR #2184, opened 2024-10-09
use https for other relogin tests

kradalby:
continuation of #2164

### DIFF
--- a/integration/auth_web_flow_test.go
+++ b/integration/auth_web_flow_test.go
@@ -80,7 +80,12 @@ func TestAuthWebFlowLogoutAndRelogin(t *testing.T) {
 		"user2": len(MustTestVersions),
 	}
 
-	err = scenario.CreateHeadscaleEnv(spec, hsic.WithTestName("weblogout"))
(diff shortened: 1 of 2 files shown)


hopleus:
@kradalby Perhaps you should add `hsic.WithTLS()` to `auth_oidc_test.go`?

hopleus:
@kradalby 

When `integrationTests` were run locally (using `WithTls()`), the same problem occurred as in the current tests.
When the following code was added to `runTailscaleUp`.
```go
loginURL.Host = fmt.Sprintf('%s:8080', headscale.GetIP())
loginURL.Scheme = 'http'

if len(headscale.GetCert()) > 0 {
	loginURL.Scheme = 'https'
}

insecureTransport := &http.Transport{
	TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // nolint
}
```
the error stopped occurring

The error itself `2024/10/09 09:52:00 failed to register client (ts-1-72-jqug4h): failed to parse auth page` contains `Client sent an HTTP request to an HTTPS server.`.

PR #2233, opened 2024-11-12
Use a more direct and less error-prone return value

pinglanlu:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.


kradalby:
There are probably a lot more of these, it either needs to be fully consistent or should not matter. I dont agree that the impact is that grand, happy for you to take a pass over the whole codebase, but a single change like this doesnt make too much sense imo.

PR #2247, opened 2024-11-19
Websocket derp test fixes

### LINKED ISSUE #2241: [Bug]

### Current Behavior

TestDERPServerWebsocketScenario is broken after websockets was removed from all clients but JS one: https://github.com/tailscale/tailscale/commit/020cacbe702463f14a5d2d5427819c491c7e6578

### Expected Behavior

> Okay... at glance I'm not sure how to fix it without either a custom client (compiled with `GOOS=js`) or custom client builds, both of which I believe we'd prefer to avoid, since the point is to ensure we are compatible with the official client, right?
> 

I agree that it is something to in general avoid, I am also not sure if you can run the JS client outside of a browser? Maybe it can be ran with NodeJS in a container? It looks like it can be forced into a Go client with a build flag?
I must admit I am not stoked about having to maintain that kind of container to test this, but maybe it can be done for only this test. I dont really want to remove the feature, but if it is untestable that might be a step to go for _if it breaks in the future_, it can be left until someone notices, but we are not maintaining features that cant be tested.

> On the other hand, I _specifically_ put work into being able to detect this exact scenario (behaviour of the client-under-test changing) in spite of the interface (env var). Seeing it turn red, it feels like it was worth the effort. :)

I agree, that is good, now we just need to make sure we continue to have tests.

### Steps To Reproduce

.

### Environment

```markdown
- Tailscale version: HEAD
```

### Runtime environment

### Anything else?

_No response_

enoperm:
- [ ] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

Fixes #2241

### DIFF
--- a/integration/embedded_derp_test.go
+++ b/integration/embedded_derp_test.go
@@ -55,7 +55,7 @@ func TestDERPServerWebsocketScenario(t *testing.T) {
 	spec := map[string]ClientsSpec{
 		"user1": {
 			Plain:         0,
-			WebsocketDERP: len(MustTestVersions),
+			WebsocketDERP: 2,
 		},
 	}
 
@@ -239,10 +239,13 @@ func (s *EmbeddedDERPServerScenario) CreateHeadscaleEnv(
 
 		if clientCount.WebsocketDERP > 0 {
 			// Containers that use DERP-over-WebSocket
+			// Note that these clients *must* be built
+			// from source, which is currently
+			// only done for HEAD.
 			err = s.CreateTailscaleIsolatedNodesInUser(
 				hash,
 				userName,
-				"all",
+				"head",
 				clientCount.WebsocketDERP,
 				tsic.WithWebsocketDERP(true),
 			)
--- a/integration/embedded_derp_test.go
+++ b/integration/embedded_derp_test.go
@@ -245,7 +245,7 @@ func (s *EmbeddedDERPServerScenario) CreateHeadscaleEnv(
 			err = s.CreateTailscaleIsolatedNodesInUser(
 				hash,
 				userName,
-				"head",
+				tsic.VersionHead,
 				clientCount.WebsocketDERP,
 				tsic.WithWebsocketDERP(true),
 			)
--- a/integration/tsic/tsic.go
+++ b/integration/tsic/tsic.go
@@ -12,6 +12,7 @@ import (
 	"net/netip"
 	"net/url"
 	"os"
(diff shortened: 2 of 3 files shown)


integration/embedded_derp_test.go
```diff
@@ -239,10 +239,13 @@ func (s *EmbeddedDERPServerScenario) CreateHeadscaleEnv(
...
 		if clientCount.WebsocketDERP > 0 {
 			// Containers that use DERP-over-WebSocket
+			// Note that these clients *must* be built
+			// from source, which is currently
+			// only done for HEAD.
 			err = s.CreateTailscaleIsolatedNodesInUser(
 				hash,
 				userName,
-				"all",
+				"head",
```
kradalby, line 248: This can be the `VersionHead` constant you added

integration/tsic/tsic.go
```diff
@@ -175,6 +188,18 @@ func WithNetfilter(state string) Option {
 	}
 }
 
+// WithBuildTag adds an additional value to the `-tags=` parameter
+// of the Go compiler, allowing callers to customize the Tailscale client build.
+// This option is only meaningful when invoked on **HEAD** versions of the client.
+// Attempts to use it with any other version is a bug in the calling code.
+func WithBuildTag(tag string) Option {
+	return func(tsic *TailscaleInContainer) {
```
kradalby, line 196: I propose panicking here if the tsic passed is not `HEAD`, just to fail fast so the dev can fix it without faffing around.

kradalby (approved):
two nits, otherwise good

integration/embedded_derp_test.go
enoperm: Good idea, I probably missed this instance because I only looked at the linter output when I introduced that constant.

integration/tsic/tsic.go
enoperm: Done, for now I have opted to panic with the same error value I return from the other checks - although at this point I think it might be safe to remove the build tags check [here](https://github.com/juanfont/headscale/pull/2247/files#diff-2ba815dc33efc14a3e231f81a0cba347b8db54804f18602e12d6410865ffab74R284), but since I wrote it to be generic and check for *any* build time config, it may still catch mistakes on future build options. I am not against removing it if desired, though.

kradalby (approved):

kradalby:
Thanks!

Issue #2210, opened 2024-10-24
[Bug] testing for server_url containing base_domain is too restrictive


quite:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have `server_url: https://homer.example.com` and `dns.base_domain: h`. This does not work, headscale complains with `server_url cannot contain the base_domain [..]`.

@mtoohey31 recently noted the same issue in a [comment](https://github.com/juanfont/headscale/pull/2034#discussion_r1805755064) to https://github.com/juanfont/headscale/pull/2034:

> I could be wrong, but I wonder if the `string.Contains` check might be too strict of a test. I want to use (for example) `https://bar.foo.com` as my `server_url`, and just `foo` as my `dns.base_domain`, but this check rejects that. I've manually removed the check and this seems to work without issue.

### Expected Behavior

I don't want to be overly and unnecessarily restricted about the choice of `base_domain`.
I *think* that what should *not* be allowed are the following:
 
- `base_domain` that is a prefix of `server_url`'s hostname
- `base_domain` equal to `server_url`'s hostname
- maybe also `server_url`'s hostname that is a prefix of `base_domain`?

Possible plain prefix comparison is not suitable, but dot-separated parts needs to be taken into consideration separately? All this needs some more thinking through. Any ideas?

### Steps To Reproduce

I have `server_url: https://homer.example.com` and `dns.base_domain: h`. This does not work, headscale complains with `server_url cannot contain the base_domain [..]`.

### Environment

```markdown
- OS: Debian 12
- Headscale version: 0.23.0
- Tailscale version: 1.76.1
```

### Runtime environment

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

hopleus:
I assume that using a second-level domain in dns.base_domain, may cause conflicts, for example, when using `*.base.domain` at NPM or CloudFlare level (**example 1**), as well as possible conflicts with `server_url` (**example 2**).

**Example 1:**
There is a `test-stage.base.domain` that leads to a specific portal.
If a node with the same name test-stage is added, the portal will never open because the dns headscale will see this DNS record.

**Example 2:**
```
server_url: https://gateway.headscale.net
dns.base_domain: headscale.net
```
Add a node with the name gateway, that's it, the Headscale API working on `gateway.headscale.net` is no longer working, so the connection with the TailScale client is lost.

[...]

PR #2248, opened 2024-11-21
config: loosen up BaseDomain and ServerURL checks

motiejus:
Requirements [here][1]:

> OK:
> server_url: headscale.com, base: clients.headscale.com
> server_url: headscale.com, base: headscale.net
>
> Not OK:
> server_url: server.headscale.com, base: headscale.com
>
> Essentially we have to prevent the possibility where the headscale
> server has a URL which can also be assigned to a node.
>
> So for the Not OK scenario:
>
> if the server is: server.headscale.com, and a node joins with the name
> server, it will be assigned server.headscale.com and that will break
> the connection for nodes which will now try to connect to that node
> instead of the headscale server.

Fixes https://github.com/juanfont/headscale/issues/2210

[1]: https://github.com/juanfont/headscale/issues/2210#issuecomment-2488165187

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/hscontrol/types/config_test.go
+++ b/hscontrol/types/config_test.go
@@ -140,7 +140,7 @@ func TestReadConfig(t *testing.T) {
 				return LoadServerConfig()
 			},
 			want:    nil,
-			wantErr: "server_url cannot be a suffix of the base_domain, this will cause the headscale server and embedded DERP to become unreachable from the Tailscale node.",
+			wantErr: errServerURLSuffix.Error(),
 		},
 		{
 			name:       "base-domain-not-in-server-url",
@@ -361,6 +361,10 @@ func TestSafeServerURL(t *testing.T) {
 		{
 			serverURL:  "https://headscale.com",
 			baseDomain: "clients.subdomain.headscale.com",
+		},
+		{
+			serverURL:  "https://headscale.kristoffer.com",
+			baseDomain: "mybase",
 		},
 		{
 			serverURL:  "https://server.headscale.com",
--- a/hscontrol/types/config_test.go
+++ b/hscontrol/types/config_test.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
@@ -139,7 +140,7 @@ func TestReadConfig(t *testing.T) {
 				return LoadServerConfig()
 			},
 			want:    nil,
-			wantErr: "server_url cannot contain the base_domain, this will cause the headscale server and embedded DERP to become unreachable from the Tailscale node.",
+			wantErr: "server_url cannot be a suffix of the base_domain, this will cause the headscale server and embedded DERP to become unreachable from the Tailscale node.",
 		},
 		{
 			name:       "base-domain-not-in-server-url",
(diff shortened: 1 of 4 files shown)


hscontrol/types/config_test.go
```diff
@@ -333,3 +334,59 @@ tls_letsencrypt_challenge_type: TLS-ALPN-01
 	err = LoadConfig(tmpDir, false)
 	assert.NoError(t, err)
 }
+
+// OK
+// server_url: headscale.com, base: clients.headscale.com
+// server_url: headscale.com, base: headscale.net
```
quite, line 340: It is said elsewhere that server url must not be a suffix of the base_domain. To me, server is a suffix of base here -- yes, additionally they are equal. Perhaps the wording using "suffix" needs to be adjusted?

hscontrol/types/config_test.go
```diff
@@ -139,7 +140,7 @@ func TestReadConfig(t *testing.T) {
 				return LoadServerConfig()
 			},
 			want:    nil,
-			wantErr: "server_url cannot contain the base_domain, this will cause the headscale server and embedded DERP to become unreachable from the Tailscale node.",
+			wantErr: "server_url cannot be a suffix of the base_domain, this will cause the headscale server and embedded DERP to become unreachable from the Tailscale node.",
```
kradalby, line 143: this can be checked against `errServerURLSuffix.Error()`, then we dont have to update if we change the wording.

hscontrol/types/config_test.go
kradalby: you are write, it should probably be reformulated since we allow different domains, open to proposals.

hscontrol/types/config_test.go
```diff
@@ -333,3 +334,59 @@ tls_letsencrypt_challenge_type: TLS-ALPN-01
...
+// server_url: headscale.com, base: headscale.net
+//
+// NOT OK
+// server_url: server.headscale.com, base: headscale.com
+func TestSafeServerURL(t *testing.T) {
+	tests := []struct {
+		serverURL, baseDomain,
+		wantErr string
+	}{
+		{
```
kradalby, line 349: another case that should be allowed is if you completely ignore "real" TLDs:

serverURL: headscale.kristoffer.com
baseDomain: mybase

resulting in nodes having: `node1.mybase`, `server1.mybase`.

hscontrol/types/config_test.go
motiejus: See updated error message. I think the new one is more intuitive, see if you like it better

kradalby (approved):

motiejus:
@kradalby since we're here, I would appreciate if you could stamp my NixOS backport. :) https://github.com/NixOS/nixpkgs/pull/357969

Hypnotist1148:
I want to set my base_url to _example.com_ and the server_url to _entry.example.com_, I also have set split dns for entry.example.com to quad9's dns servers as to be still able to contact the headscale node.
Could you implement a overwrite so I am allowed to set this?

kradalby:
No, that is a foot gun we don't want to support or carry the issue burden for

Hypnotist1148:
Okay, what other options would there be, I don't really want the base domain to be a subdomain, and don't want to use a separate domain for the control plane

kradalby:
There are no other options, tailscale was designed in this way, upstream uses two domains, login.tailscale.com for the server and ts.net for basename.

PR #2261, opened 2024-11-29
restructure command/api to use stable IDs

### LINKED ISSUE #2246: [Bug] CLI should use Stable IDs instead of usernames
Currently a series of CLI commands are taking usernames as arguments for doing certain operations which is no longer stable as usernames will not be unique.

After https://github.com/juanfont/headscale/pull/2170, usernames that cannot be resolved to a single user will fail.

This works for now, but we need to make it possible to execute all of the commands for the specific ID being addressed.

kradalby:
fixes #2246


kradalby:
@nblock I did a quick doc search, and I dont think we refer to any of the changed parts of the `headscale users` command, but let me know if you see something I missed.

PR #2271, opened 2024-12-05
Add worker reading extra_records_path from file

### LINKED ISSUE #2262: [Feature] make it possible to read extra_records from file
### Use case

When `use_username_in_magic_dns` is removed, there will be a gap where DNS entries might not be available for some configuration.

A possible workaround is to generate these entries and put them in `extra_records` and generate them based on your headscale. See https://github.com/juanfont/headscale/issues/1369#issuecomment-2339901185 for more information.

### Description

To facilitate this use case (and others where you want to inject dns entries programmatically, we should add the ability to consume extra_records from a JSON file. 

For example, this will allow administrators to use `headscale` commands as part of a script to generate a JSON file by looping over the entries in the host list and create entries containing the username.

### Contribution

Let us create a new config option like `extra_records_path` or maybe paths, which sets up a file watcher reading the DNS entries on change. 
It should hash the file and check against the old hash to make sure we dont sent any updates if the file or content has not changed.

If the content has changed, propagate it to the nodes in the tailnet with our update mechanism, see if we can have a fastpath that only sends DNS updates.

kradalby:
Add worker reading extra_records_path from file.

Fixes #2262

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -12,8 +12,8 @@ on:
       # Stable release tags
       - v[0-9]+.[0-9]+.[0-9]+
     paths:
-      - 'docs/**'
-      - 'mkdocs.yml'
+      - "docs/**"
+      - "mkdocs.yml"
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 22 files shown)


kradalby:
Current status is that the filewatcher doesnt work, unsure if this is in the container env, or something else.

juanfont (approved):

PR #2280, opened 2024-12-11
feat: allow setting node registration expiration via config

### LINKED ISSUE #2277: [Feature] Configurable Registration Cache timeouts
### Use case

We have a use-case where we're remotely supporting a headscale installation.

We have users email us their registration key, which more often than not is outside the registration cache.

### Description

Create a config option to modify the node registration cache

### Contribution

https://github.com/juanfont/headscale/blob/main/hscontrol/app.go#L124 can reference values from config

chriswiggins:
I haven't done any unit / integration tests yet as I'd like to be pointed in the right direction with where it should be.

- [ ] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

Fixes #2277


hscontrol/types/config.go
```diff
@@ -935,6 +941,8 @@ func LoadServerConfig() (*Config, error) {
 			NodeMapSessionBufferedChanSize: viper.GetInt(
 				"tuning.node_mapsession_buffered_chan_size",
 			),
+			NodeRegistrationCacheExpiration: viper.GetDuration("tuning.node_registration_cache_expiration"),
+			NodeRegistrationCacheCleanup:    viper.GetDuration("tuning.node_registration_cache_cleanup"),
```
kradalby, line 945: I think these should be "nestable" in yaml, so 
```suggestion
			NodeRegistrationCacheExpiration: viper.GetDuration("tuning.node_registration_cache.expiration"),
			NodeRegistrationCacheCleanup:    viper.GetDuration("tuning.node_registration_cache.cleanup"),
```

but its fine to keep them as fields in go.

kradalby (commented):
This looks fine, I think this actually opens for us to test this, which it isnt atm.
Maybe like a variant of https://github.com/juanfont/headscale/blob/main/integration/auth_web_flow_test.go#L73, where the login flow has a wait to ensure they timeout so they are cleaned up and out of the cache and therefor fail. You can override the setting with a quite short expiry.

kradalby:
Happy for this to be picked up again, but in the meantime I will close it.

Issue #2301, opened 2024-12-16
[Feature] Enable CORS inside Headscale.


Jisse-Meruma:
### Feature Overview: Injecting `Access-Control-Allow-Origin` Response Header

This feature aims to inject the `Access-Control-Allow-Origin` response header directly through the configuration file (`config.yaml`). The primary purpose of this feature is to simplify the setup process for those using Headscale with a Web Tailscale Client.

[Learn more about the Tailscale SSH Console](https://tailscale.com/blog/ssh-console)

#### Justification

1. **Simplified Setup for Tailscale Web Client**: 
   -  Tailscale uses a browser client to set up an SSH console.
   - My company Leaning Technologies uses a Tailscale web client to establish peer-to-peer connections between two browsers.
   - We use Headscale for companies that prefer not to use the Tailscale Control Plane
   
2. **Eliminating the Proxy Layer**: 
   - **Before**: Users needed to deploy and configure an additional proxy server behind the Headscale server to manage Cross-Origin Resource Sharing (CORS).
   - **After**: This feature allows users to directly configure CORS support through the `config.yaml` file, eliminating the need for an extra proxy layer.

3. **Following the Same Principle**:
   - Headscale also supports integrated TLS, This is also to eliminate the use of a proxy server. This feature extends that Idea to also let you enable CORS

4. **Quality of Life Improvement**:
   - The ability to enable CORS through the configuration file simplifies deployment and configuration, making it more user-friendly and reducing setup complexity.

#### Configuration Details

The `Access-Control-Allow-Origin` header can be configured by adding the following to the `config.yaml` file:

```yaml
  # Examples:
  # "*" - Allow access from any origin (use with caution).
  # "http://example.com" - Allow access only from the specified origin.
  # "" - Disable CORS, no cross-origin requests allowed.
  Access-Control-Allow-Origin: "*" # Allows all origins 
```

### Feature Overview: Injecting `Access-Control-Allow-Origin` Response Header

This feature allows the injection of the `Access-Control-Allow-Origin` response header directly through the `config.yaml` file. The primary purpose is to simplify the setup process for users of Headscale with a Web Tailscale Client, making it easier to handle Cross-Origin Resource Sharing (CORS) without needing to set up a proxy server.

#### Configuration Details

The `Access-Control-Allow-Origin` header can be configured by adding the following to the `config.yaml` file:

### Contribution

- [X] I can write the design doc for this feature
- [X] I can contribute this feature

### How can it be implemented?

[...]

Issue #1283, opened 2023-03-24
make api support cors


xrain0610:
I want a configuration in yaml configuration file, it can allow cors params, like origin,method etc...

almereyda:
Is this maybe a duplicate of #623?

kradalby:
Closing as duplicate

Issue #623, opened 2022-06-09
Handle CORS headers and OPTIONS method for HTTP API


routerino:
**Bug description**

When trying to use a browser to generate API requests (like, hypothetically, if you're building a web frontend for headscale), the browser expects to to use CORS to determine if it can talk to the external server. The browser does this by the following:
* Sending a pre-flight `OPTIONS` request, expecting back a 204 response with the CORS headers attached
* Once the response is accepted, sending the real API request with all the data

For this to work, we need two things:
* The server (headscale) to generate CORS headers (and have it be configurable to set the domains appropriately)
* The server to accept `OPTIONS` requests without authorization.

**To Reproduce**
Generate a fetch request from a browser in a separate domain. Such as:

``` javascript
let apiKey = '<my api key>';
let url = 'https://<my headscale domain>/api/v1/machine';

window.fetch(url, {
    method: 'GET',
    headers: {
        Authorization: `Bearer ${apiKey}`
    }}).then((resp) => resp.json()).then(function (data) {console.log(data);}).catch(function (error) {console.log(error);});});
```

If no CORS headers are specified, you get this nice error in the browser console:
![image](https://user-images.githubusercontent.com/45954722/172862437-f64c47d2-8ae1-46b1-95ab-87123487c74e.png)

If you have the right headers (if you, for example, inject them with a reverse proxy) but the OPTIONS request is blocked by authorization, you get this nice error instead:
![image](https://user-images.githubusercontent.com/45954722/172862213-d8c060cf-eb58-4a4b-b486-e8dcf5942364.png)

Because the OPTIONS request is returning a 401 unauthorized when it shouldn't.

Both are not ideal. You can *fix* both with a reverse proxy, but you certainly shouldn't have to. The web server (gin?) should return OPTIONS with a 204 and be setting the CORS headers on all requests (and the CORS headers should be configurable).

**Context info**

These problems were fixed externally by routing through a Caddy reverse proxy using these matching settings:

```
@hs-options {
	host hs.<my-domain>
	method OPTIONS
}
@hs-other {
	host hs.<my-domain>
}
handle @hs-options {
	header {
		Access-Control-Allow-Origin https://<my-frontend-subdomain>
		Access-Control-Allow-Headers *
	}
	respond 204
}
handle @hs-other {
	reverse_proxy http://headscale:8080 {
		header_down Access-Control-Allow-Origin https://<my-frontend-subdomain>
		header_down Access-Control-Allow-Headers *
	}
}
```

[...]

PR #2302, opened 2024-12-16
Headscale: Added an option to set an Access-Control-Allow-Origin resp…

Jisse-Meruma:
add a header to enable Cross-Origin Resource Sharing (CORS) #2301 

- [X] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


http_Headscale_tutorial.md
```diff
@@ -0,0 +1,99 @@
+# Setting Up a Simple Headscale Control Server Using HTTP
```
kradalby, line 1: I dont think this is the place for this documentation, and we have our own documentation. This looks like internal notes.

config-example.yaml
```diff
@@ -40,6 +40,13 @@ grpc_listen_addr: 127.0.0.1:50443
 # are doing.
 grpc_allow_insecure: false
 
+# The Access-Control-Allow-Origin header specifies which origins are allowed to access resources.
+# Options:
+# - "*" to allow access from any origin (not recommended for sensitive data).
+# - "http://example.com" to only allow access from a specific origin.
+# - "" to disable Cross-Origin Resource Sharing (CORS).
+Access-Control-Allow-Origin: ""
```
kradalby, line 48: This does not follow the format of the config file and has not been discussed. This isnt really anything we need and people needing this can handle it in Nginx and the likes.

http_Headscale_tutorial.md
Jisse-Meruma: You are correct the has been reverted

config-example.yaml
Jisse-Meruma: Hi Kradalby,

I have modified the code to align with the requested configuration file format. 

I did open an accompanying issue for discussion: #2301. I also see mentions of a project chat, could you point me to it?

The idea of this feature is to simplify the Headscale server setup and have everything in one place. Tailscale continues to grow in the browser sector; for example, they recently released their new SSH client. Our company has just released WebVM 2.0, which is a virtual machine inside the browser. We use this product with Tailscale to establish fast peer-to-peer connections between machines. With this small change, the Headscale setup for the browser would be much easier and much simpler, especially in combination with your integrated TLS support. This feature has already been requested in two other issues but without success. Here are their references: #623 #1283 .

routerino: > This does not follow the format of the config file and has not been discussed. This isnt really anything we need and people needing this can handle it in Nginx and the likes.

FYI expecting people to rely on a reverse proxy to intercept requests and inject CORS is not normally acceptable behavior. I (as the maintainer of headscale ui) get people confused about this pretty much twice a month and opening issues as a result. It also results in people who aren't just relying on the browser to ignore CORS entirely (IE: when making calls via node instead), which lowers security within both applications.

A simple `UI_DOMAIN` configuration that optionally sets the appropriate CORS information would solve a lot of issues. I would not suggest allowing `*` anyway since browsers are planning to block that syntax.

Jisse-Meruma: > A simple `UI_DOMAIN` configuration that optionally sets the appropriate CORS information would solve a lot of issues. I would not suggest allowing `*` anyway since browsers are planning to block that syntax.

We do want to keep the flexibility of using the `*` available, but if indeed browsers are planning to block this syntax, I will make sure to update the feature to check for this syntax. To avoid any confusion, I would prefer to stick with the normal CORS syntax. This `*` is indeed another security risk, but this flexibility is also possible by using a reverse proxy. However, if Headscale agrees to not allow `*`, then I would gladly make another commit with a fix.

Jisse-Meruma:
Hi, I would like to know if there is a possibility to further discuss this feature?

kradalby (commented):
I posted some comments, I do not know enough about this topic other than how little I know and that I never do it correctly. I am uncomfortable to add something that has security implications that I do not have the knowledge to say if correct, or will be a problem/need change in the future.

This has been done a lot of times before, and it has currently led to a maintenance burden that is so large that I dont really have time to spend time on PRs like this where I need to figure out what is the correct way (which then will potentially have the recursive effect of having me do this again later).

My main thing is not to be negative, I am not saying no, I am just saying that all of this things needs thought, and I have a lot of other inflight things atm.

hscontrol/app.go
```diff
@@ -454,10 +454,21 @@ func (h *Headscale) ensureUnixSocketIsAbsent() error {
...
+		next.ServeHTTP(w, r)
+	})
+}
+
 func (h *Headscale) createRouter(grpcMux *grpcRuntime.ServeMux) *mux.Router {
 	router := mux.NewRouter()
 	router.Use(prometheusMiddleware)
 
+	if h.cfg.AccessControlAllowOrigins != "" {
+		router.Use(h.corsHeadersMiddleware)
```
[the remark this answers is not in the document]
Jisse-Meruma: Hi @kradalby,

Thank you for your feedback. I understand your concern about applying this to all endpoints.

I agree that not all endpoints require CORS headers. However, the Tailscale Client does use several non-API routes — specifically /register, /key, /machine, /derp, and /verify — to communicate with the Headscale server via HTTP requests. For these endpoints, implementing CORS headers is necessary to allow proper functionality when CORS is enabled. Additionally, the Tailscale SSH web client operates for example in a similar manner.

At the same time, I understand that certain endpoints such as /health, /apple, /windows, and /swagger do not need CORS headers.

I will give you a code draft of our idea on how to implement this in a simple matter.

If you have any questions feel free to ask me for any clarification if needed.

config-example.yaml
```diff
@@ -40,6 +40,13 @@ grpc_listen_addr: 127.0.0.1:50443
 # are doing.
 grpc_allow_insecure: false
 
+# The Access-Control-Allow-Origin header specifies which origins are allowed to access resources.
+# Options:
+# - "*" to allow access from any origin (not recommended for sensitive data).
```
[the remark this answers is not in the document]
Jisse-Meruma: Hi @kradalby,

Thank you for your input. I completely understand the concern about providing a wildcard `*` and the importance of specifying domains explicitly.

We can certainly disable the use of `*` for domain whitelisting and enforce explicit domain declarations. Our initial idea was to implement an "Allow list" where users can specify any allowed domain. 

As recommended by MDN: `If the server supports clients from multiple origins, it must return the origin for the specific client making the request.` Additionally, MDN advises adding the `Vary: Origin` header to inform clients that the response varies based on the `Origin` header.

Further information and recommendations can be found https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin.

---
I will show you a code draft to implement this feature soon. Feel free to give any feedback on the idea.

config-example.yaml
```diff
@@ -40,6 +40,13 @@ grpc_listen_addr: 127.0.0.1:50443
...
+# - "http://example.com" to only allow access from a specific origin.
+# - "" to disable Cross-Origin Resource Sharing (CORS).
+access_control_allow_origin: ""
```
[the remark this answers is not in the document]
Jisse-Meruma: Hi @kradalby,

Yes I suggest to have an structure like this:

```yaml
api:
    cors_allow_origins: [] # as a list
```
     
or

```yaml
cors:
     allow_origins: [] # as a list
```
---
I would value your input on which approach you prefer.

Jisse-Meruma:
Hi @kradalby,

I hope this message finds you well.

As per your recent comments, I have made some changes to the code. I'll wait once you have the time available to review the updated code.

Jisse-Meruma:
Hi @kradalby, Would it be possible to continue the discussion of this PR?

kradalby:
Sorry, I have been swamped, I've read up a bit this morning and this is my thoughts:

I think these headers should be limited to the api, that covers the main use case of allowing CORS for the different WebUIs. I do not think we should add it to all the other endpoints, that is quite an exotic use case and while I agree with @routerino that it is bad UX to ask people using the API and WebUI to setup a proxy to have CORS, I do think that this is exotic enough to say that you need to do it in a proxy.

I do not really want to support `*`, so setting a domain would be required. From what I understand from the spec, you can only have a single origin, not a list, so we should allow the list of a single domain (I might be wrong about this one, please enlighten me).
E.g. headscale server is: myheadscale.net and ui is: admin.myheadscale.net

Jisse-Meruma:
Hi @kradalby,

As of writing, the current commit no longer supports the use of `*` for the `Access-Control-Allow-Origin` header, as it could pose a security risk. You can now only set a specific origin. However, considering that multiple clients might connect to the server, I have implemented the flexibility to specify multiple origins.

The code loops through the list of allowed origins and checks if the request origin is in the list. If it is, the corresponding header is added to enable Cross-Origin Resource Sharing (CORS) for that specific origin. This technique is a standard approach when dealing with a list of Allowed origins, as documented here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin.

I understand your concerns about enabling CORS on sensitive routes and the potential security risks involved. Given that we no longer allow the use of `*` for the `Access-Control-Allow-Origin` header, administrators can now specify only particular URLs to permit connections to Headscale via the web.

Even if a rouge website was allowed to communicate with Headscale through the web, it would still require to go to the mandatory security layers to fully register a node. Additionally, Tailscale is beginning to implement web clients that can already connect to the official control server with CORS enabled.

To add an extra layer of security, we can consider allowing CORS on sensitive endpoints only for origins that use the HTTPS protocol. This ensures that the data is transmitted securely and reduces the risk associated with allowing CORS on less secure origins.

However, if this approach is not preferred, I am more than willing to make adjustments to enable CORS only for the WEBUI API.

Thank you for your understanding and please let me know if you have any further suggestions or concerns.
Best regards,
Jisse Meruma

Jisse-Meruma:
Hi @kradalby, I know you are quite busy. But still I would like to know if there is a change we could move this forward. I would like to know your opinion of my latest suggestion so that I can make a code draft for you.

scarytom:
I agree with @Jisse-Meruma that this is a sensible feature to add.

This PR seems to be ready, or very close to ready barring a final code review.  Is there anything we can do to help @kradalby?

kradalby:
We wont move forward with this for now, might revisit in the [API milestone](https://github.com/juanfont/headscale/milestone/15) (v0.30.0).

PR #2314, opened 2024-12-22
feat: Add PKCE Verifier for OIDC

Rorical:
From PR #1812 :

To fix the error "Could not exchange code for the token" when using the PKCE method, a verifier should be generated and used during the authentication process.

This change include change in configuration, oidc handling method and documents.

```yaml
oidc:
  # Optional: PKCE (Proof Key for Code Exchange) configuration
  # PKCE adds an additional layer of security to the OAuth 2.0 authorization code flow
  # by preventing authorization code interception attacks
  # See https://datatracker.ietf.org/doc/html/rfc7636
  pkce:
    # Enable or disable PKCE support (default: false)
    enabled: false
    # PKCE method to use:
    # - plain: Use plain code verifier
    # - S256: Use SHA256 hashed code verifier (default, recommended)
    method: S256
```

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -25,6 +25,7 @@ jobs:
           - TestOIDCAuthenticationPingAll
           - TestOIDCExpireNodesBasedOnTokenExpiry
           - TestOIDC024UserCreation
+          - TestOIDCAuthenticationWithPKCE
           - TestAuthWebFlowAuthenticationPingAll
           - TestAuthWebFlowLogoutAndRelogin
           - TestUserCommand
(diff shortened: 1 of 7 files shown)


kradalby (approved):
Great, I'm off the computer for a few days, could you push the draft as discussed and try to write up an issue for the proxy for later while it is fresh in memory?

kradalby:
Could you also add a changelog line?

Rorical:
Sure, draft is at #2315

Rorical:
Should this be done in v0.24.0 or Next?

Rorical:
> Should this be done in v0.24.0 or Next?

I mean where should I write this change in changelog

kradalby:
0.24 is good

PR #2315, opened 2024-12-22
add Tampering Test for PKCE

Rorical:
From PR #1812 , above #2314 .

This PR shows how we can mimic a MitM attack inside our integration tests and use that to test if the PKCE verifier is working. However, this code requires lots of change only for this specific test.

Ideally, we need a proxy that can mimic all type of MitM attack for all tests, so we can reuse the code and enhance integration test.


kradalby:
I'm going to close this to keep the PRs clean, I think we are in a good place with OIDC testing for now.

PR #2322, opened 2024-12-30
feat: Create headscale user and group as system user/groups

### LINKED ISSUE #2278: [Bug] Non-system headscale user and group from DEB packages

### Current Behavior

DEB packages from v0.23.0 release create non-system user and group with `/home/headscale` home dir and `/bin/sh` shell.

### Expected Behavior

Creating system user and system group with `/var/lib/headscale` home dir and `/usr/sbin/nologin` shell.
Similar to the way described in the[documentation](https://headscale.net/stable/setup/install/official/#using-standalone-binaries-advanced) for installation from official standalone binaries
```shell
sudo useradd \
 --create-home \
 --home-dir /var/lib/headscale/ \
 --system \
 --user-group \
 --shell /usr/sbin/nologin \
 headscale
```

### Steps To Reproduce

Install v0.23.0 using DEB package from the [GitHub releases page](https://github.com/juanfont/headscale/releases/tag/v0.23.0) on a system where no headscale user and group exists.

### Environment

```markdown
- OS: Ubuntu 22.04
- Headscale version: 0.23.0
- Tailscale version:
```

### Runtime environment

### Anything else?

I see that [postinstall.sh](https://github.com/juanfont/headscale/blob/main/docs/packaging/postinstall.sh) was updated in #2134 after release of v0.23.0 with fixes for home dir and shell, but maybe it still makes sense to create system user and system group?
E.g. by adding `-r` option for `useradd` and `groupadd` and adding `-S` option for Alpine's `addgroup`, like
```shell
useradd -r -s "$HEADSCALE_SHELL" -d "$HEADSCALE_HOME_DIR" -c "headscale default user" "$HEADSCALE_USER"
groupadd -r "$HEADSCALE_GROUP"
addgroup -S "$HEADSCALE_GROUP"
```

jyundt:
When creating the headscale user and group, create both as system groups rather than creating them as 'user' groups.

FIXES #2278

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/docs/packaging/postinstall.sh
+++ b/docs/packaging/postinstall.sh
@@ -31,13 +31,13 @@ ensure_headscale_path() {
 
 create_headscale_user() {
 	printf "PostInstall: Adding headscale user %s\n" "$HEADSCALE_USER"
-	useradd -s "$HEADSCALE_SHELL" -d "$HEADSCALE_HOME_DIR" -c "headscale default user" "$HEADSCALE_USER"
+	useradd -r -s "$HEADSCALE_SHELL" -d "$HEADSCALE_HOME_DIR" -c "headscale default user" "$HEADSCALE_USER"
 }
 
 create_headscale_group() {
 	if command -V systemctl >/dev/null 2>&1; then
 		printf "PostInstall: Adding headscale group %s\n" "$HEADSCALE_GROUP"
-		groupadd "$HEADSCALE_GROUP"
+		groupadd -r "$HEADSCALE_GROUP"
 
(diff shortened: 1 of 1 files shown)


kradalby:
How will this impact already created users?

chriswiggins:
If the install target is using systemd, would it not be beneficial to use DynamicUser?

The cockpit project does this now and it works really well: https://github.com/cockpit-project/cockpit/blob/main/src/systemd/cockpit-wsinstance-socket-user.service

You set a dependency on the user unit, and then the service uses that dynamically, using cgroups which means it is adequately sandboxed

nblock (approved):

PR #2338, opened 2025-01-10
use headscale server url as domain instead of base_domain

### LINKED ISSUE #2318: [Bug] headscale uses cfg.BaseDomain for building FQDNs

### Current Behavior

The config currently knows of:
* `cfg.BaseDomain`: not configurable through the config.yaml, set to `cfg.DNSConfig.BaseDomain`, used for `Domain` in `MapResponse`
* `cfg.DNSConfig.BaseDomain`: described as the magic DNS base domain.

`MapResponse.Domain` is used by GUIs to show the tailnet organization name (text below your username).
Currently, you can only configure the magic DNS base domain which will be used as the organization name. This is also used to build the FQDNs in host mappings.

### Expected Behavior

Be able to configure the organization name (organization domain) independently of the magic DNS base domain, as it is possible with tailscale.

### Steps To Reproduce

1. run tailscale with webclient=true and observe that the magic DNS (called suffix in `tailscale dns status`) base domain is used as the org name at the top.

### Environment

```markdown
- OS: n/a
- Headscale version: 0.23
- Tailscale version: 1.78.1 macos/windows
```

### Runtime environment

### Anything else?

I have a PR ready to fix this.

### LINKED ISSUE #2336: "Domain" to display in GUI applications
image

At the moment, the `base_domain` (`fap` in the example above) is used in the GUI application to describe which server the account belongs to, I think this should be changed to the domain part of the `server_url` as it is more descriptive to which server you are logging into, e.g.:

```
server_url: https://headscale.example.no
```

should result in the screenshot saying: 
kristoffer@myemail.com
headscale.example.no

kradalby:
Fixes #2336
Closes #2344
Closes #2318

### DIFF
--- a/hscontrol/mapper/mapper.go
+++ b/hscontrol/mapper/mapper.go
@@ -105,8 +105,7 @@ func generateUserProfiles(
 
 	var profiles []tailcfg.UserProfile
 	for _, user := range userMap {
-		profiles = append(profiles,
-			user.TailscaleUserProfile())
+		profiles = append(profiles, user.TailscaleUserProfile())
 	}
 
 	return profiles
@@ -455,7 +454,7 @@ func (m *Mapper) baseWithConfigMapResponse(
 
 	resp.DERPMap = m.derpMap
 
-	resp.Domain = m.cfg.BaseDomain
+	resp.Domain = m.cfg.Domain()
 
 	// Do not instruct clients to collect services we do not
 	// support or do anything with them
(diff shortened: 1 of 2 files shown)


kradalby:
I was just going to post on yours, I do not think it should be based on the base domain at all, so I do not think introducing another layer of config and falling back to the wrong one is the ideal setup. 

Since Headscale support one tailnet, I think using the name of the server makes more sense. the naming of base_name is already confusing enough and introducing a second one would not help I believe.

kradalby:
the problem isnt the fallback, its adding more configuration options, we are increasingly getting more strict as there are has previously been added two many nobbs making it hard for us to move forward without making anger. In this case we will continue to go for less configurability and defaults based on what we think are best practices.

Issue #2299, opened 2024-12-16
TestOIDC024UserCreation is flaky


kradalby:
TestOIDC024UserCreation can flake because user order is not guaranteed, example: https://github.com/juanfont/headscale/actions/runs/12342220536/job/34441801242?pr=2298

We should make the output from the program (sort db output?) or something similar stable output.

ljluestc:
```

package integration

import (
	"context"
	"sort"
	"testing"

	"github.com/juanfont/headscale/integration/hsic"
	"github.com/stretchr/testify/assert"
)

// TestOIDC024UserCreation tests the creation of users via OIDC.
func TestOIDC024UserCreation(t *testing.T) {
	// Setup the integration test context
	integration, err := hsic.NewTestingContext(t, nil)
	assert.NoError(t, err)
	defer integration.Close()

	// Simulate OIDC user creation
	// Replace with actual OIDC setup and user creation logic
	err = integration.CreateUser(context.Background(), "user1")
	assert.NoError(t, err)
	err = integration.CreateUser(context.Background(), "user2")
	assert.NoError(t, err)

	// Retrieve users from the database
	users, err := integration.ListUsers(context.Background())
	assert.NoError(t, err)

	// Sort users by Name to ensure deterministic order
	sort.Slice(users, func(i, j int) bool {
		return users[i].Name < users[j].Name
	})

	// Assert the users are as expected
	expectedUsers := []hsic.User{
		{Name: "user1"},
		{Name: "user2"},
	}
	assert.Equal(t, expectedUsers, users, "Users should match expected list in sorted order")
}

// ListUsers retrieves all users from the database.
// This is a placeholder; replace with actual database query logic.
func (i *hsic.TestingContext) ListUsers(ctx context.Context) ([]hsic.User, error) {
	var users []hsic.User
	// Replace with actual database query, e.g., using GORM or SQL
	// Example: i.DB.Find(&users)
	return users, nil
}

// CreateUser creates a user in the database.
// This is a placeholder; replace with actual OIDC user creation logic.
func (i *hsic.TestingContext) CreateUser(ctx context.Context, name string) error {
	// Replace with actual user creation logic, e.g., via OIDC flow
	return nil
}
```

kradalby:
Superseded by #3125

PR #2345, opened 2025-01-16
mark TestOIDC024UserCreation as flaky

kradalby:
Updates https://github.com/juanfont/headscale/issues/2299


kradalby:
Not working, tailscale stuff has a regex for their repos

PR #2350, opened 2025-01-17
set oidc.map_legacy_users false

kradalby:
set `oidc.map_legacy_users` to false as per deprecation plan from 0.24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Next
 
+### Changes
+
+- `oidc.map_legacy_users` is now `false` by default
+  [#2350](https://github.com/juanfont/headscale/pull/2350)
+
 ## 0.24.0 (2025-01-17)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Next
 
+### Changes
+
+- `oidc.map_legacy_users` is not `false` by default
```
nblock, line 7: is now

config-example.yaml
```diff
@@ -387,7 +387,7 @@ unix_socket_permission: "0770"
 #   # Disabling this feature will cause all new logins to be created as new users.
 #   # Note this option will be removed in the future and should be set to false
 #   # on all new installations, or when all users have logged in with OIDC once.
```
nblock, line 389: Maybe this paragraph should be rewritten to reflect the new default.

nblock (changes requested):

nblock (approved):

PR #2380, opened 2025-01-26
Spell check and updating docs

Nbelles:
- [ ] added unit tests
- [~] added integration tests
- [X] updated CHANGELOG.md

I noticed a slight inconsistency in the public docs (the `/windows` and `/apple` pages were referenced but not `/android`) and I figured while I was making this PR that I might as well do a quick spell check across the codebase. Most everything was updating spelling in comments. 

There is one actual code change in an integration test (`/integration/cli_test.go`) that had a string that had a typo in it. That is probably the only thing that could use a pair of eyes from an actual codebase developer to make sure that doesn't break anything. 

Thanks!

### DIFF
--- a/docs/about/clients.md
+++ b/docs/about/clients.md
@@ -10,7 +10,7 @@ headscale.
 | OpenBSD | Yes                                                                                                   |
 | FreeBSD | Yes                                                                                                   |
 | Windows | Yes (see [docs](../usage/connect/windows.md) and `/windows` on your headscale for more information)   |
-| Android | Yes (see [docs](../usage/connect/android.md))                                                         |
+| Android | Yes (see [docs](../usage/connect/android.md) and `/android` on your headscale for more information)   |
 | macOS   | Yes (see [docs](../usage/connect/apple.md#macos) and `/apple` on your headscale for more information) |
 | iOS     | Yes (see [docs](../usage/connect/apple.md#ios) and `/apple` on your headscale for more information)   |
 | tvOS    | Yes (see [docs](../usage/connect/apple.md#tvos) and `/apple` on your headscale for more information)  |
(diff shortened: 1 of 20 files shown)


docs/about/clients.md
```diff
@@ -10,7 +10,7 @@ headscale.
 | OpenBSD | Yes                                                                                                   |
 | FreeBSD | Yes                                                                                                   |
 | Windows | Yes (see [docs](../usage/connect/windows.md) and `/windows` on your headscale for more information)   |
-| Android | Yes (see [docs](../usage/connect/android.md))                                                         |
+| Android | Yes (see [docs](../usage/connect/android.md) and `/android` on your headscale for more information)   |
```
nblock, line 13: There is no `/android` endpoint, please revert this change.

nblock (changes requested):

docs/about/clients.md
Nbelles: Sorry about that, I saw thought the `/android` endpoint was auto-generated using the same `../usage/connect/<os>.md` naming as the other OS's. When I tried visiting `/android` on my instance it didn't give me any errors so I assumed it was just a bug that it didn't actually load the content of the `android.md` file properly. 

Reverted that change.

integration/cli_test.go
```diff
@@ -1840,7 +1840,7 @@ func TestPolicyBrokenConfigCommand(t *testing.T) {
 			{
 				// This is an unknown action, so it will return an error
 				// and the config will not be applied.
-				Action:       "acccept",
+				Action:       "accept",
```
nblock, line 1843: The comment above explicitly tests for an unknown action, please revert this change.

nblock (changes requested):

integration/cli_test.go
Nbelles: Would it be reasonable to change this for something that isn't explicitly a typo? Maybe something that makes more logical sense for it to not be valid while still passing a spell check search?

nblock: Sounds reasonable, yes.

Nbelles: Just updated to `"unknown-action"`, let me know if that doesn't work and I can change it to something else.

nblock (approved):

nblock:
Thanks for incorporating the feedback. There is a conflict now, please rebase and force-push. Then, we're good to go.

PR #2383, opened 2025-01-27
stricter hostname validation and replace

kradalby:


nblock (approved):
A entry in the "breaking" section of the changelog would be good as some hostnames that worked before will now be rejected/renamed.

PR #2420, opened 2025-02-09
use tsweb debugger

kradalby:
This PR switches the homegrown debug endpoint to using tsweb.Debugger, a neat toolkit with batteries included for pprof and friends, and making it easy to add additional debug info:

I've started out by adding a bunch of "introspect" endpoints
image

So users can see the acl, filter, config, derpmap and connected nodes as headscale sees them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,15 @@
 
 ## Next
 
-
 ### Changes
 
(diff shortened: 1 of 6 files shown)


kradalby:
Also statsviz:
image

juanfont (approved):

Issue #2256, opened 2024-11-25
[Feature] Adding an option to prevent tailscale clients adding `~.` to its DNS search domains for systemd-resolved.


AlynxZhou:
### Use case

systemd-resolved works as a multicast DNS server, which means by default when it queries domains, it will try all available providers, and use the fastest result, unless the domain matches one of DNS search domains of one provider, then it will only use this provider to query.

Tailscale by default registers itself as one of providers of systemd-resolved, but it adds `~.` to its DNS search domains, this means all domains will match, and makes systemd-resolved's multicast invalid, all DNS queries go to tailscale's DNS. User may only wants to query tailscale domains via tailscale.

### Description

Setting `tailcfg.DNSConfig.FallbackResolvers` instead of `tailcfg.DNSConfig.Resolvers` can prevent tailscale clients from adding `~.`. In previous version this behavior is controlled by `override_local_dns`, but this option is removed because it does not work as intended, and then we lose a way to control this behavior.

Is there any chance to add an option to control setting `tailcfg.DNSConfig.FallbackResolvers` instead of `tailcfg.DNSConfig.Resolvers`?

### Contribution

- [ ] I can write the design doc for this feature
- [X] I can contribute this feature

### How can it be implemented?

Just bring back <https://github.com/juanfont/headscale/pull/905/files#diff-0e426a43248661127a0c0ee115aef7a1093b635f8993b3f7ebb1dd9f05b8f249R406-R410>, but with a better option to describe what it does?

nblock:
The Tailscale SaaS behaves as follows when changing the setting "Override local DNS":

When enabled in the control plane, Tailscale configures systemd-resolved as default route (`+DefaultRoute`) and also
configures a route-only domain `~.` so all domains will match and Tailscale will receive all DNS requests.

```shell
$ resolvectl status tailscale0
Link 4 (tailscale0)
    Current Scopes: DNS
         Protocols: +DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 100.100.100.100
       DNS Servers: 100.100.100.100
        DNS Domain: taile8db11.ts.net ~.
     Default Route: yes
```

When disabled in the control plane, Tailscale will not set the default route (`-DefaultRoute`) and the route-only domain
is not present in the list of DNS domains. It will only handle requests that match the list of domains.

```shell
$ resolvectl status tailscale0
Link 4 (tailscale0)
    Current Scopes: DNS
[...]

PR #2438, opened 2025-02-20
Restore support for "Override local DNS"

nblock:
Tailscale allows to override the local DNS settings of a node via "Override local DNS" [1]. Restore this flag with the same config setting name `dns.override_local_dns` but disable it by default to align it with Tailscale's default behaviour.

Tested with Tailscale 1.80.2 and systemd-resolved on Debian 12.

With `dns.override_local_dns: false`:

```
Link 12 (tailscale0)
Current Scopes: DNS
     Protocols: -DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
   DNS Servers: 100.100.100.100
    DNS Domain: tn.example.com ~0.e.1.a.c.5.1.1.a.7.d.f.ip6.arpa [snip]
```

With `dns.override_local_dns: true`:

```
Link 12 (tailscale0)
Current Scopes: DNS
     Protocols: +DefaultRoute -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
   DNS Servers: 100.100.100.100
    DNS Domain: tn.example.com ~.
```

[1] https://tailscale.com/kb/1054/dns#override-local-dns

Fixes: #2256

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


kradalby (approved):

hscontrol/types/config.go
```diff
@@ -287,6 +288,7 @@ func LoadConfig(path string, isFile bool) error {
 
 	viper.SetDefault("dns.magic_dns", true)
 	viper.SetDefault("dns.base_domain", "")
+	viper.SetDefault("dns.override_local_dns", true)
```
SuperSandro2000, line 291: I think https://github.com/juanfont/headscale/pull/2438/files#r2046953908 is correct. I also just ran into this and needed to turn local dns override off.

fallingcats:
This being on by default now immediately breaks old configs that didn't specify the option and don't have any global nameserver set. Was that intentional?

nblock:
> This being on by default now immediately breaks old configs that didn't specify the option and don't have any global nameserver set. Was that intentional?

This was enabled (and enforced) for 0.23.x, 0.24.x and 0.25.x and an upgrade without adjusting the configuration will still have this option enabled. But using "Override local DNS" without specifying global nameservers is an error which is now checked for.

lgrn:
> This being on by default now immediately breaks old configs that didn't specify the option and don't have any global nameserver set. Was that intentional?

For searchability, I'd like to point out that despite our config **having** global nameservers set server-side, this broke mobile clients on both iOS and Android until `override_local_dns: true` was added (I assume it's false by default when the line is missing). The errors visible were (selected for brevity):

* `You are logged out. The last login error was: fetch control key`
* `failed to resolve`
* `no DNS fallback candidates remain`

z3ntu:
To add my 2 cents, I think this has also broken the config for me, on Linux I couldn't resolve the hostnames anymore that are set up using `extra_records`. I believe for some reason the Android client still was able to resolve it, but maybe that was just a DNS cache, I didn't check much more.
With the headscale config updated to `override_local_dns: true` and restarting everything appears to make everything work again as expected.

PR #2461, opened 2025-03-04
:memo: Update documentation to support i18n, add Chinese.

afterow:

:memo: Updated documentation to support i18n, and added Chinese.


kradalby:
I am sorry, but I am not comfortable with having a lot of content that I cannot validate and that might drive support up or contain information we cannot control is not an option.

Issue #2333, opened 2025-01-08
[Bug] OIDC users don't always get a username


ToxicMushroom:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When logging in for the first time with oidc I only got a display_name and no username/name set in headscale.

This can be worked around currently by `docker exec headscale headscale users rename --identifier 1 -r merlijn` 
But that'd be inconvenient for future users so it should probably get resolved.

### Expected Behavior

All the required user fields get set when logging in with oidc,
 - either by requiring certain claims are present in the idToken.
 - or by making the cli commands more flexible: by accepting user ids as an alternative.

### Steps To Reproduce

1. Setup OIDC (can see my config below)
2. Authenticate new user (My oidc resource server is ran on Kanidm, returned claims are also below)
3. List the user's fields (also below)

### Environment

```markdown
- OS: Debian 12
- Headscale version: ghcr.io/juanfont/headscale:0.24.0-beta.2-debug
- Tailscale version: 1.78.1
```

### Runtime environment

- [X] Headscale is behind a (reverse) proxy
- [X] Headscale runs in a container

### Anything else?

`root@zungenbrecher:/opt/vpn_exit_node# docker exec headscale headscale users list -o json`
```json
[
    {
        "id": 1,
        "created_at": {
            "seconds": 1735582411,
            "nanos": 102993003
        },
        "display_name": "Merlijn",
        "provider_id": "oauth2urlthing",
        "provider": "oidc"
    }
]
```

<details>
  <summary>oidc claims for my user with "profile", "email", "groups" scopes</summary>

```json
claims: {
    "sub": "2fa57e05-fcc9-43db-8ca6-a98602346aa5",
    "exp": 1736341021,
    "name": "Merlijn",
    "preferred_username": "merlijn@idm.melijn.com",
    "groups": [
        "idm_all_persons@idm.melijn.com",
        "00000000-0000-0000-0000-000000000035",
        "idm_all_accounts@idm.melijn.com",
        "00000000-0000-0000-0000-000000000036",
        "poweruser@idm.melijn.com",
        "91f2f258-3122-4644-a860-cbfe2cce5b73",
        "seafile_users@idm.melijn.com",
        "1163f5cc-6dee-46c8-85db-fd635247c9bd",
        "friends@idm.melijn.com",
        "553b933b-0f01-432d-88a8-d69b182c4cf6",
[...]

Issue #2295, opened 2024-12-15
[Feature] [Docs] add OIDC documentation for v 24.0


masterwishx:
### Use case

updated to 24.0beta1 all fine. 
should we change in config for OIDC ?  
`scope: ["openid", "profile", "email"]`  
to add `preferred_username,name,picture` ? using Authentik.
We should update the documentation for this. 

### Description

Update documentation for OIDC in v 24.0

### Contribution

- [ ] I can write the design doc for this feature
- [ ] I can contribute this feature

### How can it be implemented?

Update documentation I can give my example of current using Authentik

nblock:
I plan to update the current OIDC documentation and provide more structure. It should probably be structured like this:

* Short introduction, possibly with a list of known limitations
* Generic setup instructions: the OIDC configuration of Headscale should be described once with a description of required parameters (`issuer`, `client_id`, `client_secret` or `client_secret_path`) and some useful optional parameters such as `scope`.
* IdP specific instructions where we could rehash parts of the existing configuration.

Anything missing that should be covered?

@masterwishx it'd be great if you could share your Authentik configuration, thx.

PR #2493, opened 2025-03-20
OIDC: Fetch UserInfo to get EmailVerified if necessary

benley:
Some OIDC providers (including Okta) do not include an `email_verified` claim in ID tokens, so headscale will never trust their email addresses.  The OIDC spec says that the set of ["standard claims"](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) can be returned in either the UserInfo response or the ID token, so we can try requesting UserInfo if necessary to find the `email_verified` claim that we're looking for.  If this also fails, the behavior of headscale is unchanged from before.

Please let me know if this approach looks acceptable, and I can try to add test coverage and update docs as needed.

Related to #2333  and #2295 

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


hscontrol/oidc.go
```diff
@@ -273,6 +280,16 @@ func (a *AuthProviderOIDC) OIDCCallbackHandler(
 		return
 	}
 
+	// If EmailVerified is missing, we can try to get it from UserInfo
+	if !claims.EmailVerified {
+		var userinfo *oidc.UserInfo
+		userinfo, err = a.oidcProvider.UserInfo(req.Context(), oauth2.StaticTokenSource(oauth2Token))
+		if err != nil {
```
tfly-singlestore, line 287: not sure what is consuming this upstream - do we need to be returning httpError for consuming function?

tfly-singlestore (approved):

hscontrol/oidc.go
benley: This error is one we can ignore and continue, because we can just treat EmailVerified as false if we're not able to fetch the UserInfo endpoint.  I'm avoiding the assumption that every OIDC provider will cooperate here, since headscale currently doesn't require UserInfo at all.

hscontrol/oidc.go
```diff
@@ -356,7 +372,14 @@ func (a *AuthProviderOIDC) extractIDToken(
...
+}
 
+// extractIDToken extracts the ID token from the oauth2 token.
+func (a *AuthProviderOIDC) extractIDToken(
+	ctx context.Context,
+	oauth2Token *oauth2.Token,
+) (*oidc.IDToken, error) {
 	rawIDToken, ok := oauth2Token.Extra("id_token").(string)
 	if !ok {
 		return nil, NewHTTPError(http.StatusBadRequest, "no id_token", errNoOIDCIDToken)
```
dhyanio, line 385: It could panic if the `Extra` method does not return a string.
Suggestion: 
```go
rawIDToken, ok := oauth2Token.Extra("id_token").(string)
if !ok || rawIDToken == "" {
    return nil, NewHTTPError(http.StatusBadRequest, "missing or invalid id_token", errNoOIDCIDToken)
}
```

kradalby: I dont think this is correct, the ok check of the type cast will cover this.

hscontrol/oidc.go
kradalby: Yes, it doesnt sound like we want to error on this, I assume that in the "liberally" spec that is oidc, a provider does not need to implement this and it might error for no reason.

kradalby (approved):
I think this looks reasonable, is it confirmed that some of the providers (like Okta), sets the email verified in the user info?

Question for further improvement, is there more info we potentially should "fall back" to getting from userinfo if we dont get it in the claim?

hscontrol/oidc.go
dhyanio: yeh, right :)

benley:
No rush, but I think this is ready for another look.

alteriks:
Hi, I've created https://github.com/juanfont/headscale/pull/2505 with docs regarding how to configure Authelia. Since [v4.39.0](https://github.com/authelia/authelia/releases/tag/v4.39.1), by default Authelia removed 'default' claims from ID Token. If headscale starts using UserInfo endpoint we won't have to send those claims to Headscale. 
Ping me, if you'd like me to verify your code with Authelia.

kradalby:
@benley looks good, will merge when the TestUpdateHostnameFromClient passes, mostly a formality.

Would you be up for a PR to use this info for the other fields too?

I imagine we prioritise the claims, and then fall back to userinfo.

benley:
> Would you be up for a PR to use this info for the other fields too?
> 
> I imagine we prioritise the claims, and then fall back to userinfo.

Sure, I can followup with a second PR to do the rest of the fields. Thank you for merging this one!

PR #2509, opened 2025-03-28
Only read relevant nodes from database in PeerChangedResponse

Enkelmann:
Querying the whole database becomes a performance issue if a lot of nodes exist in it, even if all those nodes are offline. In `PeerChangedResponse` we actually know which nodes need to be queried, so we can restrict the database queries to these nodes.

In my tests, `PeerChangedResponse` usually needed to query less than 5 nodes, so the change resulted in an order of magnitude better performance for a database with several hundred nodes (most of them offline).

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

I can also write a test if desired. The PR should not change any behavior, so I wanted to base tests on existing tests for `PeerChangedResponse`. However, I did not find any existing tests for this function, so I am happy for pointers where to start for testing `PeerChangedResponse`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,7 +87,7 @@ The new policy can be used by setting the environment variable
   [#2493](https://github.com/juanfont/headscale/pull/2493)
   - If a OIDC provider doesn't include the `email_verified` claim in its ID
     tokens, Headscale will attempt to get it from the UserInfo endpoint.
-- Improve performance by only querying relevant nodes from the database for node updates
+- Improve performance by only querying relevant nodes from the database for node updates [#2509](https://github.com/juanfont/headscale/pull/2509)
 
 ## 0.25.1 (2025-02-25)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@ The new policy can be used by setting the environment variable
   [#2493](https://github.com/juanfont/headscale/pull/2493)
   - If a OIDC provider doesn't include the `email_verified` claim in its ID
     tokens, Headscale will attempt to get it from the UserInfo endpoint.
+- Improve performance by only querying relevant nodes from the database for node updates
 
 ## 0.25.1 (2025-02-25)
 
--- a/hscontrol/db/node.go
+++ b/hscontrol/db/node.go
@@ -58,26 +58,23 @@ func ListPeers(tx *gorm.DB, nodeID types.NodeID) (types.Nodes, error) {
 	return nodes, nil
 }
 
-func (hsdb *HSDatabase) ListNodes(nodeIDs ...types.NodeIDs) (types.Nodes, error) {
+// ListNodes queries the database for either all nodes if no parameters are given
+// or for the given nodes if at least one node ID is given as parameter
+func (hsdb *HSDatabase) ListNodes(nodeIDs ...types.NodeID) (types.Nodes, error) {
 	return Read(hsdb.DB, func(rx *gorm.DB) (types.Nodes, error) {
 		return ListNodes(rx, nodeIDs...)
 	})
 }
 
-func ListNodes(tx *gorm.DB, nodeIDs ...types.NodeIDs) (types.Nodes, error) {
-	if len(nodeIDs) > 0 && len(nodeIDs[0]) == 0 {
-		return types.Nodes{}, nil
-	}
-	var nodeFilter types.NodeIDs = nil
-	if len(nodeIDs) > 0 {
-		nodeFilter = nodeIDs[0]
-	}
+// ListNodes queries the database for either all nodes if no parameters are given
+// or for the given nodes if at least one node ID is given as parameter
+func ListNodes(tx *gorm.DB, nodeIDs ...types.NodeID) (types.Nodes, error) {
 	nodes := types.Nodes{}
 	if err := tx.
 		Preload("AuthKey").
 		Preload("AuthKey.User").
 		Preload("User").
-		Where(nodeFilter).Find(&nodes).Error; err != nil {
+		Where(nodeIDs).Find(&nodes).Error; err != nil {
 		return nil, err
 	}
 
--- a/hscontrol/db/node.go
+++ b/hscontrol/db/node.go
@@ -35,21 +35,26 @@ var (
 	)
 )
 
-func (hsdb *HSDatabase) ListPeers(nodeID types.NodeID) (types.Nodes, error) {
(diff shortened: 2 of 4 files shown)


hscontrol/mapper/mapper.go
```diff
@@ -257,11 +259,6 @@ func (m *Mapper) PeerChangedResponse(
 ) ([]byte, error) {
 	resp := m.baseMapResponse()
 
-	peers, err := m.ListPeers(node.ID)
```
kradalby, line 260: I agree this is not necessary, but instead of listing all, we should just list the one from the change map.

It would make more sense to make `ListPeers(node NodeID, peerIDs ...NodeID)` and only fetch the ones that are needed here.

hscontrol/mapper/mapper.go
```diff
@@ -273,13 +270,21 @@ func (m *Mapper) PeerChangedResponse(
 	}
 
 	changedNodes := make(types.Nodes, 0, len(changedIDs))
-	for _, peer := range peers {
-		if slices.Contains(changedIDs, peer.ID) {
-			changedNodes = append(changedNodes, peer)
+	for _, changedID := range changedIDs {
+		if changedID != node.ID {
+			changedNode, err := m.GetNode(changedID)
```
kradalby, line 275: The problem with doing it this way, and not as mentioned above is that we dont get transactional consistency. In this case, nodes might change in between calls to GetNode. While this might not be an obvious problem, it might lead to incredibly hard to debug problems.

kradalby (changes requested):
I think the general change makes sense, in general a change request is (in my observation) not more than a subset of nodes, like you call out. 

I am up for this, but it needs to be fetched as ones.

hscontrol/mapper/mapper.go
Enkelmann: I have reworked the code to use exactly one database query. I will also add a test to the database query function at least.

hscontrol/mapper/mapper.go
Enkelmann: I opted against changing `ListPeers` for now. It is also used for full map responses, where I think all peers are needed (although I have not looked too deeply into this function yet). But I can of course change the naming scheme if you like something like `ListPeersSubset` better than `ListNodesSubset`.

Enkelmann:
The code is reworked to use exactly one database query and I also added a unit test. I have some concern that my unit test does not cover the case of a NodeID being zero, although the code to add nodes in the test seem to suggest that this case cannot happen anyway.

hscontrol/db/node.go
```diff
@@ -77,6 +77,28 @@ func ListNodes(tx *gorm.DB) (types.Nodes, error) {
 	return nodes, nil
 }
 
+func (hsdb *HSDatabase) ListNodesSubset(nodeIDs types.NodeIDs) (types.Nodes, error) {
```
kradalby, line 80:
```suggestion
func (hsdb *HSDatabase) ListNodes(nodeIDs ...types.NodeID) (types.Nodes, error) {
```

I would prefer to just have the same function for this, but use the list expansion argument like this so it remains optional for getting all.

hscontrol/mapper/mapper.go
kradalby: I think modifying both `ListNodes` and `ListPeers` to be able to take a optional list arg would be the best.

kradalby (changes requested):
Much better, couple of comments to simplify and keep around lest functions/code.

CHANGELOG.md
```diff
@@ -87,6 +87,7 @@ The new policy can be used by setting the environment variable
   [#2493](https://github.com/juanfont/headscale/pull/2493)
   - If a OIDC provider doesn't include the `email_verified` claim in its ID
     tokens, Headscale will attempt to get it from the UserInfo endpoint.
+- Improve performance by only querying relevant nodes from the database for node updates
```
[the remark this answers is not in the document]
Enkelmann: Done

hscontrol/db/node.go
Enkelmann: Is implemented.

hscontrol/mapper/mapper.go
```diff
@@ -496,6 +491,20 @@ func (m *Mapper) ListPeers(nodeID types.NodeID) (types.Nodes, error) {
 	return peers, nil
 }
 
+func (m *Mapper) ListNodes(nodeIDs ...types.NodeIDs) (types.Nodes, error) {
```
kradalby, line 494: You need to pass it as types.NodeID (without the s), now you are passing it as a list of lists

Enkelmann: That is by design. When using an optional argument we face the problem of having two different types of empty lists, both of which are needed:
- The empty list of not using the optional argument. We want to return all nodes in this case.
- The empty list of using the optional argument, but with a list where all nodes have been filtered out. We want to return an empty list here and *not all nodes*.

One could use the `nil` and `[]type.NodeID{}` variants of empty lists to distinguish these two cases. But in my experience, code depending on different behavior between these two empty list variants is a future bug waiting to happen. I know that my design also has a hidden footgun in that one could pass more than one list of nodes as parameter to the function. But I would prefer having two `ListNodes` functions before changing the parameter type to `...types.NodeID`.

Side remark: I will wait until we have settled on a design for `ListNodes` before I take a second look at `ListPeers`.

kradalby: > The empty list of using the optional argument, but with a list where all nodes have been filtered out. We want to return an empty list here and not all nodes.

This scenario does not make any sense to me, in the case where all nodes has been filtered out, you would have an empty list of IDs and you would instead not call for the database and return early.

kradalby: I think it will be fine with a docstring explaining the filter. Current behaviour is not that ergonomic and I would want it to be.

Enkelmann:
Ok, `ListNodes` now returns all nodes when no parameter is given and the behavior is documented in a docstring. If you like this version, I can take a look at `ListPeers`.

hscontrol/mapper/mapper.go
Enkelmann: Although I do not like it from a coding-style perspective, I implemented it now.

kradalby:
Looks great, thank you!

Enkelmann:
I added the same functionality to ListPeers (currently not used anywhere).
As far as I can tell the integration test failures have nothing to do with the PR, so I have not done anything about them.

PR #2515, opened 2025-04-02
Add unraid-headscale-admin web UI to docs

ich777:
This PR adds unraid-headscale-admin to the docs.

unraid-headscale-admin is a simple Plugin for Unraid which adds a simple WebUI for Headscale directly into Unraid which then can communicate in Local Mode (when running the official Headscale Docker container directly on the same machine through `docker exec`) or API Mode (I think the API Mode is self explanatory) with Headscale.


nblock (changes requested):
Thx, could you please format the markdown table?

ich777:
> Thx, could you please format the markdown table?

Done, wasn't sure if you want proper formatting in the source too. :)

nblock (approved):

PR #2544, opened 2025-04-23
config: disallow same server url and base_domain

### LINKED ISSUE #2354: [Bug] DNS Resolve Error for the base_domain(server_rul) at v0.24.0

### Current Behavior

1. `config.yaml`
```yaml
server_url: example.net
dns:
  magic_dns: true
  base_domain: example.net
``` 

2. try to `ping example.net`
```bash
➜  ~ ping example.net
ping: cannot resolve example.net: Unknown host
```

3. or try to open `https://example.net/apple`

Image

4. magic_dns works
```bash
➜  ~ ping test              
PING test.example.net (100.64.0.7): 56 data bytes
64 bytes from 100.64.0.7: icmp_seq=0 ttl=64 time=253.130 ms
64 bytes from 100.64.0.7: icmp_seq=1 ttl=64 time=86.409 ms
^C
--- test.example.net ping statistics ---
2 packets transmitted, 2 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 86.409/169.769/253.130/83.361 ms
➜  ~ ping test.example.net    
PING test.example.net (100.64.0.7): 56 data bytes
64 bytes from 100.64.0.7: icmp_seq=0 ttl=64 time=94.611 ms
64 bytes from 100.64.0.7: icmp_seq=1 ttl=64 time=103.333 ms
^C
--- test.example.net ping statistics ---
2 packets transmitted, 2 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 94.611/98.972/103.333/4.361 ms
```

### Expected Behavior

it should resolve `base_domain`?, i mean it should bypass the DNS querying

### Steps To Reproduce

see up

### Environment

```markdown
- OS: MacOS
- Headscale version: 0.24.0
- Tailscale version: 1.78.1
```

### Runtime environment

### Anything else?

_No response_

kradalby:
Remove a footgun.

Fixes #2354

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,11 @@ towards this code.
 The new policy can be used by setting the environment variable
 `HEADSCALE_EXPERIMENTAL_POLICY_V2` to `1`.
 
+#### Other breaking
+
+- Disallow `server_url` and `base_domain` to be equal
+  [#2544](https://github.com/juanfont/headscale/pull/2544)
+
 ### Changes
 
 - Use Go 1.24 [#2427](https://github.com/juanfont/headscale/pull/2427)
--- a/hscontrol/types/config.go
+++ b/hscontrol/types/config.go
@@ -33,6 +33,7 @@ const (
 var (
 	errOidcMutuallyExclusive = errors.New("oidc_client_secret and oidc_client_secret_path are mutually exclusive")
 	errServerURLSuffix       = errors.New("server_url cannot be part of base_domain in a way that could make the DERP and headscale server unreachable")
+	errServerURLSame         = errors.New("server_url cannot use the same domain as base_domain in a way that could make the DERP and headscale server unreachable")
 	errInvalidPKCEMethod     = errors.New("pkce.method must be either 'plain' or 'S256'")
 )
 
@@ -997,6 +998,10 @@ func isSafeServerURL(serverURL, baseDomain string) error {
 	server, err := url.Parse(serverURL)
 	if err != nil {
 		return err
+	}
+
+	if server.Hostname() == baseDomain {
+		return errServerURLSame
 	}
 
 	serverDomainParts := strings.Split(server.Host, ".")
--- a/hscontrol/types/config_test.go
+++ b/hscontrol/types/config_test.go
@@ -423,6 +423,7 @@ func TestSafeServerURL(t *testing.T) {
 		{
 			serverURL:  "https://headscale.com",
 			baseDomain: "headscale.com",
+			wantErr:    errServerURLSame.Error(),
 		},
 		{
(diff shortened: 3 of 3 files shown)


motiejus (approved):

nblock (approved):

IamTaoChen:
I have some questions regarding the relationship between `base_domain` and `server_url`.

I initially set them to the same value because I want all devices to appear under a specific subdomain pattern like `*.example.net`. I have a dedicated top-level domain reserved just for Headscale usage.

However, I’m a bit confused after reading this PR (#2544), especially in comparison to PR #2338.

lukaslindnermusic:
> I have some questions regarding the relationship between `base_domain` and `server_url`.
> 
> I initially set them to the same value because I want all devices to appear under a specific subdomain pattern like `*.example.net`. I have a dedicated top-level domain reserved just for Headscale usage.
> 
> However, I’m a bit confused after reading this PR (#2544), especially in comparison to PR #2338.

Hey there 😊 
Yeah, I am a little bit confused about this, too. Why can't I have my tailscale instance run on i.e. `vpn.domain.tld` and then my magicdns uses `device1.vpn.domain.tld`?
What's the reason for this change?

Wh1t3Fox:
@lukaslindnermusic just set your `server_url` to `https://vpn.domain.tld:443` and you'll be good

xcpn:
I also don't understand this change, this messed up some other stuff on my server.

I have two services on my Debian server, `headscale.example.com` and `ntfy.example.com`.

If I change `base_domain` to `example.com` and have `tailscale set --accept-dns=true` (the Debian server is also a node configured as exit-node), I can't even send notifications via NTFY from the server.

I now need to disable `--accept-dns` and re-enable every time I want to send a notification using curl, otherwise I get `curl: (6) Could not resolve host: ntfy.example.com`.

Can you comment on that @kradalby, what's the reason we can't use the same url? Maybe I could change the `base_domain` to `headscale2.example.com` even though that subdomain doesn't exist?

Edit: Actually, this change is better. Don't know if this is the reason they removed the option to have both `server_url` and `base_domain` the same, but it actually introduced a problem by having both of them set to `headscale.example.com`.

If I have `server_url` and `base_domain` the same, and also have configured the Tailscale client with `tailscale set --accept-dns=true`, then something weird is happening with commands such as the one below. Assuming has to do with the fact that MagicDNS is same as the Headscale domain and everything is being used on the same system?

```
~# openssl s_client -servername headscale.example.com -connect headscale.example.com:443 | openssl x509 -noout -dates
```

This throws errors and doesn't return the dates for the certificate. This is why it introduced issued with my `ntfy.example.com` service too when I changed the `base_domain` to just `example.com`. I now changed it to `magic.example.com` and everything is fine.

PR #2230, opened 2024-11-08
feat: Add autogroup ACLs

gabe565:
I have taken @vinhjaxt's autogroup implementation from https://github.com/juanfont/headscale/issues/657#issuecomment-2336625885, added some code improvements, fixed lint, and added tests. This is still a WIP, but so far it's working as intended.

- [x] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [x] added unit tests
- [x] added integration tests
- [ ] updated documentation if needed
- [x] updated CHANGELOG.md

Fixes #657

Issue #2417, opened 2025-02-07
Tags tracking bug


kradalby:
This bug is a tracking bug for all work related to reworking tags.

Within the scope of this work:

- Tags should not be owned by users
- They can be set from the following sources:
  - CLI
    - can only "force" tags that exist in policy
  - Pre auth keys
    - pre auth keys can only be created with tags that exist in policy.
  - Hostinfo (with approval) 
- Handled in ACL
- Updates are propagated properly 
- `[types.Node].IsTagged()`

kradalby:
All of the bugs in v0.28 milestone and this tracking has been implementing, putting us on track to support tags properly.

PR #2572, opened 2025-05-07
feat: add autogroup:member, autogroup:tagged

vdovhanych:
This implements two autogroups `autogroup:member` and `autogroup:tagged`. I talked with @kradalby about leaving out `autogroup:self` for now, as that is a bit trickier in the end.

The logic of the autogroups and test is taken from the draft pr #2230, but it has been implemented for the new policy structure. 

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,8 @@ working in v1 and not tested might be broken in v2 (and vice versa).
   [#2438](https://github.com/juanfont/headscale/pull/2438)
 - Add documentation for routes
   [#2496](https://github.com/juanfont/headscale/pull/2496)
+- Add support for `autogroup:member`, `autogroup:tagged`
+  [#2572](https://github.com/juanfont/headscale/pull/2572)
 
 ## 0.25.1 (2025-02-25)
 
--- a/hscontrol/policy/v2/types.go
+++ b/hscontrol/policy/v2/types.go
@@ -383,15 +383,20 @@ type AutoGroup string
 
 const (
 	AutoGroupInternet AutoGroup = "autogroup:internet"
+	AutoGroupMember   AutoGroup = "autogroup:member"
 	AutoGroupNonRoot  AutoGroup = "autogroup:nonroot"
+	AutoGroupTagged   AutoGroup = "autogroup:tagged"
 
 	// These are not yet implemented.
-	AutoGroupSelf   AutoGroup = "autogroup:self"
-	AutoGroupMember AutoGroup = "autogroup:member"
-	AutoGroupTagged AutoGroup = "autogroup:tagged"
+	AutoGroupSelf AutoGroup = "autogroup:self"
 )
 
-var autogroups = []AutoGroup{AutoGroupInternet}
+var autogroups = []AutoGroup{
+	AutoGroupInternet,
+	AutoGroupMember,
+	AutoGroupNonRoot,
+	AutoGroupTagged,
+}
 
 func (ag AutoGroup) Validate() error {
 	if slices.Contains(autogroups, ag) {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -122,6 +122,7 @@ working in v1 and not tested might be broken in v2 (and vice versa).
   [#2438](https://github.com/juanfont/headscale/pull/2438)
 - Add documentation for routes
   [#2496](https://github.com/juanfont/headscale/pull/2496)
+- Add support for `autogroup:member`, `autogroup:tagged`
```
kradalby, line 125: Lets add the PR to this

hscontrol/policy/v2/types.go
```diff
@@ -382,16 +382,21 @@ func (p Prefix) Resolve(_ *Policy, _ types.Users, nodes types.Nodes) (*netipx.IP
 type AutoGroup string
 
 const (
+	AutoGroupMember   AutoGroup = "autogroup:member"
 	AutoGroupInternet AutoGroup = "autogroup:internet"
+	AutoGroupTagged   AutoGroup = "autogroup:tagged"
```
kradalby, line 387: nit: we can sort this list alphabetically

hscontrol/policy/v2/types.go
```diff
@@ -409,13 +414,44 @@ func (ag *AutoGroup) UnmarshalJSON(b []byte) error {
...
+	case AutoGroupTagged:
+		// autogroup:tagged represents all devices with a tag in the tailnet.
+		for _, node := range nodes {
+			if len(node.ForcedTags) != 0 {
+				node.AppendToIPSet(&build)
+				continue
+			}
+			if node.Hostinfo != nil && len(node.Hostinfo.RequestTags) != 0 {
+				node.AppendToIPSet(&build)
+			}
```
kradalby, line 447: So this one is a bit more complicated, these are the tags that the node requests, but that doesnt mean its allowed to have them. These has to be checked against the policy to see of the node is allowed to selfassign the tag, this is done in here:

https://github.com/juanfont/headscale/blob/main/hscontrol/policy/v2/types.go#L213

And as you can see by the lengthy comment, the plan is to rewrite that a bit (probably in next release), so you could first just copy that and then I'll fix it up at the same time.

hscontrol/policy/v2/types.go
```diff
@@ -409,13 +414,44 @@ func (ag *AutoGroup) UnmarshalJSON(b []byte) error {
...
+
+		return build.IPSet()
+
+	case AutoGroupTagged:
+		// autogroup:tagged represents all devices with a tag in the tailnet.
+		for _, node := range nodes {
+			if len(node.ForcedTags) != 0 {
+				node.AppendToIPSet(&build)
+				continue
+			}
```
kradalby, line 444: The forced tags dont technically need a "am I allowed check", like mentioned bellow, but it does need "does this exist" check, so you can use the same map.

hscontrol/policy/v2/types.go
```diff
@@ -409,13 +414,44 @@ func (ag *AutoGroup) UnmarshalJSON(b []byte) error {
...
-	return nil, nil
+	case AutoGroupMember:
+		// autogroup:member represents all untagged devices in the tailnet.
+		for _, node := range nodes {
+			if len(node.ForcedTags) != 0 {
+				continue
+			}
+			if node.Hostinfo != nil && len(node.Hostinfo.RequestTags) != 0 {
+				continue
+			}
```
kradalby, line 432: See the comment bellow, this isnt really the ideal signal, because if the node requests a node, but it isnt allowed it, it still belongs to a user.

We should ultimately have `node.IsTagged()`, but that would be work for the next PR. 

This has been helpful because I've added things I didnt realise yet to #2417

hscontrol/policy/v2/types.go
```diff
@@ -922,12 +958,12 @@ type Policy struct {
 }
 
 var (
-	autogroupForSrc       = []AutoGroup{}
-	autogroupForDst       = []AutoGroup{AutoGroupInternet}
-	autogroupForSSHSrc    = []AutoGroup{}
-	autogroupForSSHDst    = []AutoGroup{}
+	autogroupForSrc       = []AutoGroup{AutoGroupMember, AutoGroupTagged}
```
kradalby, line 961: Add a TODO for me here to add these kind of checks for tagOwners and autoApprovers too.

hscontrol/policy/v2/types.go
```diff
@@ -1244,3 +1280,19 @@ func unmarshalPolicy(b []byte) (*Policy, error) {
...
+	}
+
+	var build netipx.IPSetBuilder
+	build.AddPrefix(netip.MustParsePrefix("::/0"))
+	build.AddPrefix(netip.MustParsePrefix("0.0.0.0/0"))
+
+	allTheIps, _ := build.IPSet()
+
+	return allTheIps
+}
```
kradalby, line 1298: This isn't used, so let us remove it for now.

hscontrol/policy/v2/types_test.go
```diff
@@ -765,6 +765,92 @@ func TestResolvePolicy(t *testing.T) {
...
+					ForcedTags: []string{"tag:test"},
+					IPv4:       ap("100.100.101.2"),
+				},
+				{
+					User: users["testuser"],
+					Hostinfo: &tailcfg.Hostinfo{
+						RequestTags: []string{"tag:test"},
+					},
+					IPv4: ap("100.100.101.3"),
+				},
```
kradalby, line 787: Per the comment above, this should be included, as this test would not populate the `tagOwner` map.

We might need a version of the ToResolve test that will take full packetfilter and resolve it.

hscontrol/policy/v2/types_test.go
```diff
@@ -765,6 +765,92 @@ func TestResolvePolicy(t *testing.T) {
...
+					ForcedTags: []string{"tag:test"},
+					IPv4:       ap("100.100.101.3"),
+				},
+				{
+					User: users["user4"],
+					Hostinfo: &tailcfg.Hostinfo{
+						RequestTags: []string{"tag:test"},
+					},
+					IPv4: ap("100.100.101.4"),
+				},
```
kradalby, line 818: Same here

hscontrol/policy/v2/types_test.go
```diff
@@ -765,6 +765,92 @@ func TestResolvePolicy(t *testing.T) {
...
+					ForcedTags: []string{"tag:test"},
+					IPv4:       ap("100.100.101.2"),
+				},
+				{
+					User: users["testuser"],
+					Hostinfo: &tailcfg.Hostinfo{
+						RequestTags: []string{"tag:test"},
+					},
+					IPv4: ap("100.100.101.3"),
+				},
```
kradalby, line 841: Same here

kradalby (changes requested):

kradalby:
Great start!

hscontrol/policy/v2/types_test.go
vdovhanych: I reworked the test logic. This should test the functionality better. But honestly, I'm not really sure if this is enough or if a more robust testing should be added.

hscontrol/policy/v2/types_test.go
vdovhanych: Also reworked, but same as in previous comment applies. Not sure about it.

hscontrol/policy/v2/types.go
vdovhanych: done in https://github.com/juanfont/headscale/commit/2b3ee902a711b9c5936ce6021c405c1fcb08103b

kradalby (approved):
I think this looks great, I am happy to merge this and add more tests as we continue testing.

vdovhanych:
> I think this looks great, I am happy to merge this and add more tests as we continue testing.

Okay, cool. I will squash the fixup commit. 
Also, @kradalby do you think I should remove some comments from the code? They were more for me than something essential to have there.

kradalby:
None of them seem to out of place, let's leave them and then we can update them as needed

vdovhanych:
Squashed

Issue #2475, opened 2025-03-13
[Bug] Internal server error when logging in using Google OIDC


adriangoransson:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Similar to https://github.com/juanfont/headscale/issues/2465 which was closed.

An existing (since Headscale 0.23) tried to log in for the first time since upgrading to 0.25, without success.

```
2025-03-13T08:09:14Z ERR http internal server error error="creating or updating user: constraint failed: UNIQUE constraint failed: users.name (2067)" code=500
2025-03-13T08:09:24Z ERR user msg: invalid code error="could not exchange code for token: oauth2: \"invalid_grant\" \"Bad Request\"" code=403
2025-03-13T08:09:34Z ERR http internal server error error="creating or updating user: constraint failed: UNIQUE constraint failed: users.name (2067)" code=500
2025-03-13T08:10:07Z ERR http internal server error error="creating or updating user: constraint failed: UNIQUE constraint failed: users.name (2067)" code=500
2025-03-13T08:11:07Z ERR http internal server error error="creating or updating user: constraint failed: UNIQUE constraint failed: users.name (2067)" code=500
```

Even after destroying the user (along with the associated nodes and preauthkeys) and restarting headscale, the issue persists.

I tried to investigate how the OIDC handling has changed over the versions, but from what I could gather most issues seem to stem from the email having possibly changed? This has not happened in our case.

### Expected Behavior

The current user should be migrated to the new format.

### Steps To Reproduce

1. Setup a user with Google OIDC on Headscale 0.23.0.
2. Migrate to 0.24 and then 0.25
3. Try to reauthenticate

### Environment

```markdown
- OS: macOS 15.3.1
- Headscale version: 0.25.1
- Tailscale version: 1.80.2
```

### Runtime environment

- [x] Headscale is behind a (reverse) proxy
- [x] Headscale runs in a container

### Anything else?

```yaml
# config.yaml

map_legacy_users: true

oidc:
  only_start_if_oidc_is_available: true
  issuer: https://accounts.google.com
  client_id: <redacted>
  client_secret_path: /secrets/headscale/oidc_client_secret

  expiry: 180d
  use_expiry_from_token: false

  scope: ["openid", "profile", "email"]
  extra_params:
    domain_hint: <redacted>

  allowed_domains:
    - <redacted>

  strip_email_domain: true
```

[...]

PR #2611, opened 2025-05-18
feat(oidc): allow email prefix as username fallback

ventsislav-georgiev:
When using OIDC login and the username is not properly passed with the claims the login fails with:
```
ERR http internal server error error="creating or updating user: constraint failed: UNIQUE constraint failed: users.name (2067)" code=500
```

The proposed solution is to use the email prefix (before @) as the username when the email is marked as verified. _[Question] Should we use the whole email as fallback?_

This resolves the OIDC login from trusted domains like using Google's OAuth 2.0.
Issue ref: https://github.com/juanfont/headscale/issues/2475

### DIFF
--- a/hscontrol/types/users_test.go
+++ b/hscontrol/types/users_test.go
@@ -307,6 +307,7 @@ func TestOIDCClaimsJSONToUser(t *testing.T) {
 			want: User{
 				Provider: util.RegisterMethodOIDC,
 				Email:    "test@test.no",
+				Name:     "test", // Expect email prefix to be used as fallback name
 				ProviderIdentifier: sql.NullString{
 					String: "/test",
 					Valid:  true,
@@ -325,6 +326,7 @@ func TestOIDCClaimsJSONToUser(t *testing.T) {
 			want: User{
 				Provider: util.RegisterMethodOIDC,
 				Email:    "test2@test.no",
+				Name:     "test2", // Expect email prefix to be used as fallback name
 				ProviderIdentifier: sql.NullString{
 					String: "/test2",
 					Valid:  true,
@@ -444,6 +446,26 @@ func TestOIDCClaimsJSONToUser(t *testing.T) {
 					Valid:  true,
 				},
 				ProfilePicURL: "https://cdn.casbin.org/img/casbin.svg",
+			},
+		},
+		{
+			name: "empty-username-use-email-prefix",
+			jsonstr: `
+{
+  "sub": "123456789",
+  "email": "johndoe@example.com",
+  "email_verified": true,
+  "iss": "https://auth.example.com"
+}
+			`,
+			want: User{
+				Provider: util.RegisterMethodOIDC,
+				Email:    "johndoe@example.com",
+				Name:     "johndoe", // Should use email prefix
+				ProviderIdentifier: sql.NullString{
+					String: "https://auth.example.com/123456789",
+					Valid:  true,
+				},
 			},
 		},
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # CHANGELOG
 
 ## Next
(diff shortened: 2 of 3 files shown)


hscontrol/types/users_test.go
```diff
@@ -446,6 +448,26 @@ func TestOIDCClaimsJSONToUser(t *testing.T) {
...
+  "sub": "123456789",
+  "email": "johndoe@example.com",
+  "email_verified": true,
+  "iss": "https://auth.example.com"
+}
+			`,
+			want: User{
+				Provider: util.RegisterMethodOIDC,
+				Email:    "johndoe@example.com",
+				Name:     "johndoe", // Should use email prefix
```
kradalby, line 464: I do not think we really want to do this as it would be unpredictable, what do you do if a oidc has two domains and colliding users? The full email would be a possibility, but then we should use the email field in the db.

kradalby (commented):
I think that the issue isnt that username doesnt get populated, but rather that there exists codepaths which use username, and not email. 

Instead of chopping and gluing we should rather fix that, so that the user is defined by the field that is set, email, username, Provider identity and so on.

ventsislav-georgiev:
@kradalby Completely agree that there could be a bigger change involving the removal of the expectation that the username is unique. However I'm not that familiar with the code base and this is a must have feature at the moment for us.

Since 0.24 we are having issues with the user creation and each new user requires manual operations.

This change is working for us and since we have only one allowed domain in the OIDC, we are guaranteed to have unique usernames. Could we make this configurable? If someone is in our situation this is a perfectly valid solution.

kradalby:
Have you tried 0.26?

It would be helpful to hear how that works with your setup, and particularly, if you could set up a separate test setup from scratch with a new database and your oidc setup and how that behaves.

I'm going to close this PR, the other issue seem to have become a bit of a mix of different issues, it would be great if you test like mentioned above and then create a new issue with your observation on 0.26 for both your current setup and for a clean setup.

ventsislav-georgiev:
Yes, I've meant from 0.24 this is happening and in the issue I've references I said it has worsen with 0.26: https://github.com/juanfont/headscale/issues/2475#issuecomment-2881583450

We use Google OAuth 2.0, which does not provide the custom preferred_username claim.
What happens in 0.26 is a new user comes from the OIDC, his username is left blank, and is created in the sqlite table for the users. Then when next one comes, his username is also blank and the creation fails with: 
```
ERR http internal server error error="creating or updating user: constraint failed: UNIQUE constraint failed: users.name (2067)" code=500
```

Because the first user already exists with the blank username.

ventsislav-georgiev:
I haven't tested from scratch as it is difficult for me to reset or setup a separate test instance.

The issue occurs with existing db which was gradually updated from version 0.22.1 onward. Now on 0.26.0.

These are the indexes from the sqlite db in our instance:
image

It appears this migration hasn't been run: https://github.com/juanfont/headscale/blob/d2879b2b3675c6c38b1a39dd7a7b44c1679287a9/hscontrol/db/db.go#L544

I suppose I have to manually update the table to have the expected indexes?

kradalby:
> The issue occurs with existing db which was gradually updated from version 0.22.1 onward. Now on 0.26.0.

While it should not, doing gradually upgrades _might_ be a problem if you did every .patch release and so on. It is very strange tho that it didnt pick up the migration...

If you take a backup, you could try to delete it from the migration table to force it to run again. 

If it is an option, and you can sanitise the database and send it to me, that might be helpful. If this is an option, getting the old 0.22.1 version and the 0.26.0 version would be helpful.

kradalby:
ah great, can you post your findings from this on the issue so others can check if they are also missing the update?

DDL wont be that helpful, but its ok, glad you fixed it.

kradalby:
actually, you can email me the ddl, find my email in my github

PR #2640, opened 2025-06-01
Correctly document the default for dns.override_local_dns

dotlambda:
See https://github.com/juanfont/headscale/pull/2438#discussion_r2046953908.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -270,9 +270,9 @@ dns:
   # `hostname.base_domain` (e.g., _myhost.example.com_).
   base_domain: example.com
 
-  # Whether to use the local DNS settings of a node (default) or override the
-  # local DNS settings and force the use of Headscale's DNS configuration.
-  override_local_dns: false
+  # Whether to use the local DNS settings of a node or override the local DNS 
+  # settings (default) and force the use of Headscale's DNS configuration.
+  override_local_dns: true
 
   # List of DNS servers to expose to clients.
   nameservers:


nblock (changes requested):
Thanks for bringing this up, but I think we should change the [default value in case the option `dns.override_local_dns` is **missing**](https://github.com/juanfont/headscale/blob/b8044c29ddc59d9c6346337d589b73a7e5b0511e/hscontrol/types/config.go#L292) in the configuration.

dotlambda:
> we should change the [default value in case the option `dns.override_local_dns` is **missing**](https://github.com/juanfont/headscale/blob/b8044c29ddc59d9c6346337d589b73a7e5b0511e/hscontrol/types/config.go#L292) in the configuration instead.

Sure but that would be another breaking change.

nblock:
> Sure but that would be another breaking change.

Good point, depends on who you ask. Changing the default in the config is a breaking change for users that either started using Headscale with 0.26.0 or upgraded *and* migrated the config.

Changing the fallback value in the code is a breaking change for users that upgraded but did not migrate the configuration while upgrading to 0.26.0.

@kradalby what is your take?

nblock:
> Sure but that would be another breaking change.

Let's go with your suggestion. Thx for bringing it up!

nblock:
@dotlambda could you please rebase and force-push? Github CI seems to be broken again :(

PR #2649, opened 2025-06-19
cmd/hi: fixes and qol

kradalby:
fixes some extraction of artefacts and makes it a dev shell command


cmd/hi/cleanup.go
```diff
@@ -47,25 +47,40 @@ func killTestContainers(ctx context.Context) error {
...
+				_ = cli.ContainerKill(ctx, cont.ID, "KILL")
+			}
+			
+			// Then remove the container
+			if err := cli.ContainerRemove(ctx, cont.ID, container.RemoveOptions{
+				Force: true,
+			}); err == nil {
+				removed++
 			}
 		}
```
ghost: In `killTestContainers`, have you considered implementing a timeout or exponential backoff for container removal in case a container is temporarily locked? This could help avoid situations where a container fails to remove immediately but would succeed with a retry.

cmd/hi/cleanup.go
```diff
@@ -138,7 +165,12 @@ func cleanCacheVolume(ctx context.Context) error {
 	defer cli.Close()
 
 	volumeName := "hs-integration-go-cache"
-	_ = cli.VolumeRemove(ctx, volumeName, true)
+	err = cli.VolumeRemove(ctx, volumeName, true)
+	if err != nil {
+		fmt.Printf("Go module cache volume not found or already removed\n")
+	} else {
+		fmt.Printf("Removed Go module cache volume: %s\n", volumeName)
+	}
```
ghost: The error handling for `VolumeRemove` now includes a log message, but it doesn't distinguish between 'volume not found' and other potential errors. Would it be more informative to check for specific error types to provide more accurate feedback?

cmd/hi/docker.go
```diff
@@ -207,6 +232,16 @@ func generateRunID() string {
...
 }
 
+// generateRunIDFromContainerName extracts the run ID from container name.
+func generateRunIDFromContainerName(containerName string) string {
+	// Extract run ID from container name like "headscale-test-suite-20250618-143802"
+	parts := strings.Split(containerName, "-")
+	if len(parts) >= 2 {
+		return strings.Join(parts[len(parts)-2:], "-")
+	}
+	return containerName
```
ghost: The function `generateRunIDFromContainerName` assumes a specific container naming format. If this format changes in the future, this function might break. Have you considered adding a fallback mechanism or a more robust parsing approach?

cmd/hi/docker.go
```diff
@@ -89,6 +91,17 @@ func runTestContainer(ctx context.Context, config *RunConfig) error {
...
+	time.Sleep(2 * time.Second)
+
+	// Extract artifacts from test containers before cleanup
+	if err := extractArtifactsFromContainers(ctx, resp.ID, logsDir, config.Verbose); err != nil && config.Verbose {
+		log.Printf("Warning: failed to extract artifacts from containers: %v", err)
+	}
+
+	// Always list control files regardless of test outcome
+	listControlFiles(logsDir)
+
```
ghost: The 2-second sleep before extracting artifacts seems arbitrary. Would it be better to implement a more deterministic approach to ensure container output is flushed, perhaps by checking container status or using a notification mechanism?

cmd/hi/tar_utils.go
```diff
@@ -0,0 +1,95 @@
...
+				if _, err := io.Copy(outFile, tr); err != nil {
+					return fmt.Errorf("failed to copy file contents: %w", err)
+				}
+				return nil
+			}
+		}
+	}
+
+	return fmt.Errorf("file %s not found in tar", fileName)
+}
```
ghost: In `extractFileFromTar`, the function returns an error if the file isn't found, but the caller often silently ignores these errors. Would it be better to distinguish between 'file not found' errors and other extraction errors to avoid unnecessary error logs for expected situations?

cmd/hi/docker.go
```diff
@@ -356,9 +398,293 @@ func listControlFiles(logsDir string) {
...
+						log.Printf("Including container %s (created %s)", containerName, createdTime.Format("15:04:05"))
+					}
+				}
+				break
+			}
+		}
+	}
+	
+	return testRunContainers
+}
```
ghost: The time window of 5 minutes in `getCurrentTestContainersByTime` could potentially include containers from other test runs if they happen to start within this window. Have you considered using a shorter time window or additional heuristics to improve the accuracy of container selection?

cmd/hi/tar_utils.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		if err != nil {
+			return fmt.Errorf("failed to read tar header: %w", err)
+		}
+
+		// Clean the path to prevent directory traversal
+		cleanName := filepath.Clean(header.Name)
+		if strings.Contains(cleanName, "..") {
+			continue // Skip potentially dangerous paths
+		}
```
ghost, line 70: The path traversal protection in `extractDirectoryFromTar` uses string.Contains to check for '..', which could have false positives. Would it be more secure to use a dedicated path traversal detection function or regex pattern?

cmd/hi/tar_utils.go
```diff
@@ -0,0 +1,95 @@
...
+
+		targetPath := filepath.Join(targetDir, filepath.Base(cleanName))
+
```
ghost, line 73: The path handling in `extractDirectoryFromTar` only uses filepath.Base() which might discard directory structure in the tar. Was this intentional, or should the relative directory structure be preserved?

integration/tsic/tsic.go
```diff
@@ -310,6 +311,9 @@ func New(
 			)
 		}
 
+		// Add integration test labels if running under hi tool
+		dockertestutil.DockerAddIntegrationLabels(tailscaleOptions, "tailscale")
+
```
ghost: There are three separate calls to `dockertestutil.DockerAddIntegrationLabels` for tailscale containers in this file (lines 314-316, 328-330, and 341-343). Could these be consolidated to reduce code duplication? Perhaps by moving the call to a common helper function that's used in all three container creation paths?

PR #2658, opened 2025-06-25
remove unneeded check

stavros-k:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

There is no need to check if parent is set here, we only care if the specific key is set.
Before this change, only by applying env `HEADSCALE_NOISE_PRIVATE_KEY_PATH=/some/path` would trigger this error.
I had to also set env `HEADSCALE_NOISE=true` to bypass it.


kradalby (approved):
Great find, thanks

PR #2659, opened 2025-06-25
add health command

### LINKED ISSUE #2275: [Feature] healthcheck command
### Use case

Docker HEALTHCHECK

### Description

Headscale does not come with a (easy) way of checking health using only the CLI. This is especially painful when running Headscale inside a docker container

### Contribution

Should be pretty simple as it could tie into the already existing `/health` endpoint

stavros-k:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Adds a health command -  closes #2275

### DIFF
--- a/cmd/headscale/cli/health.go
+++ b/cmd/headscale/cli/health.go
@@ -1,0 +1,39 @@
+package cli
+
+import (
+	"fmt"
+
+	v1 "github.com/juanfont/headscale/gen/go/headscale/v1"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(healthCmd)
+}
+
+var healthCmd = &cobra.Command{
+	Use:   "health",
+	Short: "Check the health of the Headscale server",
+	Long:  "Check the health of the Headscale server. This command will return an exit code of 0 if the server is healthy, or 1 if it is not.",
+	Run: func(cmd *cobra.Command, args []string) {
+		output, _ := cmd.Flags().GetString("output")
+		ctx, client, conn, cancel := newHeadscaleCLIWithConfig()
+		defer cancel()
+		defer conn.Close()
+
+		response, err := client.Health(ctx, &v1.HealthRequest{})
+		if err != nil {
+			ErrorOutput(err, "Error checking health", output)
+		}
+
+		if response.Status == v1.HealthStatus_Pass {
+			SuccessOutput("Headscale server is healthy", "", output)
+		}
+
+		ErrorOutput(
+			fmt.Errorf("server health check failed: %s", response.Status),
+			"Headscale server is unhealthy",
+			output,
+		)
+	},
+}
--- a/hscontrol/grpcv1.go
+++ b/hscontrol/grpcv1.go
@@ -819,4 +819,18 @@ func (api headscaleV1APIServer) DebugCreateNode(
 	return &v1.DebugCreateNodeResponse{Node: newNode.Node.Proto()}, nil
 }
 
+func (api headscaleV1APIServer) Health(
+	ctx context.Context,
+	request *v1.HealthRequest,
+) (*v1.HealthResponse, error) {
+	if err := api.h.state.PingDB(ctx); err != nil {
+		log.Error().
+			Err(err).
+			Msg("Health check failed: database ping failed")
+		return &v1.HealthResponse{Status: v1.HealthStatus_Fail}, nil
+	}
+
+	return &v1.HealthResponse{Status: v1.HealthStatus_Pass}, nil
+}
+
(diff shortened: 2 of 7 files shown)


stavros-k:
I've tested both http setup and https with existing certs. and health passes

proto/headscale/v1/headscale.proto
```diff
@@ -209,3 +217,14 @@ service HeadscaleService {
   //     };
   // }
 }
+
+enum HealthStatus {
+  Fail = 0;
+  Pass = 1;
```
kradalby: I do not think that a bool here is meaningful given the underlying check is "is the db there" which is actually not a good indicator of health. 

We could start implementing this interface, but I think it should return a json with fields about what is checks, for example:

```
{
  db_ping: true
}
```

and then over time we can add more when we know what health is.

hscontrol/grpcv1.go
```diff
@@ -819,4 +819,18 @@ func (api headscaleV1APIServer) DebugCreateNode(
 	return &v1.DebugCreateNodeResponse{Node: newNode.Node.Proto()}, nil
 }
 
+func (api headscaleV1APIServer) Health(
+	ctx context.Context,
+	request *v1.HealthRequest,
+) (*v1.HealthResponse, error) {
+	if err := api.h.state.PingDB(ctx); err != nil {
```
kradalby, line 783: This function is from even before I joined the project, not sure how meaningful it is. Specially since we preliminary use SQLite and checking if the database is available is more of a network thing.

cmd/headscale/cli/health.go
```diff
@@ -0,0 +1,39 @@
...
+		defer cancel()
+		defer conn.Close()
+
+		response, err := client.Health(ctx, &v1.HealthRequest{})
+		if err != nil {
+			ErrorOutput(err, "Error checking health", output)
+		}
+
+		if response.Status == v1.HealthStatus_Pass {
+			SuccessOutput("Headscale server is healthy", "", output)
```
kradalby: said JSON data should be outputted here for implementers of UI and so on.

cmd/headscale/cli/health.go
```diff
@@ -0,0 +1,29 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		output, _ := cmd.Flags().GetString("output")
+		ctx, client, conn, cancel := newHeadscaleCLIWithConfig()
+		defer cancel()
+		defer conn.Close()
+
+		response, err := client.Health(ctx, &v1.HealthRequest{})
+		if err != nil {
+			ErrorOutput(err, "Error checking health", output)
+		}
```
ghost, line 25: In cmd/headscale/cli/health.go, the error handling at line 23-25 doesn't exit the program with a non-zero exit code. Since this is a health check command, shouldn't it communicate failure through its exit code as mentioned in the Long description?

cmd/headscale/cli/health.go
```diff
@@ -0,0 +1,29 @@
...
+
+		SuccessOutput(response, "", output)
+	},
+}
```
ghost, line 29: In cmd/headscale/cli/health.go, there's a missing `return` statement after the successful health check at line 30. Without a return, the code will always execute the error output code, even when the health check passes.

cmd/headscale/cli/health.go
stavros-k: ErrorOutput does exit the program.

hscontrol/grpcv1.go
stavros-k: It calls the std lib sql.DB.Ping(). I'm not really sure what ping means for sqlite, but I would assume it checks if the file is still open.

cmd/headscale/cli/health.go
stavros-k: I'm now sending the json back

proto/headscale/v1/headscale.proto
stavros-k: I've changed it to use this format now, please check.
Should we also add an `api: true` which I assume if this command succeeds, api is running too.

cmd/headscale/cli/health.go
stavros-k: After success the function ends.

proto/headscale/v1/headscale.proto
```diff
@@ -209,3 +217,9 @@ service HeadscaleService {
   //     };
   // }
 }
+
+message HealthRequest {}
+
+message HealthResponse {
+  bool db_ping = 1;
```
kradalby: last comment:

```suggestion
  bool database_connectivity = 1;
```

vic1707:
Hi, any updates on this?

PR #2671, opened 2025-07-06
all: use immutable node view in read path

kradalby:
This commit changes most of our (*)types.Node to
types.NodeView, which is a readonly version of the
underlying node ensuring that there is no mutations
happening in the read path.

Based on the migration, there didnt seem to be any, but the
idea here is to prevent it in the future and simplify other
new implementations.


hscontrol/poll.go
```diff
@@ -268,18 +272,22 @@ func (m *mapSession) serveLongPoll() {
...
+			m.node, err = m.h.state.GetNodeByID(nv.ID())
 			if err != nil {
 				m.errf(err, "Could not get machine from db")
 
 				return
 			}
+			// Update the node view to reflect the latest node state
+			// TODO(kradalby): This should become a full read only path, with no update for the node view
+			// in the new mapper model.
+			nv = m.node.View()
```
ghost, line 284: In hscontrol/poll.go's serveLongPoll, there's a comment mentioning 'TODO(kradalby): This should become a full read only path, with no update for the node view in the new mapper model.' This seems like a leftover task that could be addressed in this PR since it's already focused on making the read path immutable. Is this something that should be completed now or is it intentionally left for future work?

Issue #2319, opened 2024-12-28
[Feature] Support for node attributes (nodeattrs)


nielscil:
### Use case

I would like to use headscale and NextDNS together. Tailscale supports configuring different profiles using node attributes. This is something I would like to use in headscale too.

In addition some other features also need the use of node attributes, like #2146

### Description

This feature request would implement the use of `nodeattrs` inside the policy file and support nextDNS profile options as metioned in the official [tailscale docs](https://tailscale.com/kb/1218/nextdns).

example of policy
``` json
{
	"acls": ["..."],
	"nodeAttrs": [
		{
			"target": ["*"],
			"attr": [
                                "nextdns:abc123",
				"nextdns:no-device-info",
			],
		},
	],
}
```

_this issue is a duplicate of #1407, but is closed because it got stale._

### Contribution

- [ ] I can write the design doc for this feature
- [X] I can contribute this feature

### How can it be implemented?

- Add  `nodeAttrs` to policy file and parse inside `ACLPolicy`
- Retrieve attributes in `mapper.go`
- When using NextDNS DOH (already checked in `mapper.go`)
    - When `nextdns`-attribute is provided,  change the profile of the uri
    - When `nextdns:no-device-info` is provided, skip adding the extra device info (currently done in `mapper.go`)
- Add attributes to `tnode.CapMap` inside `tail.go`.

kradalby:
I think we already support nextdns via the config, just not via nodeattrs.

nielscil:
Yeah, but that is on global level, isn't it? I would like to have different profile for different users.

mindolo:
Hello!

Just wanted to note that support for node attributes would be critical for enabling MagicDNS AAAA records, a feature recently added to Tailscale (1.84+).

This is now supported via a new `magicdns-aaaa` node attribute, as seen in this upstream issue: https://github.com/tailscale/tailscale/issues/1152

Implementing this would also resolve the functionality requested in this old and closed issue: https://github.com/juanfont/headscale/issues/1027

And btw, thank you for headscale.

tywil04:
Just to bump, support for `nodeAttrs` would (I believe) enable support for Taildrive: https://tailscale.com/kb/1369/taildrive.

englut:
Would also like support for `nodeAttrs` as I want to [disable captive portal detection](https://tailscale.com/kb/1457/captive-portals#disable-captive-portal-detection).

kradalby:
[...]

PR #2724, opened 2025-08-09
feat: Add `magicdns_peer_aaaa` config option

### LINKED ISSUE #2723: [Feature] Add support for the MagicDNSPeerAAAA Node Attribute
### Use case

Recent (1.84+) versions of tailscale support a new `magicdns-aaaa` node attribute that configures MagicDNS to respond to AAAA queries.

See references:
 * https://github.com/ninech/tailscale/commit/9555912f29c404ac06d3b4a4b615f30e64b4c2c0
 * https://github.com/tailscale/tailscale/issues/1152#issuecomment-3138350402

### Description

#1027 is a (closed) issue to support AAAA records in MagicDNS. It was closed due to lack of Upstream support, but now that gap has closed significantly with the addition of the new `magicdns-aaaa` Node Attribute, this is an opt-in feature that users can enable.

In Tailscale this optin seems to be done via policies, but headscale is lacking support for NodeAttribute configuration in the policy (#2319)

There is a way to implement this via a global configuration tunable without requiring full NodeAttribute configuration capabilities.

### Contribution

While there is a long term issue open to write support for Node Attribute configuration via policies (#2319) I think there is a quicker way to support the new feature in a similar way to `RandomizeClientPort`.

As it stands, RandomizeClientPort (headscale 0.26.1) is implemented as a configuration parameter (`randomize_client_port` in the yaml file) rather than an ACL policy tuneable, and it is set directly in the Node mapper: https://github.com/juanfont/headscale/blob/v0.26.1/hscontrol/mapper/tail.go#L124-L126

I am proposing to implement support for `MagicDNSPeerAAAA` in a very similar way.
If the config option is enabled _and_ the Node has a valid Overlay IPv6, the NodeAttribute will be set, or skipped otherwise.

I have an implementation for this and the actual code change is minimal (4 lines in `hscontrol/types/config.go` and 3 in `hscontrol/mapper/tail.go` ignoring comments and tests).
I will be submitting a PR for this later today.

mindolo:
This change introduces a new boolean config option, `magicdns_peer_aaaa`.

When enabled, it adds the `magicdns-aaaa` node attribute to devices with a valid overlay IPv6 address.
This attribute allows MagicDNS to resolve AAAA (IPv6) DNS queries for other nodes in the tailnet.

This is a new opt-in feature for Tailscale 1.84+ and is implemented as a global config option, similar to `RandomizeClientPort`, since support for configuring node attributes via policies is not yet available.

Fixes #2723 

- [x] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -50,6 +50,7 @@ jobs:
           - TestDERPVerifyEndpoint
           - TestResolveMagicDNS
           - TestResolveMagicDNSExtraRecordsPath
+          - TestMagicDNSPeerAAAA
           - TestDERPServerScenario
           - TestDERPServerWebsocketScenario
           - TestPingAllByIP
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -404,3 +404,11 @@ logtail:
 # default static port 41641. This option is intended as a workaround for some buggy
 # firewall devices. See https://tailscale.com/kb/1181/firewalls/ for more information.
 randomize_client_port: false
+
+
+# Enables MagicDNS AAAA query resolution for tailnet nodes with IPv6.
+# When this option is enabled, devices that have a valid overlay IPv6 address will
+# automatically gain the `magicdns-aaaa` Node Attribute.
+# This attribute allows them to resolve AAAA (IPv6) DNS queries through MagicDNS for 
+# other nodes within the tailnet.
+magicdns_peer_aaaa: false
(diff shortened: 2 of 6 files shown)


mindolo:
I opened this draft PR to ask for help as I was stuck on the integration tests, but as soon as I did, it all clicked and I was able to fix the test and get it to pass.

kradalby:
I posted an update on https://github.com/juanfont/headscale/issues/2319 (nodeAttrs).

I suspect we want to implement this in the "correct" way in the policy instead of adding more things to the config file (like we do for random port number, which should also be nodeAttrs)

PR #2727, opened 2025-08-14
Fix fatal error on missing config file

oneingan:
This PR fixes a fatal error when no config file is found during CLI operations, ensuring consistent behavior with empty config files. Previously, `LoadConfig` incorrectly checked for `fs.ErrNotExist` instead of Viper's `ConfigFileNotFoundError`, causing the CLI to fail for commands like `headscale nodes list` when relying on environment variables (e.g., `HEADSCALE_CLI_ADDRESS`). Now, a missing config file logs a warning and falls back to defaults, matching the empty config file case.

### DIFF
--- a/hscontrol/types/config.go
+++ b/hscontrol/types/config.go
@@ -336,11 +336,11 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetDefault("prefixes.allocation", string(IPAllocationStrategySequential))
 
 	if err := viper.ReadInConfig(); err != nil {
-		if errors.Is(err, fs.ErrNotExist) {
-			log.Warn().Msg("No config file found, using defaults")
-			return nil
-		}
-
+    	if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+        	log.Warn().Msg("No config file found, using defaults")
+        	return nil
+    	}
+    	
 		return fmt.Errorf("fatal error reading config file: %w", err)
 	}
 


hscontrol/types/config.go
```diff
@@ -336,11 +336,11 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetDefault("prefixes.allocation", string(IPAllocationStrategySequential))
 
 	if err := viper.ReadInConfig(); err != nil {
-		if errors.Is(err, fs.ErrNotExist) {
-			log.Warn().Msg("No config file found, using defaults")
-			return nil
-		}
-
+    	if _, ok := err.(viper.ConfigFileNotFoundError); ok {
```
nblock, line 339: There's a [golangci-lint error](https://github.com/juanfont/headscale/actions/runs/16963878176/job/48311279382?pr=2727).:

```
 hscontrol/types/config.go:339:18: type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)
    	if _, ok := err.(viper.ConfigFileNotFoundError); ok {
    	            ^
```

Please have a look at this.

The documentation mentions this bug specifically. Can you remove this documentation block, please?

https://github.com/juanfont/headscale/blob/3e3c72ea6f36cfff1f467b4e3f350886973d32bb/docs/ref/remote-cli.md?plain=1#L69-L74

nblock (dismissed):

PR #2732, opened 2025-08-15
Add DERP documentation

nblock:
Document the embedded DERP server along with some typical configuration scenarios.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md


docs/ref/derp.md
```diff
@@ -0,0 +1,145 @@
...
+relay traffic between two nodes in case a direct connection can't be established. Headscale provides an embedded DERP
+server to ensure seamless connectivity between nodes.
+
+## Configuration
+
+DERP related settings are configured within the `derp` section of the [configuration file](./configuration.md). The
+following sections only use a few of the available settings, check the [example configuration](./configuration.md) for
+all available configuration options.
+
+### Enable embedded DERP
```
kradalby, line 13: Maybe there should be a quick paragraph explaining what embedded DERP is before stating it is disabled?

docs/ref/derp.md
```diff
@@ -0,0 +1,145 @@
...
+
+```yaml title="config.yaml" hl_lines="2"
+derp:
+  urls: []
+```
+
+!!! warning "Single point of failure"
+
+    Removing Tailscale's DERP servers means that there is now just a single DERP server available for clients. This is a
+    single point of failure and could hamper connectivity.
```
kradalby, line 49: This is a good warning, I see that both "Enable embedded DERP" and "Remove Tailscale's DERP servers" is at the same level in the hierarchy while the latter section does refer to only disable if you have enabled the embedded one. I'm just slightly concern that people might jump to this section independently and disable it and get in trouble.

nblock: Good point, extended the `config.yaml` snippet to include same lines as with the "Enable embedded DERP" section and included a sentence to the warning section that one should check connectivity with the embedded DERP before removing the default DERP map.

kradalby (approved):

PR #2734, opened 2025-08-16
Allow all traffic if acls field is omited from the policy

### LINKED ISSUE #2672: [Bug] Default ACLs differ from headscale

### Current Behavior

Currently, when setting empty ACLs (`{}`), a deny-all default policy is applied. This is the exact opposite of Tailscale (see [documentation](https://tailscale.com/kb/1192/acl-samples#allow-all-default-acl)). Being unable to verify through the docs, issues, or commits whether or not this is intentional, I decided to file an issue in case it is not, and to document it, in case it is.

### Expected Behavior

Headscale mirrors Tailscale's default behaviour.

### Steps To Reproduce

1. clear ACLs
2. ping any host
3. apply allow-all ACL
4. ping any host

### Environment

```markdown
- OS: Linux
- Headscale version: 0.26.1
- Tailscale version: 1.84.0
```

### Runtime environment

### Debug information

Image

afonsofrancof:
This PR fixes #2672.

It makes headscale follow the following tailscale rules.

image

image

In the code this translates to a `nil` `ACLs` field in the `Policy` struct if it doesn't exist in the policy dict (first scenario) , and a non-nil 0-length array in the same field when it exists but is empty (second scenario)

This is nice to have because if you're setting up a file or database policy, you can start with an empty {} as the contents, which means everything is allowed. Then you can add your specific rules later on.

It also prevents people from accidentally locking themselves out of the tailnet when they start experimenting with acls ( happened to me 😅 )

- [x] added unit tests

### DIFF
--- a/docs/ref/acls.md
+++ b/docs/ref/acls.md
@@ -9,9 +9,38 @@ When using ACL's the User borders are no longer applied. All machines
 whichever the User have the ability to communicate with other hosts as
 long as the ACL's permits this exchange.
 
-## ACLs use case example
+## ACL Setup
 
-Let's build an example use case for a small business (It may be the place where
+To enable and configure ACLs in Headscale, you need to specify the path to your ACL policy file in the `policy.path` key in `config.yaml`.
+
+Your ACL policy file must be formatted using [huJSON](https://github.com/tailscale/hujson).
+
+Info on how these policies are written can be found
+[here](https://tailscale.com/kb/1018/acls/).
+
+Please reload or restart Headscale after updating the ACL file. Headscale may be reloaded either via its systemd service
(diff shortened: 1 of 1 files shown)


afonsofrancof:
Let me know if I should add a section to the docs that explains this behavior

kradalby:
Thank you! great find

> Let me know if I should add a section to the docs that explains this behavior

Yes please, and if the tailscale docs you screenshotted is easily referable, add a link there.

afonsofrancof:
Docs done.
Let me know if I need to change anything there :)

kradalby:
I'll let @nblock take a look

PR #2741, opened 2025-08-21
derp: increase update frequency and harden on failures

kradalby:
This commit changes the example config and default time for
how often we refresh the DERP map. If not changed, this
is typically Tailscales official map.

Second commit makes updating of the DERP map from file and url
more robust by retrying with exponential backoff if it fails
and upon failure, keep the old DERP map if we cannot successfully
build a new one.

### DIFF
--- a/hscontrol/derp/derp.go
+++ b/hscontrol/derp/derp.go
@@ -1,13 +1,19 @@
 package derp
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
+	"hash/crc64"
 	"io"
 	"maps"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
+	"sync"
+	"time"
 
 	"github.com/juanfont/headscale/hscontrol/types"
 	"github.com/spf13/viper"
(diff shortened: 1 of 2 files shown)


hscontrol/derp/derp.go
```diff
@@ -79,53 +79,48 @@ func mergeDERPMaps(derpMaps []*tailcfg.DERPMap) *tailcfg.DERPMap {
...
+		return
+	}
+
+	for _, region := range dm.Regions {
+		if len(region.Nodes) == 0 {
+			continue
+		}
 
-	return derpMap
+		rand.Shuffle(len(region.Nodes), func(i, j int) {
```
mikeodr: Is there a way to make this deterministic per Tailnet?

This helps spread the load across nodes in a DERP region.

kradalby: Now it will try to use the base_domain as a seed so it should be relatively stable per headscale.

mikeodr (approved):

hscontrol/derp/derp.go
mikeodr: Love it, thanks!

PR #2756, opened 2025-08-31
<fix> Convert the derpmap parsing from yaml to json to ensure compati…

liuxocakn:
When I use a self-signed and self-hosted derp server, I need to set InsecureForTests to true in order to skip the certificate check. However, when I set InsecureForTests to true in yaml, it didn't work. I added a print of InsecureForTests in the code, and it showed false.

``` go
func loadDERPMapFromPath(path string) (*tailcfg.DERPMap, error) {
	derpFile, err := os.Open(path)
	if err != nil {
		return nil, err
	}
	defer derpFile.Close()
	var derpMap tailcfg.DERPMap
	b, err := io.ReadAll(derpFile)
	if err != nil {
		return nil, err
	}
	err = json.Unmarshal(b, &derpMap)

       // print node InsecureForTests 

	return &derpMap, err
}
```

After some trouble, I found that the definition of the DERPNode structure only has json tags and no yaml, which led to problems with the inversion of the bool type.

``` go
// DERPNode describes a DERP packet relay node running within a DERPRegion.
type DERPNode struct {
	// Name is a unique node name (across all regions).
	// It is not a host name.
	// It's typically of the form "1b", "2a", "3b", etc. (region
	// ID + suffix within that region)
	Name string

	// RegionID is the RegionID of the DERPRegion that this node
	// is running in.
	RegionID int

	// HostName is the DERP node's hostname.
	//
	// It is required but need not be unique; multiple nodes may
	// have the same HostName but vary in configuration otherwise.
	HostName string

	// CertName optionally specifies the expected TLS cert common
	// name. If empty, HostName is used. If CertName is non-empty,
	// HostName is only used for the TCP dial (if IPv4/IPv6 are
	// not present) + TLS ClientHello.
	//
	// As a special case, if CertName starts with "sha256-raw:",
	// then the rest of the string is a hex-encoded SHA256 of the
	// cert to expect. This is used for self-signed certs.
	// In this case, the HostName field will typically be an IP
	// address literal.
	CertName string `json:",omitempty"`

	// IPv4 optionally forces an IPv4 address to use, instead of using DNS.
	// If empty, A record(s) from DNS lookups of HostName are used.
	// If the string is not an IPv4 address, IPv4 is not used; the
	// conventional string to disable IPv4 (and not use DNS) is
	// "none".
	IPv4 string `json:",omitempty"`

	// IPv6 optionally forces an IPv6 address to use, instead of using DNS.
	// If empty, AAAA record(s) from DNS lookups of HostName are used.
	// If the string is not an IPv6 address, IPv6 is not used; the
	// conventional string to disable IPv6 (and not use DNS) is
	// "none".
	IPv6 string `json:",omitempty"`

	// Port optionally specifies a STUN port to use.
	// Zero means 3478.
	// To disable STUN on this node, use -1.
	STUNPort int `json:",omitempty"`

	// STUNOnly marks a node as only a STUN server and not a DERP
	// server.
	STUNOnly bool `json:",omitempty"`

	// DERPPort optionally provides an alternate TLS port number
	// for the DERP HTTPS server.
	//
	// If zero, 443 is used.
	DERPPort int `json:",omitempty"`

	// InsecureForTests is used by unit tests to disable TLS verification.
	// It should not be set by users.
	InsecureForTests bool `json:",omitempty"`
	// or
	//   InsecureForTests bool `json:",omitempty" yaml:"InsecureForTests,omitempty"`

	// STUNTestIP is used in tests to override the STUN server's IP.
	// If empty, it's assumed to be the same as the DERP server.
	STUNTestIP string `json:",omitempty"`

	// CanPort80 specifies whether this DERP node is accessible over HTTP
	// on port 80 specifically. This is used for captive portal checks.
	CanPort80 bool `json:",omitempty"`
}
```

Since the DERPNode code is not in the headscale, I think it would be more reasonable to modify its parsing method to json.

### DIFF
--- a/hscontrol/derp/derp.go
+++ b/hscontrol/derp/derp.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/juanfont/headscale/hscontrol/types"
 	"github.com/spf13/viper"
-	"gopkg.in/yaml.v3"
 	"tailscale.com/tailcfg"
 )
 
@@ -32,7 +31,7 @@ func loadDERPMapFromPath(path string) (*tailcfg.DERPMap, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = yaml.Unmarshal(b, &derpMap)
+	err = json.Unmarshal(b, &derpMap)
 
 	return &derpMap, err
 }


kradalby:
Would this not remove the ability to parse YAML? 

IIRC, it was this way because YAML is a superset of JSON, so if you write it in YAML or JSON, then the yaml parser can do it, but the JSON parser probably cannot?

That said, we can mark the changelog with breaking and change it to only support JSON.

liuxocakn:
> Would this not remove the ability to parse YAML?
> 
> IIRC, it was this way because YAML is a superset of JSON, so if you write it in YAML or JSON, then the yaml parser can do it, but the JSON parser probably cannot?
> 
> That said, we can mark the changelog with breaking and change it to only support JSON.

I'm not quite sure about the compatibility of YAML with JSON, but when I deployed my own project using the version from a month ago and the latest Headscale mainline version, I encountered a YAML parsing issue: InsecureForTests should have been parsed as true, but it was parsed as false, which led me to suspect for a while that it was a problem with my deployment or that headscale forced the local configuration not to set InsecureForTests to true for security reasons.

After I reviewed the source code, I found that the tags in the DERPNode type definition do not have YAML but only JSON, and its source code is not in headscale. Therefore, I personally suggest that only supporting json parsing is more appropriate for maintenance and consistency. So this is my idea for patch! Of course, this is just my personal opinion.

Regarding the issue of YAML parsing DERPNode, separate code can be written for verification. When there is only the main function, only the yaml file of DERPNode is parsed, and the same version of go as headscale is used, problems will occur: The three bool type variables STUNOnly, InsecureForTests, and CanPort80 will be parsed incorrectly, while STUNOnly will occasionally be parsed successfully（the other variables have been correctly parsed）.

ArcticLampyrid:
Try lowercase `insecurefortests`, I remember it will work.

liuxocakn:
> Try lowercase `insecurefortests`, I remember it will work.

image

This is my YAML configuration file. Do you mean I made a mistake, and it should be `insecurefortests` `canport80`?

I'll test it right away. It should be like this. I see that the other fields are in this format

liuxocakn:
> Try lowercase `insecurefortests`, I remember it will work.

Thank you very much. It is indeed for this reason that headscale reading yaml is now working properly. I'm so glad today that I've solved two problems that have been bothering me for several months.

liuxocakn:
> Would this not remove the ability to parse YAML?
> 
> IIRC, it was this way because YAML is a superset of JSON, so if you write it in YAML or JSON, then the yaml parser can do it, but the JSON parser probably cannot?
> 
> That said, we can mark the changelog with breaking and change it to only support JSON.

Changing 'InsecureForTests: true' to 'insecurefortests: true' can work. It's my fault. I'm not very familiar with go or ymal. Mainly, after I wrote my personal website server in go, I haven't used it anymore. I also don't have the chance to use go in my daily life. Haha

PR #2765, opened 2025-09-10
cmd: add option to get and set policy directly from database

### LINKED ISSUE #2630: [Bug] Invalid ACL stored in DB, crash on startup

### Current Behavior

I've used headscale-admin to update te the ACL, then I restarted headscale. On startup, it complained about a host referenced in the ACLs that wasn't defined (I had just deleted it thinking that it wasn't needed anymore).

On startup, it complained about the missing host:
```
handshake-headscale-1  | 2025-05-27T09:18:31Z INF Opening database database=sqlite3 path=/var/lib/headscale/db.sqlite
handshake-headscale-1  | 2025-05-27T09:18:31Z FTL home/runner/work/headscale/headscale/cmd/headscale/cli/serve.go:24 > Error initializing error="creating new headscale: loading ACL policy: creating policy manager: parsing policy: Host \"siemens-distillery\" is not defined in the Policy, please define or remove the reference to it"
```
and then exited.

I patched the database manually using sqlite3 (I basically copied an older version to the currect version in the `policies` table, and that let me start up headscale again.

This is with docker.io/headscale/headscale:v0.26.0 and docker.io/goodieshq/headscale-admin:v0.25.6. 

### Expected Behavior

Headscale should either refuse to store the broken ACL, or ignore the broken (part of the) ACLs on startup, so users can fix it through the command line or the web interface.

### Steps To Reproduce

1. Bring up headscale and headscale admin with docker-compose
2. Add a host to the policies
3. Add an ACL referencing that host
4. Save the config
5. Remote the host
6. Save the config; note that there is no error
7. Restart headscale and observe that it won't start. 

### Environment

```markdown
- OS: docker compose
- Headscale version: v0.26.0
- Tailscale version: n/a
- Headscale Admin: v0.25.6
```

### Runtime environment

### Debug information

see above

kradalby:
this commit adds a new `--bypass-grpc-access-and-database-directly` flag to `headscale policy get|set` which allows the operator to directly get or set the policy from the database. This is useful if there is a broken policy in the database that prevents the server from starting.

Fixes #2630

### DIFF
--- a/cmd/headscale/cli/policy.go
+++ b/cmd/headscale/cli/policy.go
@@ -9,10 +9,10 @@ import (
 	"github.com/juanfont/headscale/hscontrol/db"
 	"github.com/juanfont/headscale/hscontrol/policy"
 	"github.com/juanfont/headscale/hscontrol/types"
+	"github.com/juanfont/headscale/hscontrol/util"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"tailscale.com/types/views"
-	"tailscale.com/util/prompt"
 )
 
 const (
@@ -52,7 +52,13 @@ var getPolicy = &cobra.Command{
 		output, _ := cmd.Flags().GetString("output")
 		var policy string
 		if bypass, _ := cmd.Flags().GetBool(bypassFlag); bypass {
-			if !prompt.YesNo("DO NOT run this command if an instance of headscale is running, are you sure headscale is not running?") {
+			confirm := false
+			force, _ := cmd.Flags().GetBool("force")
+			if !force {
+				confirm = util.YesNo("DO NOT run this command if an instance of headscale is running, are you sure headscale is not running?")
+			}
+
+			if !confirm && !force {
 				ErrorOutput(nil, "Aborting command", output)
 				return
 			}
@@ -60,7 +66,6 @@ var getPolicy = &cobra.Command{
 			cfg, err := types.LoadServerConfig()
 			if err != nil {
 				ErrorOutput(err, fmt.Sprintf("Failed loading config: %s", err), output)
-				return
 			}
 
 			d, err := db.NewHeadscaleDatabase(
@@ -70,13 +75,11 @@ var getPolicy = &cobra.Command{
 			)
 			if err != nil {
 				ErrorOutput(err, fmt.Sprintf("Failed to open database: %s", err), output)
-				return
 			}
 
 			pol, err := d.GetPolicy()
 			if err != nil {
 				ErrorOutput(err, fmt.Sprintf("Failed loading Policy from database: %s", err), output)
-				return
 			}
 
 			policy = pol.Data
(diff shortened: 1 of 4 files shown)


cmd/headscale/cli/policy.go
```diff
@@ -41,21 +50,55 @@ var getPolicy = &cobra.Command{
...
-		defer conn.Close()
-
-		request := &v1.GetPolicyRequest{}
-
-		response, err := client.GetPolicy(ctx, request)
-		if err != nil {
-			ErrorOutput(err, fmt.Sprintf("Failed loading ACL Policy: %s", err), output)
+		var policy string
+		if bypass, _ := cmd.Flags().GetBool(bypassFlag); bypass {
+			if !prompt.YesNo("DO NOT run this command if an instance of headscale is running, are you sure headscale is not running?") {
```
nblock: Headscale crashes when answering the question with `n`:

```console
$ headscale --force policy get --bypass-grpc-and-access-database-directly
DO NOT run this command if an instance of headscale is running, are you sure headscale is not running? [y/n] n
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x55d04da6fab9]

goroutine 1 [running]:
github.com/juanfont/headscale/cmd/headscale/cli.ErrorOutput({0x0?, 0x0?}, {0x55d04daf69e0, 0x10}, {0x0, 0x0})
        /home/user/headscale/cmd/headscale/cli/utils.go:172 +0x39
github.com/juanfont/headscale/cmd/headscale/cli.init.func21(0x55d0501c3b20, {0x55d04dadff11?, 0x4?, 0x55d04dadfde1?})
        /home/user/headscale/cmd/headscale/cli/policy.go:56 +0x50a
github.com/spf13/cobra.(*Command).execute(0x55d0501c3b20, {0xc00009ca20, 0x2, 0x2})
        /home/user/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1019 +0xa91
github.com/spf13/cobra.(*Command).ExecuteC(0x55d0501c4ba0)
        /home/user/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1148 +0x46f
github.com/spf13/cobra.(*Command).Execute(...)
        /home/user/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1071
github.com/juanfont/headscale/cmd/headscale/cli.Execute()
        /home/user/headscale/cmd/headscale/cli/root.go:103 +0x1a
main.main()
        /home/user/headscale/cmd/headscale/headscale.go:42 +0x1fe
```

cmd/headscale/cli/policy.go
```diff
@@ -41,21 +50,55 @@ var getPolicy = &cobra.Command{
...
-		defer cancel()
-		defer conn.Close()
-
-		request := &v1.GetPolicyRequest{}
-
-		response, err := client.GetPolicy(ctx, request)
-		if err != nil {
-			ErrorOutput(err, fmt.Sprintf("Failed loading ACL Policy: %s", err), output)
+		var policy string
+		if bypass, _ := cmd.Flags().GetBool(bypassFlag); bypass {
```
nblock, line 54: The flag does not account for the `--force` flag.

nblock: The interactive question is hidden when redirecting the output to a file:

```console
$ headscale policy get --force --bypass-grpc-and-access-database-directly > dump-1.json
... [hangs and continues only when pressing `y`]
```

Also the json output contains the question:

```
$ cat dump-1.json 
DO NOT run this command if an instance of headscale is running, are you sure headscale is not running? [y/n] {
  "acls": [
    {
      "action": "accept",
      "src": [
        "*"
      ],
      "dst": [
        "100.64.0.99"
      ]
    }
  ]
}
```

Question should be written to stderr.

nblock: Its possible to set an invalid ACL (or invalid JSON) via: `headscale policy set --bypass-grpc-and-access-database-directly --file path/to/broken.json`. Should it only allow valid ACLs?

PR #2752, opened 2025-08-29
fix: return valid AuthUrl in followup request on expired reg id

bobelev:
- **chore: make reg cache expiry tunable**
- **fix: return valid AuthUrl in followup request on expired reg id**

- [x] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [ ] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [x] added integration tests
- [ ] updated documentation if needed
- [ ] updated CHANGELOG.md

Info:
- tailscale client gets a new AuthUrl and sets entry in the regcache
- regcache entry expires
- client doesn't know about that
- client always polls followup request а gets error

When user clicks "Login" in the app (after cache expiry), they visit
invalid URL and get "node not found in registration cache". Some clients
on Windows for e.g. can't get a new AuthUrl without restart the app.

To fix that we can issue a new reg id and return user a new valid
AuthUrl.

PR #2773, opened 2025-09-11
fix: return valid AuthUrl in followup request on expired reg id

bobelev:
- **chore: make reg cache expiry tunable**
- **fix: return valid AuthUrl in followup request on expired reg id**

- [ ] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

Info:
- tailscale client gets a new AuthUrl and sets entry in the regcache
- regcache entry expires
- client doesn't know about that
- client always polls followup request а gets error

When user clicks "Login" in the app (after cache expiry), they visit
invalid URL and get "node not found in registration cache". Some clients
on Windows for e.g. can't get a new AuthUrl without restart the app.

To fix that we can issue a new reg id and return user a new valid
AuthUrl.

### DIFF
--- a/hscontrol/auth.go
+++ b/hscontrol/auth.go
@@ -40,7 +40,7 @@ func (h *Headscale) handleRegister(
 	}
 
 	if regReq.Followup != "" {
-		return h.waitForFollowup(ctx, regReq)
+		return h.waitForFollowup(ctx, regReq, machineKey)
 	}
 
 	if regReq.Auth != nil && regReq.Auth.AuthKey != "" {
@@ -142,6 +142,7 @@ func nodeToRegisterResponse(node *types.Node) *tailcfg.RegisterResponse {
 func (h *Headscale) waitForFollowup(
 	ctx context.Context,
 	regReq tailcfg.RegisterRequest,
+	machineKey key.MachinePublic,
 ) (*tailcfg.RegisterResponse, error) {
 	fu, err := url.Parse(regReq.Followup)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


bobelev:
Didn't know, that I can't reopen force-pushed PRs. This is a fixup and rebase of https://github.com/juanfont/headscale/pull/2752

kradalby:
@bobelev not had the time to look to much, but for your own testing:
- some of the relogin tests are flaky
- TestAutoApproveMultiNetwork does not run in the CI

Just so you do not hunt ghosts unrelated to your changes.

hscontrol/auth.go
```diff
@@ -159,13 +160,46 @@ func (h *Headscale) waitForFollowup(
...
 			return nodeToRegisterResponse(node), nil
 		}
 	}
 
-	return nil, NewHTTPError(http.StatusNotFound, "followup registration not found", nil)
+	// if the follow-up registration isn't found anymore, instruct the client to try a new registration
+	return h.reqToNewRegisterResponse(regReq, machineKey)
+}
+
+func (h *Headscale) reqToNewRegisterResponse(
```
kradalby, line 177: could we have a godoc comment on this func?

integration/auth_oidc_test.go
```diff
@@ -688,6 +689,108 @@ func TestOIDCReloginSameNodeNewUser(t *testing.T) {
 	}, 30*time.Second, 1*time.Second, "log out user2, and log into user1, no new node should be created")
 }
 
+func TestOIDCFollowUpUrl(t *testing.T) {
```
kradalby, line 700: This test looks great, I'm trying to improve the tests a bit with godoc that explains the flow of the test in a bit more human readable format, could you add that?

kradalby (approved):
two docs nits, otherwise looks very good

hscontrol/auth.go
bobelev: done

integration/auth_oidc_test.go
bobelev: Not much of an examples :) Done in free form.

PR #2774, opened 2025-09-11
capver: update latest

kradalby:


kradalby:
I do not understand why updating the images causes this ssh applet error.

kradalby:
If any Alpine + Docker experts want to investigate this that would be swell.

CHANGELOG.md
```diff
@@ -37,7 +39,7 @@ flows. External links are properly secured with noreferrer/noopener attributes.
 
 ## 0.27.0 (2025-10-27)
 
-**Minimum supported Tailscale client version: v1.64.0**
+**Minimum supported Tailscale client version: v1.68.0**
```
nblock: This is probably wrong?

kradalby: fixed

PR #2789, opened 2025-09-30
feat: add autogroup:self

### LINKED ISSUE #2618: Policy: autogroup support
This is tracking issue for `autogroup` support in the headscale Policy, it supersedes #657.

From the [Tailscale docs](https://tailscale.com/kb/1337/acl-syntax#autogroups):
> An [autogroup](https://tailscale.com/kb/1396/targets#autogroups) is a special group that automatically includes users, destinations, or usernames with the same properties.

Headscale aims to implement a subset of the Tailscale `autogroups`, currently supported autogroups are checked:

- [ ] `autogroup:self`
- [x] `autogroup:internet`
- [x] `autogroup:member` (as of #2572)
- [x] `autogroup:tagged` (as of #2572)
- [x] `autogroup:nonroot`

`autogroup:self` is a bit more complicated, the goal would be to get it in as part of this release cycle, but it needs some more thinking as it cannot be resolved as part of the policy and a list of nodes. It needs the information about the `self` (node or user) as well.

Some autogroups are more complicated, and we dont consider them essential for now, which means we might do them later:

- `user:*@<domain>`
- `localpart:*@<domain>`

Not relevant as these describe roles in Tailscale, which does not exist in headscale:
- `autogroup:owner`
- `autogroup:admin`
- `autogroup:auditor`
- `autogroup:billing-admin`
- `autogroup:it-admin`
- `autogroup:network-admin`

Not relevant as headscale only has one tailnet:
- `autogroup:shared`

vdovhanych:
- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

This adds autogroup:self feature and should close #2618

While the feature should now work i still feel this needs a bit more work to be "production" ready. I mainly have some concersn with the performace of headscale when this autogroup will be used in the ACL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,6 +251,8 @@ working in v1 and not tested might be broken in v2 (and vice versa).
   [#2438](https://github.com/juanfont/headscale/pull/2438)
 - Add documentation for routes
   [#2496](https://github.com/juanfont/headscale/pull/2496)
+- Add support for `autogroup:self`
+  !TODO: put the link to the PR here before merging!
 
 ## 0.25.1 (2025-02-25)
 
--- a/hscontrol/policy/v2/policy.go
+++ b/hscontrol/policy/v2/policy.go
@@ -38,6 +38,10 @@ type PolicyManager struct {
 
 	// Lazy map of SSH policies
 	sshPolicyMap map[types.NodeID]*tailcfg.SSHPolicy
+
+	// Lazy map of per-node filter rules (when autogroup:self is used)
+	filterRulesMap    map[types.NodeID][]tailcfg.FilterRule
+	usesAutogroupSelf bool
 }
 
 // NewPolicyManager creates a new PolicyManager from a policy file and a list of users and nodes.
@@ -50,10 +54,12 @@ func NewPolicyManager(b []byte, users []types.User, nodes views.Slice[types.Node
 	}
 
 	pm := PolicyManager{
-		pol:          policy,
-		users:        users,
-		nodes:        nodes,
-		sshPolicyMap: make(map[types.NodeID]*tailcfg.SSHPolicy, nodes.Len()),
+		pol:               policy,
+		users:             users,
+		nodes:             nodes,
+		sshPolicyMap:      make(map[types.NodeID]*tailcfg.SSHPolicy, nodes.Len()),
+		filterRulesMap:    make(map[types.NodeID][]tailcfg.FilterRule, nodes.Len()),
+		usesAutogroupSelf: policy.usesAutogroupSelf(),
 	}
 
 	_, err = pm.updateLocked()
@@ -72,8 +78,17 @@ func (pm *PolicyManager) updateLocked() (bool, error) {
 	// policies for nodes that have changed. Particularly if the only difference is
 	// that nodes has been added or removed.
 	clear(pm.sshPolicyMap)
+	clear(pm.filterRulesMap)
 
-	filter, err := pm.pol.compileFilterRules(pm.users, pm.nodes)
+	// Check if policy uses autogroup:self
+	pm.usesAutogroupSelf = pm.pol.usesAutogroupSelf()
+
+	var filter []tailcfg.FilterRule
+
+	var err error
+
+	// Standard compilation for all policies
+	filter, err = pm.pol.compileFilterRules(pm.users, pm.nodes)
 	if err != nil {
 		return false, fmt.Errorf("compiling filter rules: %w", err)
 	}
@@ -218,6 +233,35 @@ func (pm *PolicyManager) Filter() ([]tailcfg.FilterRule, []matcher.Match) {
 	return pm.filter, pm.matchers
 }
 
+// FilterForNode returns the filter rules for a specific node.
+// If the policy uses autogroup:self, this returns node-specific rules for security.
+// Otherwise, it returns the global filter rules for efficiency.
+func (pm *PolicyManager) FilterForNode(node types.NodeView) ([]tailcfg.FilterRule, error) {
+	if pm == nil {
+		return nil, nil
+	}
+
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+
+	if !pm.usesAutogroupSelf {
+		return pm.filter, nil
+	}
+
+	if rules, ok := pm.filterRulesMap[node.ID()]; ok {
+		return rules, nil
+	}
+
+	rules, err := pm.pol.compileFilterRulesForNode(pm.users, node, pm.nodes)
+	if err != nil {
+		return nil, fmt.Errorf("compiling filter rules for node: %w", err)
+	}
+
+	pm.filterRulesMap[node.ID()] = rules
+
+	return rules, nil
+}
+
 // SetUsers updates the users in the policy manager and updates the filter rules.
 func (pm *PolicyManager) SetUsers(users []types.User) (bool, error) {
 	if pm == nil {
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -251,6 +251,8 @@ working in v1 and not tested might be broken in v2 (and vice versa).
   [#2438](https://github.com/juanfont/headscale/pull/2438)
 - Add documentation for routes
   [#2496](https://github.com/juanfont/headscale/pull/2496)
+- Add support for `autogroup:self`
+  [#2789](https://github.com/juanfont/headscale/pull/2789)
```
nblock: Can you please move this into the `Changes` section of the upcoming release?

docs/ref/acls.md
```diff
@@ -194,13 +194,95 @@ Here are the ACL's to implement the same permissions as above:
...
+}
+```
+
+## Autogroups
+
+Headscale supports several autogroups that automatically include users, destinations, or devices with specific properties. Autogroups provide a convenient way to write ACL rules without manually listing individual users or devices.
+
+### `autogroup:internet`
+
+Allows access to the internet through [exit nodes](./routes.md#advertise-an-exit-node). Can only be used in ACL destinations.
```
nblock: The anchor is invalid.

```suggestion
Allows access to the internet through [exit nodes](routes.md#exit-node). Can only be used in ACL destinations.
```

docs/ref/acls.md
```diff
@@ -194,13 +194,95 @@ Here are the ACL's to implement the same permissions as above:
...
+  "src": ["group:users"],
+  "dst": ["autogroup:internet:*"]
+}
+```
+
+### `autogroup:member`
+
+Includes all users who are direct members of the tailnet. Does not include users from shared devices.
+
+Available as of [#2572](https://github.com/juanfont/headscale/pull/2572).
```
nblock: This might get stale over time if the feature is fixed, rewritten or updated in some way. In my opinion the sentence `Available as of ...` should be removed.

docs/ref/acls.md
```diff
@@ -194,13 +194,95 @@ Here are the ACL's to implement the same permissions as above:
...
+  "src": ["autogroup:member"],
+  "dst": ["tag:prod-app-servers:80,443"]
+}
+```
+
+### `autogroup:tagged`
+
+Includes all devices that have at least one tag.
+
+Available as of [#2572](https://github.com/juanfont/headscale/pull/2572).
```
nblock: See above.

docs/ref/acls.md
```diff
@@ -194,13 +194,95 @@ Here are the ACL's to implement the same permissions as above:
...
+  "src": ["autogroup:tagged"],
+  "dst": ["tag:monitoring:9090"]
+}
+```
+
+### `autogroup:self`
+
+Includes devices where the same user is authenticated on both the source and destination. Does not include tagged devices. Can only be used in ACL destinations.
+
+Available as of [#2789](https://github.com/juanfont/headscale/pull/2789).
```
nblock: See above.

docs/ref/acls.md
```diff
@@ -194,13 +194,95 @@ Here are the ACL's to implement the same permissions as above:
...
+{
+  "action": "accept",
+  "src": ["autogroup:member"],
+  "dst": ["autogroup:self:*"]
+}
+```
+
+*Using `autogroup:self` may cause performance degradation on the Headscale coordinator server in large deployments, as filter rules must be compiled per-node rather than globally and the current implementation is not very efficient.*
+
+if you experience performance issues, consider using more specific ACL rules or limiting the use of `autogroup:self`.
```
nblock: A mkdocs admonition is useful here. It renders a warning box.  Other changes:

- If
- Wrap rules in `{` and `}` for consistency with other code blocks.


```suggestion
!!! warning "The current implementation of `autogroup:self` is inefficient"

    Using `autogroup:self` may cause performance degradation on the Headscale coordinator server in large deployments,
    as filter rules must be compiled per-node rather than globally and the current implementation is not very efficient.

If you experience performance issues, consider using more specific ACL rules or limiting the use of `autogroup:self`.
```json
{
    // To allow internal users communications to their own nodes we can do following rules to allow access  in case autogroup:self is causing performance issues.
    { "action": "accept", "src": ["boss@"], "dst": ["boss@:*"] },
    { "action": "accept", "src": ["dev1@"], "dst": ["dev1@:*"] },
    { "action": "accept", "src": ["dev2@"], "dst": ["dev2@:*"] },
    { "action": "accept", "src": ["admin1@"], "dst": ["admin1@:*"] },
    { "action": "accept", "src": ["intern1@"], "dst": ["intern1@:*"] }
}
```
```

**Review by nblock — CHANGES_REQUESTED** [review]
Thank you very much for tackling `autogroup:self` and also documenting it!

Left a few minor comments and suggestions.

File: hscontrol/policy/v2/policy.go [hunk 7]
```diff
@@ -480,31 +458,112 @@ func (pm *PolicyManager) DebugString() string {
...
 
-// findAffectedUsers returns a set of user IDs whose nodes have changed between old and new node lists.
-func (pm *PolicyManager) findAffectedUsers(oldNodes, newNodes views.Slice[types.NodeView]) map[uint]struct{} {
-	affectedUsers := make(map[uint]struct{})
-
-	oldNodeMap := make(map[types.NodeID]uint)
+// invalidateAutogroupSelfCache intelligently clears only the cache entries that need to be
+// invalidated when using autogroup:self policies. This is much more efficient than clearing
+// the entire cache.
+func (pm *PolicyManager) invalidateAutogroupSelfCache(oldNodes, newNodes views.Slice[types.NodeView]) {
```
kradalby, line 464: Not sure how easy, but would be nice to have some tests for this function

kradalby (commented):
One comment, otherwise I'm happy with letting this in to let it be tested. 

I think we should write up in the changelog that this is an experiment and that you as a user needs to be careful and help test it and that they must understand that a bug might give more access than what is expected.
And a note about that it is quite computational expensive.

hscontrol/policy/v2/policy.go
vdovhanych: Shouldn't be that hard, i'll write a test for this functionality.

CHANGELOG.md
vdovhanych: fixed

docs/ref/acls.md
vdovhanych: fixed

docs/ref/acls.md
vdovhanych: removed the line completely, not really needed there.

docs/ref/acls.md
vdovhanych: removed the line completely, not really needed there.

docs/ref/acls.md
vdovhanych: removed the line completely, not really needed there.

docs/ref/acls.md
vdovhanych: fixed

hscontrol/policy/v2/policy.go
vdovhanych: added here - https://github.com/juanfont/headscale/pull/2789/files#diff-d7280ebf09dc40885a78b7afc790f5a527807b1407c0f0460a18a566680cb5faR70

benley:
I've been looking forward to this one, thank you for making it happen!

vdovhanych:
> I've been looking forward to this one, thank you for making it happen!

Same here @benley 😆 
Because of that i decided to take up on the challenge of building this. Hopefully it works well.

danscarfe:
Hip, hip hooray. Thanks all!

Haarolean:
@vdovhanych thanks for implementing this! ❤️

PR #2804, opened 2025-10-18
Expose pending interactive registrations via admin API

### LINKED ISSUE #2736: [Feature] Expose pending interactive registrations via admin API/CLI
### Use case

Use cases
- An admin-side service or UI that:
  - lists pending interactive registrations
  - approves a subset based on an allow/deny list, or after human review
- Automation using a periodic job that approves known devices, while still using registrationId.

### Description

Background
- Since the registrationId flow was introduced (replacing mkey-based approval for interactive registration), pending registrations are short‑lived and tracked in memory until approved (AuthURL follow-up).
- This decouples identity from machine keys and fixes multi-user/same-machine cases, which is great.

Problem today
- For interactive registrations there’s no way for an external controller to see “what’s pending” and act on it.
- The pending registrations live only in the in-memory registration cache:
- As a result, an external approval service cannot discover pending registrationIds to approve them, and polling the DB won’t help (not persisted).

Is exposing pending interactive registrations acceptable within headscale’s design?

### Contribution

_No response_

oneingan:
Closes #2736

## Overview

Initial draft PR to expose pending interactive registrations via the admin API (gRPC endpoint). This enables external services to list pending registration IDs and details for approval workflows. CLI support is not yet included—feedback welcome on whether to add it here.

Not ready for merge; seeking early feedback on design and implementation.

## Key Changes

- **gRPC API**: Added `ListPendingRegistrations` RPC to retrieve pending registrations from in-memory cache (ID, hostname, machine/node keys, expiry).
- **State Management**: New `ListPendingRegistrations` method in `hscontrol/state` to snapshot cache entries as `PendingRegistration` structs.
- **Protobuf**: Added `PendingRegistration` message and related request/response types in `proto/headscale/v1/headscale.proto`.
- **Helpers**: New `hscontrol/types/pending_registration.go` for proto construction.
- Generated files updated accordingly.

No persistence changes; aligns with in-memory cache design.

## Open Questions

- Add CLI commands (e.g., `headscale pending list` / `approve <id>`)?
- Filtering options (e.g., by expiry, hostname)?
- Auth for new endpoint?
- Testing/docs TODO.

Review diff and suggest improvements!

### DIFF
--- a/gen/go/headscale/v1/headscale_grpc.pb.go
+++ b/gen/go/headscale/v1/headscale_grpc.pb.go
@@ -19,31 +19,32 @@ import (
 const _ = grpc.SupportPackageIsVersion9
 
 const (
-	HeadscaleService_CreateUser_FullMethodName        = "/headscale.v1.HeadscaleService/CreateUser"
-	HeadscaleService_RenameUser_FullMethodName        = "/headscale.v1.HeadscaleService/RenameUser"
-	HeadscaleService_DeleteUser_FullMethodName        = "/headscale.v1.HeadscaleService/DeleteUser"
-	HeadscaleService_ListUsers_FullMethodName         = "/headscale.v1.HeadscaleService/ListUsers"
-	HeadscaleService_CreatePreAuthKey_FullMethodName  = "/headscale.v1.HeadscaleService/CreatePreAuthKey"
-	HeadscaleService_ExpirePreAuthKey_FullMethodName  = "/headscale.v1.HeadscaleService/ExpirePreAuthKey"
-	HeadscaleService_ListPreAuthKeys_FullMethodName   = "/headscale.v1.HeadscaleService/ListPreAuthKeys"
-	HeadscaleService_DebugCreateNode_FullMethodName   = "/headscale.v1.HeadscaleService/DebugCreateNode"
-	HeadscaleService_GetNode_FullMethodName           = "/headscale.v1.HeadscaleService/GetNode"
-	HeadscaleService_SetTags_FullMethodName           = "/headscale.v1.HeadscaleService/SetTags"
-	HeadscaleService_SetApprovedRoutes_FullMethodName = "/headscale.v1.HeadscaleService/SetApprovedRoutes"
-	HeadscaleService_RegisterNode_FullMethodName      = "/headscale.v1.HeadscaleService/RegisterNode"
-	HeadscaleService_DeleteNode_FullMethodName        = "/headscale.v1.HeadscaleService/DeleteNode"
-	HeadscaleService_ExpireNode_FullMethodName        = "/headscale.v1.HeadscaleService/ExpireNode"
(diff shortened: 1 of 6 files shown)


oneingan:
HI, are you interested in moving this forward?

kradalby:
I am hesitant to add this as I am not keen on the overhead of adding essentially another registration flow that we need to maintain and test. I think that using the registration flow for automatically joining nodes is the wrong approach and that this should be solved using auth keys instead. These can be issued from the API, and a flow could be built around that to approve automatically or via human interactions. 

Exposing the registration cache is exposing internal state which has an API that we likely want to retain the ability to change as we please and not expose.

MulverineX:
@kradalby that requires creation of a custom sign-in flow, when this would work out of the box with existing headscale and tailscale clients. Most hobbiest users of headscale just want to approve a pending node when their friend is trying to connect.

oneingan:
Hi @kradalby — thanks for the feedback. Understood on not wanting to expose internal cache state or take on extra registration-flow surface area.

Just to clarify where I’m coming from: I don’t work on Clan, I’m only a user — but Clan’s networking abstraction is the real motivation behind this request. It’s a good example of an orchestration model that tries to make networking “just work” by integrating multiple network providers and driving them via external tooling/policy.

If you have a moment, this write-up gives the context better than I can paraphrase:
https://clan.lol/blog/2025-wrap-up/
https://docs.clan.lol/main/guides/networking/networking/

I’m not attached to the exact mechanism I proposed here; the goal is simply to make it possible for an external controller/CLI to build a reliable approval UX around the existing interactive join behavior. If you have a preferred headscale-native solution (auth key workflow, hooks/events, etc.), I’d really appreciate your guidance and I can adjust or drop this PR accordingly.

oneingan:
You might want to look at Biscuit as a good fit for the “external controller / approval UX” part of this discussion.

Biscuit is an authorization token format designed for capability-style workflows (delegation/attenuation, expiry, policy constraints), with decentralized verification and offline attenuation. That seems aligned with the idea of issuing something from the admin API that a controller can refine and hand off, without needing to expose the registration cache as a stable API surface.

Link: https://www.biscuitsec.org/

If that direction makes sense, a possible headscale-native approach could be: issue an API-created token (auth key or similar) and let the external orchestrator attenuate it (scope/expiry/constraints) to drive a friendly “approve/join” flow, rather than listing pending registrations directly.

oneingan:
Closing as obsolete; this branch is no longer needed.

MulverineX:
@oneingan how is this obsolete? Is there now a way to approve pending interactive registrations via the API without using preauth key nonsense?

I want to give the exact same instructions to everyone on how to connect to my tailnet, and when they attempt to, I want to have a system that gives me a notification so I can confirm with them that it is their device so I can approve it. Anything else would be way too much of a hassle or too insecure.

PR #2810, opened 2025-10-21
[Fix] Headscale API Bypasses Middleware

### LINKED ISSUE #2809: [Bug] Headscale API Bypasses Middleware Check

### Current Behavior

In the latest build of headscale (main branch, commit 8becb7e54a6b565013aeff2feee5e5d639c564aa), Headscale APIs bypasses middleware check and returns sensitive data even if the API caller does not provide any API key at all. Provided below are screenshots from my Postman and my Headscale logs:

Image

### Expected Behavior

If an API caller does not provide any API key, Headscale should return 401 Unauthorized without returning any sensitive data. The request should be terminated inside the auth middleware.

### Steps To Reproduce

1. Clone and build the latest change from Headscale repository.
2. Setup a simple configuration to make sure Headscale can run.
3. Create some mock users, e.g: `headscale users create elyasasmad`.
4. Run `headscale serve`.
5. Run `curl https://example.com/api/v1/user`.
6. Headscale users are returned after the `Unauthorized` text at the beginning of the response.

### Environment

```markdown
- OS: Ubuntu 25.04
- Headscale version: Commit 8becb7e54a6b565013aeff2feee5e5d639c564aa (v0.27.0-beta.1 shows the same behavior)
- Tailscale version: 1.88.3
- Number of nodes: 14
```

### Runtime environment

### Debug information

Image

ElyasAsmad:
# Description
This PR aims to fix #2809 

# General Overview
The previous implementation checks if `err` is not null and returned from the `writer.Write` function call. In majority cases, the `Write` function call will be triggered successfully without any error. In an unauthorized scenario (where the API caller does not provide API key), Headscale already know the API caller is unauthorized but since the guard `return` statement is ONLY called if `writer.Write` returns an error, thus the middleware will continue executing.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -380,53 +380,45 @@ func (h *Headscale) httpAuthenticationMiddleware(next http.Handler) http.Handler
 		writer http.ResponseWriter,
 		req *http.Request,
 	) {
-		if err := func() error {
-			log.Trace().
+		log.Trace().
+			Caller().
+			Str("client_address", req.RemoteAddr).
+			Msg("HTTP authentication invoked")
+
+		authHeader := req.Header.Get("Authorization")
+
+		writeUnauthorized := func(statusCode int) {
+			writer.WriteHeader(statusCode)
+			if _, err := writer.Write([]byte("Unauthorized")); err != nil {
+				log.Error().Err(err).Msg("writing HTTP response failed")
+			}
+		}
+
+		if !strings.HasPrefix(authHeader, AuthPrefix) {
+			log.Error().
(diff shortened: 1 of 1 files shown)


kradalby:
This is a good catch! thank you!, I will pick up your branch in a bit, because I first want to add tests to validate this, then apply your fix to ensure we do not have a regression like this again. I might end up cherry-picking your change, but we willl see.

kradalby (approved):

PR #2831, opened 2025-10-28
remove unused ips from alloc ipset

bobelev:
With IPAllocationStrategy = "random" and strange environments (small VPN CIDR + k8s pods) users can run out of available IPs. So let's remove ips from ipset on DeleteNode.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Changes
 
+- Reclaim IPs from the IP allocator when nodes are deleted
+  [#2831](https://github.com/juanfont/headscale/pull/2831)
+
 ## 0.27.1 (2025-11-11)
 
 **Minimum supported Tailscale client version: v1.64.0**


hscontrol/db/ip.go
```diff
@@ -337,3 +337,12 @@ func (db *HSDatabase) BackfillNodeIPs(i *IPAllocator) ([]string, error) {
 
 	return ret, err
 }
+
+func (i *IPAllocator) RemoveUsedIps(ips []netip.Addr) {
```
kradalby: I had to read this a few times for it to make sense to me. 

We are not really "removing" it, we are kind of adding it back, or freeing it.

Maybe:

```suggestion
func (i *IPAllocator) FreeIPs(ips []netip.Addr) {
```

Or very technical, maybe `DeallocateIPs`

kradalby:
adding a next changelog would be great.

PR #2832, opened 2025-10-28
templates: migrate OIDC callback to elem-go, give an overhaul

### LINKED ISSUE #2115: [Improvement] use go-elem for HTML pages
### Use case

We currently have a couple of HTML files/template strings in the code using gotemplate and handcrafted HTML, its not a lot, and its fine, but a nice QoL improvement would be to make them nice, typed and generated.

### Description

I have had quite a good experience with [go-elem](https://github.com/chasefleming/elem-go) for this for some other projects, ergonomics are pretty good for writing HTML as functions, at least if we consider that it is Go and not something like Elm, Roc or Scala which might allow a bit nicer funcs.

### Contribution

I will eventually do this, but if someone is up for converting the HTML to go-elem, that would be a great PR/starter issue, make things neat and reusable.

kradalby:
apple_desktop
oidc_callback_desktop

Closes #2115 

claude was used in this PR.


hscontrol/assets/style.css
```diff
@@ -0,0 +1,145 @@
+/* CSS Variables from Material for MkDocs */
```
nblock, line 1: NIT: `style.css` and `headscale.svg` should probably be moved to the `hscontrol/assets/` directory or `favicon.ico` should be moved into `hscontrol/templates/`.

nblock (approved):

Issue #2655, opened 2025-06-24
[Bug] Inconsistent handling of OIDC claims


nblock:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Recent headscale versions fetch OIDC claims from the ID token and also query the userinfo endpoint to gather details about the user. This is nice as identity providers seem to reduce the size of tokens lately.

But user authorization [happens right before the userinfo endpoint is queried](https://github.com/juanfont/headscale/blob/081af2674b9d14f9db06528f4be896bb82752cba/hscontrol/oidc.go#L280) which means that authorization only uses the claims from the ID token. Thus, the data used for authorization and storage might be different, affected config settings:

- `oidc.allowed_domains`
- `oidc.allowed_users`
- `allowed_groups`

Also the handling of the `email_verified` is inconsistent:

* Authorization: the `email_verified` claim is not checked. The check for `oidc.allowed_domains` and `oidc.allowed_users` only considers `email`
* Storage: the e-mail address is only considered as valid if `email_verified: true`

### Expected Behavior

Information retrieval from OIDC should be identical for all uses within the codebase.

### Steps To Reproduce

-

### Environment

```markdown
- OS: -
- Headscale version: 0.26.0
- Tailscale version: -
```

### Runtime environment

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

### Debug information

-

nblock:
One possible solution to solve the `email_verified` difference is to make it configurable whether `email_verified: true` is required or not. For example:

* `oidc.require_email_verified: true`: Process the `email_verified` claim and accept `email` only if `email_verified: true`
* `oidc.require_email_verified: false`: Accept `email` regardless of `email_verified`

fredrikekre:
Since the userinfo endpoint is queried unconditionally anyway, couldn't that query simply be moved up to before the `validateOIDCAllowed*` calls?

> Maybe the email_verified: true claim is no longer that important?

I think it still is, at least if you are filtering users based on email or domain?

james-d-elliott:
Please be aware I'm not aware of the particular implementation at play here so some of these comments may not be applicable.

[...]

PR #2860, opened 2025-11-03
Configurable email_verified OIDC Claim Requirement

ImpostorKeanu:
- [x] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

---

When configured to use OIDC, `email_verified=true` is required to appear in ID token claims or user info before the email will be saved in a user's information. There are some scenarios where the `email_verified` claim cannot be sent/configured, such as when using [Cloudflare One-time pin][cloudflare-otp] -- which doesn't support adding/modifying claims. This can make writing ACLs difficult or impossible.

This PR adds a configuration that relaxes the  `email_verified=true` requirement by:

- adding a configuration field: `oidc.use_unverified_email`
- considering the setting when:
  - applying `oidc.allowed_domains` during authorization
  - user records are initialized via `User.FromClaim`

Related: #2655 

[cloudflare-otp]: https://developers.cloudflare.com/cloudflare-one/integrations/identity-providers/one-time-pin/

### DIFF
--- a/hscontrol/mapper/batcher_lockfree.go
+++ b/hscontrol/mapper/batcher_lockfree.go
@@ -31,6 +31,7 @@ type LockFreeBatcher struct {
 	workCh     chan work
 	workChOnce sync.Once // Ensures workCh is only closed once
 	done       chan struct{}
+	doneOnce   sync.Once // Ensures done is only closed once
 
 	// Batching state
 	pendingChanges *xsync.Map[types.NodeID, []change.Change]
@@ -151,10 +152,12 @@ func (b *LockFreeBatcher) Start() {
 }
 
 func (b *LockFreeBatcher) Close() {
-	// Signal shutdown to all goroutines
-	if b.done != nil {
-		close(b.done)
-	}
+	// Signal shutdown to all goroutines, only once
+	b.doneOnce.Do(func() {
+		if b.done != nil {
+			close(b.done)
+		}
+	})
 
 	// Only close workCh once using sync.Once to prevent races
 	b.workChOnce.Do(func() {


ImpostorKeanu:
Just checking in on this. Please let me know if I've gone about the process wrong.

Think it's something we're interested in accepting?

Really appreciate everything.

nblock:
> Really appreciate everything.

Thx for your PR! I'll do some testing once 0.27.1 is out and we'll come back to this afterwards.

ImpostorKeanu:
No rush at all! Thanks a ton!

config-example.yaml
```diff
@@ -361,6 +361,17 @@ unix_socket_permission: "0770"
...
+#   # Enable this setting to accept the user's email address regardless
+#   # if "email_verified: true" is sent by identity provider.
+#   #
+#   # By default, "email_verified: true" must appear in claims or user info
+#   # before Headscale will accept the principal's email address as the user
+#   # account is created after successful authentication.
+#   #
+#   # This setting is useful when claims and their mapping can't be controlled,
+#   # such as when using Cloudflare One-time pin for authentication.
+#   use_unverified_email: false
```
nblock: Can we use `email_verified_required` as setting name and invert the logic? It make its relationship to `email_verified` clearer. Thus: `email_verified_required: true`

hscontrol/types/config.go
```diff
@@ -385,10 +393,18 @@ func validateServerConfig() error {
 		}
 	}
 
+	if viper.IsSet("oidc.use_unverified_email") {
+		log.Warn().
+			Msg("unverified emails will be accepted during oidc authentication (oidc.use_unverified_email=true)")
+	} else {
+		log.Warn().Msg("only verified emails will be accepted during oidc authentication (oidc.use_unverified_email=false)")
+	}
+
```
nblock: Can you remove this block entirely as it prints a warning for static configuration that does not change during runtime?

hscontrol/oidc.go
```diff
@@ -281,11 +285,13 @@ func (a *AuthProviderOIDC) OIDCCallbackHandler(
...
 	}
 
-	// The user claims are now updated from the userinfo endpoint so we can verify the user
-	// against allowed emails, email domains, and groups.
-	if err := validateOIDCAllowedDomains(a.cfg.AllowedDomains, &claims); err != nil {
-		httpError(writer, err)
-		return
+	if bool(claims.EmailVerified) || a.cfg.UseUnverifiedEmail {
+		// The user claims are now updated from the userinfo endpoint so we can verify the user
+		// against allowed emails, email domains, and groups.
```
nblock: I think this comment is meant as introduction describing the following authorization checks. It should probably be moved outside of the block.

hscontrol/oidc.go
```diff
@@ -281,11 +285,13 @@ func (a *AuthProviderOIDC) OIDCCallbackHandler(
...
+		if err := validateOIDCAllowedDomains(a.cfg.AllowedDomains, &claims); err != nil {
```
nblock: The same rules should be applied to the `AllowedUsers` configuration.

nblock (changes requested):
Thanks for your PR!

The following behavior makes authorization and profile update consistent:

For `oidc.email_verified_required: true` (default), require the e-mail to be verified to do any processing on it (`email_verified: true`). This gives as two cases:

- `email_verified: true`:
    - accept if the email matches the `oidc.allowed_users` filter
    - accept if the domain matches the `oidc.allowed_domains` filter
    - update the email in the user's profile
- `email_verified: false` (or not provided):
    - reject even if the email would match the `oidc.allowed_users` filter (breaking)
    - reject even if the domain would match the `oidc.allowed_domains` filter (breaking)
    - don't update the email in the user's profile

For `oidc.email_verified_required: false`, ignore the `email_verified` flag during processing. This results in:

- accept if the email matches the `oidc.allowed_users` filter
- accept if the domain matches the `oidcs.allowed_groups` filter
- update the email in the user's profile

ImpostorKeanu:
So sorry for the delay on this! Got wrapped up in work and the holiday.

I'll try to knock it out later today.

Thanks!

config-example.yaml
ImpostorKeanu: Great recommendation. I always struggle with keeping the bigger picture in mind when naming something.

ImpostorKeanu:
I fumbled the tests for a few minutes but everything should be in order now.

I'm not sure why the test workflow says it's waiting for a "status to be reported" for the tests. Weird.

If I've borked stuff, let me know and I'll fix it.

nblock:
Thanks for following up and implementing most of https://github.com/juanfont/headscale/pull/2860#pullrequestreview-3470236944. It seems that the following cases are not implemented?

- `email_verified_required: true`, `email_verified: false` (or not provided):
    - reject even if the email would match the `oidc.allowed_users` filter (breaking)
    - reject even if the domain would match the `oidc.allowed_domains` filter (breaking)
    - don't update the email in the user's profile

I tested it both cases ( `oidc.allowed_users`, `oidc.allowed_domains`) and noticed that a user with an unverified email address was able to authenticate.

Could you have a look at this?

ImpostorKeanu:
> Thanks for following up and implementing most of [#2860 (review)](https://github.com/juanfont/headscale/pull/2860#pullrequestreview-3470236944). It seems that the following cases are not implemented?
> 
>     * `email_verified_required: true`, `email_verified: false` (or not provided):
>       
>       * reject even if the email would match the `oidc.allowed_users` filter (breaking)
>       * reject even if the domain would match the `oidc.allowed_domains` filter (breaking)
>       * don't update the email in the user's profile
> 
> 
> I tested it both cases ( `oidc.allowed_users`, `oidc.allowed_domains`) and noticed that a user with an unverified email address was able to authenticate.
> 
> Could you have a look at this?

Checking into it but I have a question:

As of now, [OIDC authorization checks][checks] are AND'd. Shouldn't they be OR'd?

Example:

If both `AllowedDomains` and `AllowedGroups` are configured and `claims.Email` doesn't have an allowed domain, `AllowedGroups` isn't evaluated.

[checks]: https://github.com/juanfont/headscale/blob/f3f2d30004131064bba9aebb7cd60e269decef2c/hscontrol/oidc.go#L272

ImpostorKeanu:
I refactored to OR authorization checks, cleaned stuff up, and added standalone tests for OIDC authorization. Some of the decisions are questionable. All tests are succeeding on my side. Still waiting to see if GitHub actions wreck me.

I'm open to changes/rejection. Feel free to share your thoughts, and sorry if I'm making this more complicated than it has to be.

nblock:
> As of now, [OIDC authorization checks](https://github.com/juanfont/headscale/blob/f3f2d30004131064bba9aebb7cd60e269decef2c/hscontrol/oidc.go#L272) are AND'd. Shouldn't they be OR'd?

This could be discussed as a possible follow-up to this PR, but let's don't change the current behavior in this PR and keep it focused ([docs](https://headscale.net/development/ref/oidc/#authorize-users-with-filters)).

(Feel free to rebase and squash as single commit.)

ImpostorKeanu:
> > As of now, [OIDC authorization checks](https://github.com/juanfont/headscale/blob/f3f2d30004131064bba9aebb7cd60e269decef2c/hscontrol/oidc.go#L272) are AND'd. Shouldn't they be OR'd?
> 
> This could be discussed as a possible follow-up to this PR, but let's don't change the current behavior in this PR and keep it focused ([docs](https://headscale.net/development/ref/oidc/#authorize-users-with-filters)).
> 
> (Feel free to rebase and squash as single commit.)

Ok, I reset to an earlier commit and saved those changes to another branch. I'll open another PR.

Latest push still moves authorization to a distinct function for testing while preserving the authorization flow.

nblock:
Thx, will have a look soonish

hscontrol/oidc.go
```diff
@@ -453,14 +475,57 @@ func validateOIDCAllowedUsers(
...
+		if err := validateOIDCAllowedGroups(cfg.AllowedGroups, claims); err != nil {
+			return err
+		}
+	}
+
+	trustEmail := !cfg.EmailVerifiedRequired || bool(claims.EmailVerified)
+	hasEmailTests := len(cfg.AllowedDomains) > 0 || len(cfg.AllowedUsers) > 0
+	if !trustEmail && hasEmailTests {
+		return errOIDCUnverifiedEmail
+	}
```
nblock, line 481: An "internal server error" is raised when:

- `oidc.allowed_users` is configured
- `oidc.email_verified_enabled` is enabled
- A user with an unverified e-mail address tries to authenticate.

Logs: ` ERR http internal server error error="authenticated principal has an unverified email" code=500`

Browser: ` internal server error`

This should raise a http/401 instead, similar to: https://github.com/juanfont/headscale/blob/21ba197d06ab35acda560317fa831aa8cf1d1507/hscontrol/oidc.go#L422 https://github.com/juanfont/headscale/blob/21ba197d06ab35acda560317fa831aa8cf1d1507/hscontrol/oidc.go#L444 https://github.com/juanfont/headscale/blob/21ba197d06ab35acda560317fa831aa8cf1d1507/hscontrol/oidc.go#L458

nblock (approved):
Thx, did a quick rundown with the following testcases, looks good:

- [X] evr unset: defaults to true?
- [X] evr=true, ev=true: ack + email in profile
- [X] evr=true, ev=false: ack + email not in profile
- [X] evr=true, ev=true, oidc.allowed_users
    - [X] match+email in profile
    - [X] no match: reject?
- [X] evr=true, ev=true, oidc.allowed_domains
    - [X] match+email in profile
    - [X] no match: reject?
- [X] evr=true, ev=false, oidc.allowed_users
    - [X] always reject and never update profile?
- [X] evr=true, ev=false, oidc.allowed_domains
    - [X] always reject and never update profile?
- [X] evr=false, oidc.allowed_users
    - [X] accept if match + profile update
- [X] evr=false, oidc.allowed_domains
    - [X] accept if match + profile update

@kradalby will have another look at the code.

PR #2878, opened 2025-11-11
changelog: prepare for 0.27.1

kradalby:
claude was used in this pr


nblock (approved):
Probably bump the version in `mkdocs.yml`

Issue #2838, opened 2025-10-30
[Bug] autogroup:self will cover other ACL rules


hanzec:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When following ACL rules applied

```
    {
      "action": "accept",
      "src": ["autogroup:member"],
      "dst": ["autogroup:self:*"]
    },
    {
      "action": "accept",
      "src": ["group:home"],
      "dst": ["192.168.2.0/24:*",
                  "tag:router:*"]
    }
```

running `tailscale status` on nodes register by user in group:home only return current user's node. 

### Expected Behavior

running `tailscale status` should return following nodes:

- current user's node
- nodes with tag:route

### Steps To Reproduce

inserting above to ACL Rules, `tailscale status` only return current user's node. 
removing autogroup rules, `tailscale status` return nodes with tag:router 

### Environment

```markdown
- OS: client node running MacOS 26.0.1, router node using OPNSense
- Headscale version: 0.27.0
- Tailscale version: client 1.91.26, router 1.88.3
```

### Runtime environment

- [x] Headscale is behind a (reverse) proxy
- [x] Headscale runs in a container

### Debug information

full configuration log:

```
{
  "groups": {
    "group:infer-admin": ["test-1@example.com"],
    "group:home": ["test-1@example.com", "test-2@example.com"]
  },

  "tagOwners": {
    "tag:home": ["group:home"],
    "tag:home-router": ["group:infer-admin"]
  },

  "acls": [
    {
      "action": "accept",
      "src": ["sa-headplane@example.com"],
      "dst": ["*:*"]
    },
    {
      "action": "accept",
      "src": ["autogroup:member"],
      "dst": ["autogroup:self:*"]
    },
    {
      "action": "accept",
      "src": ["group:home"],
      "dst": ["192.168.2.0/24:*",
              "tag:home-router:*"]
    }
  ]
}
```

vdovhanych:
I just tested this with a quick test file i wrote with help of Cursor and i can't seem to replicate this issue in the tests, i used the exact acl you provided for the test and this was the outcome. 

**Test Configuration**
- User1 has 2 devices (laptop + phone)
- Router owned by router-owner (different user)
- Router tagged with tag:home-router

**ACL allows:**
- autogroup:member -> autogroup:self (same-user connectivity)
- group:home -> tag:home-router (cross-user connectivity to tagged router)

**Results**
```
[...]

PR #2882, opened 2025-11-12
chore: fix filterHash to work with autogroup:self in the acls

vdovhanych:
- [x] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

This attempts to fix additional issue mentioned here - https://github.com/juanfont/headscale/issues/2838#issuecomment-3518891097

### DIFF
--- a/hscontrol/policy/v2/policy.go
+++ b/hscontrol/policy/v2/policy.go
@@ -47,6 +47,14 @@ type PolicyManager struct {
 	usesAutogroupSelf bool
 }
 
+// filterAndPolicy combines the compiled filter rules with policy content for hashing.
+// This ensures filterHash changes when policy changes, even for autogroup:self where
+// the compiled filter is always empty.
+type filterAndPolicy struct {
+	Filter []tailcfg.FilterRule
+	Policy *Policy
+}
+
 // NewPolicyManager creates a new PolicyManager from a policy file and a list of users and nodes.
 // It returns an error if the policy file is invalid.
 // The policy manager will update the filter rules based on the users and nodes.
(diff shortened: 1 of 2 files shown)


hscontrol/policy/v2/policy.go
```diff
@@ -98,7 +90,39 @@ func (pm *PolicyManager) updateLocked() (bool, error) {
...
+		TagOwners     TagOwners
+		ACLs          []ACL
+		AutoApprovers AutoApproverPolicy
+		SSHs          []SSH
+	}
+	type filterAndPolicy struct {
+		Filter []tailcfg.FilterRule
+		Policy hashablePolicy
+	}
+	var hashablePol hashablePolicy
```
kradalby: Why do we need to shadow the policy with a new type?

hscontrol/policy/v2/policy.go
```diff
@@ -98,7 +90,39 @@ func (pm *PolicyManager) updateLocked() (bool, error) {
...
+	// like 'validated' which shouldn't affect policy comparison.
+	type hashablePolicy struct {
+		Groups        Groups
+		Hosts         Hosts
+		TagOwners     TagOwners
+		ACLs          []ACL
+		AutoApprovers AutoApproverPolicy
+		SSHs          []SSH
+	}
+	type filterAndPolicy struct {
```
kradalby: lets move the types out of the function, they can still be unexported.

hscontrol/policy/v2/policy.go
vdovhanych: Oh yeah, you're right, it should work with simply using the pm.pol i will try it and run the tests if that works.

vdovhanych:
I think this should be working now.

PR #2885, opened 2025-11-13
make tags first class node owner

kradalby:
This PR changes tags to be something that exists on nodes in addition to users, to being its own thing. It is part of moving our tags support towards the correct tailscale compatible implementation.

There are probably rough edges in this PR, but the intention is to get it in, and then start fixing bugs from 0.28.0 milestone (long standing tags issue) to discover what works and what doesnt.

Updates #2417
Closes #2619

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -57,7 +57,6 @@ jobs:
           - TestNodeCommand
           - TestNodeExpireCommand
           - TestNodeRenameCommand
-          - TestNodeMoveCommand
           - TestPolicyCommand
           - TestPolicyBrokenConfigCommand
           - TestDERPVerifyEndpoint


CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ at creation time. When listing keys, only the prefix is shown (e.g.,
 `hskey-auth-{prefix}-{secret}`. Legacy plaintext keys continue to work for
 backwards compatibility.
 
+### Tags
+
+Tags are now implemented following the Tailscale model where tags and user ownership are mutually exclusive. Devices can be either user-owned (authenticated via web/OIDC) or tagged (authenticated via tagged PreAuthKeys). Tagged devices receive their identity from tags rather than users, making them suitable for servers and infrastructure. Applying a tag to a device removes user-based authentication. See the [Tailscale tags documentation](https://tailscale.com/kb/1068/tags) for details on how tags work.
```
kradalby, line 26: user devices can be authenticated via PreAuthKeys.

CHANGELOG.md
```diff
@@ -30,6 +34,8 @@ release.
 
 ### BREAKING
 
+- **Tags**: The gRPC `SetTags` endpoint now prevents setting tags on user-owned nodes. Only nodes registered with tagged PreAuthKeys can have tags modified via the API.
```
kradalby: This is incorrect, a user-owned node can be "converted" to a tagged node by tagging it. That will cause it to become a tagged node and not a user-owned node.

We need to ensure the implementation reflects this.

hscontrol/auth.go
```diff
@@ -234,18 +234,23 @@ func isAuthKey(req tailcfg.RegisterRequest) bool {
...
 
 		// Headscale does not implement the concept of machine authorization
 		// so we always return true here.
 		// Revisit this if #2176 gets implemented.
 		MachineAuthorized: true,
 	}
+
+	// For user-owned nodes, include User and Login information
+	// For tagged nodes, these fields should be empty as the node is owned by tags
+	if !node.IsTagged() && node.UserView().Valid() {
```
kradalby: There is a special user for tagged devices, should we use this here? called TaggedDevicesUserID and TaggedDevices in types/users.go

hscontrol/db/node.go
```diff
@@ -349,12 +351,20 @@ func RegisterNodeForTest(tx *gorm.DB, node types.Node, ipv4 *netip.Addr, ipv6 *n
 		panic("RegisterNodeForTest can only be called during tests")
 	}
 
-	log.Debug().
+	logEvent := log.Debug().
```
kradalby, line 354: I like this style, we should do it more to clean up very long loglines that make the code hard to read.

hscontrol/db/node.go
```diff
@@ -349,12 +351,20 @@ func RegisterNodeForTest(tx *gorm.DB, node types.Node, ipv4 *netip.Addr, ipv6 *n
...
 		Str("node", node.Hostname).
 		Str("machine_key", node.MachineKey.ShortString()).
-		Str("node_key", node.NodeKey.ShortString()).
-		Str("user", node.User.Username()).
-		Msg("Registering test node")
+		Str("node_key", node.NodeKey.ShortString())
+
+	if node.User != nil {
+		logEvent.Str("user", node.User.Username())
```
kradalby: I'm not sure if this gets persisted and needs to be saved to `logEvent`

hscontrol/db/preauth_keys.go
```diff
@@ -23,7 +23,7 @@ var (
 )
 
 func (hsdb *HSDatabase) CreatePreAuthKey(
-	uid types.UserID,
+	uid *types.UserID, // NOW OPTIONAL - can be nil for system-created tagged keys
```
kradalby: I dont like this comment, NOW OPTIONAL doesnt age well. the rest, commenting what can be nil is useful.

In principle, I would expect keys to still have an owner, so maybe this should not be nil. It is just that if the key is created with tag, the _key_ has an owner, but the node is tagged.

hscontrol/db/preauth_keys.go
```diff
@@ -149,14 +172,19 @@ func ListPreAuthKeysByUser(tx *gorm.DB, uid types.UserID) ([]types.PreAuthKey, e
...
 		return nil, err
 	}
 
 	return keys, nil
 }
 
-var ErrPreAuthKeyFailedToParse = errors.New("failed to parse AuthKey")
+var (
+	ErrPreAuthKeyFailedToParse    = errors.New("failed to parse AuthKey")
+	ErrPreAuthKeyNotTaggedOrOwned = errors.New("preauthkey must be either tagged or owned by user")
```
kradalby: I think we should style it `auth-key` anywhere user facing, so keep it the same in the code, variable, types and such, but any user facing messaging and docs should be `auth-key`, not:
- AuthKey
- PreAuthKey
- pre-auth-key
- preauthkey
- authkey
- preauth-key
- pre-authkey

and so on.

hscontrol/grpcv1.go
```diff
@@ -320,13 +320,42 @@ func (api headscaleV1APIServer) SetTags(
...
+	// Validate tag format
 	for _, tag := range request.GetTags() {
 		err := validateTag(tag)
 		if err != nil {
 			return nil, err
 		}
 	}
 
+	// IMPORTANT: Enforce User XOR Tags rule
+	// Cannot set tags on user-owned nodes - tags define ownership
```
kradalby: This is not correct, as mentioned in the readme, a user-owned node can be converted to a tagged node, but it cannot be converted back.

hscontrol/grpcv1.go
```diff
@@ -508,13 +537,20 @@ func nodesToProto(state *state.State, nodes views.Slice[types.NodeView]) []*v1.N
 	for index, node := range nodes.All() {
 		resp := node.Proto()
 
+		// IMPORTANT: Implement tags-as-identity model
```
kradalby: I dont like screaming comments, like IMPORTANT.

hscontrol/mapper/tail.go
```diff
@@ -99,7 +100,21 @@ func tailNode(
...
+				//nolint:gosec // G115: User IDs will never overflow int64
+				return tailcfg.UserID(int64(types.TaggedDevices.ID))
+			}
+
+			// User-owned nodes: use the actual user ID
+			// Note: UserID().Get() returns 0 if not set, which is fine for tests
+			// but in production, non-tagged nodes should always have a valid UserID
+			//nolint:gosec // G115: User IDs will never overflow int64
+			return tailcfg.UserID(int64(node.UserID().Get()))
+		}(),
```
kradalby: I think this should be a UserID func on the node/nodeview structure.

hscontrol/policy/v2/policy.go
```diff
@@ -664,41 +664,41 @@ func (pm *PolicyManager) invalidateAutogroupSelfCache(oldNodes, newNodes views.S
 	// Check for removed nodes
 	for nodeID, oldNode := range oldNodeMap {
 		if _, exists := newNodeMap[nodeID]; !exists {
-			affectedUsers[oldNode.User().ID] = struct{}{}
+			affectedUsers[oldNode.User().Model().ID] = struct{}{}
```
kradalby: On the UserView, we should have a ID() func that maps .Model().ID so we do not have to write it out in full everywhere. do it and replace all the `.Model().ID` calls.

hscontrol/state/node_store.go
```diff
@@ -405,7 +405,7 @@ func snapshotFromNodes(nodes map[types.NodeID]types.Node, peersFunc PeersFunc) S
 	// Build nodesByUser, nodesByNodeKey, and nodesByMachineKey maps
 	for _, n := range nodes {
 		nodeView := n.View()
-		userID := types.UserID(n.UserID)
+		userID := types.UserID(*n.UserID)
```
kradalby: We should have a typed UserID func on NodeView, like mentioned in tail.go comment.

hscontrol/types/node.go
```diff
@@ -196,55 +201,41 @@ func (node *Node) HasIP(i netip.Addr) bool {
...
-	if len(node.ForcedTags) > 0 {
+	// Check both node.Tags and AuthKey.Tags for backwards compatibility
+	// During migration, some nodes might have tags on PreAuthKey but not yet on node
+	if len(node.Tags) > 0 {
 		return true
 	}
 
 	if node.AuthKey != nil && len(node.AuthKey.Tags) > 0 {
 		return true
 	}
```
kradalby: This should not be the case anymore. Tags are added to Tags on node at registration time, and the authkey should not be relevant again after that point. So we need to remove this dependencies.

hscontrol/types/node.go
```diff
@@ -196,55 +201,41 @@ func (node *Node) HasIP(i netip.Addr) bool {
...
-	// and hostinfo.requestedtags.
-	// Do this in other work.
-	// #2417
-
-	tags = append(tags, node.ForcedTags...)
-	sort.Strings(tags)
-	tags = slices.Compact(tags)
+	if node.AuthKey != nil && slices.Contains(node.AuthKey.Tags, tag) {
+		return true
+	}
```
kradalby: Same as comment above, used tags should only exist on the node structure, so we need to remove this dependency. The authkey tags are only relevant at the time of authentication.

hscontrol/types/preauth_key.go
```diff
@@ -23,16 +23,19 @@ type PreAuthKey struct {
 	Prefix string
 	Hash   []byte // bcrypt
 
-	UserID    uint
-	User      User `gorm:"constraint:OnDelete:SET NULL;"`
+	// For tagged keys: UserID tracks who created the key (informational)
+	// For user-owned keys: UserID tracks the node owner
+	// Can be nil for system-created tagged keys
+	UserID *uint
```
kradalby, line 29: We should use uint64 for all ID types. 
ideally we should use the concrete UserID type, but not sure if that will trip up GORM.

hscontrol/types/preauth_key.go
```diff
@@ -23,16 +23,19 @@ type PreAuthKey struct {
...
 	Ephemeral bool `gorm:"default:false"`
 	Used      bool `gorm:"default:false"`
 
-	// Tags are always applied to the node and is one of
-	// the sources of tags a node might have. They are copied
-	// from the PreAuthKey when the node logs in the first time,
-	// and ignored after.
+	// Tags assigned to nodes created with this key.
+	// ONLY applied during initial authentication with the key.
+	// If non-empty, this creates tagged nodes.
```
kradalby: I think this is inaccurate, it is only set during authentication or via SetTags in the API?

Issue #1545, opened 2023-09-07
Support for WireGuard only peers


Nemo157:
## Why

Tailscale just [announced](https://tailscale.com/blog/mullvad-integration/) their support for integrated Mullvad exit nodes. Being able to configure a similar setup via Headscale and an independent Mullvad account (or other wireguard VPN provider) would be useful for those of us without a Tailscale account.

## Description

I haven't looked deeply into the details, but it's my understanding that this is [implemented via a "WireGuard only peer"](https://github.com/tailscale/tailscale/pull/7821) feature, and then support in the Tailscale coordination server to synchronize these peers with Mullvad. I assume it would be possible for Headscale to allow manually configuring these peer types.

sachiniyer:
It also seems like mullvad [publishes a script](https://github.com/mullvad/mullvad-wg.sh/blob/main/mullvad-wg.sh) to connect to mullvad servers. The most interesting thing is that you basically link public keys with your account (which is why I think that there is a preconfiguration step to register devices in their announcement).

It seems like the simplest implementation could be to create an external script that calls [`registerNodeCmd`](https://github.com/juanfont/headscale/blob/7edc953d35f8c783b77a9409ab4208f50d5187f8/cmd/headscale/cli/nodes.go#L108) on mullvad endpoints, and then calling the [mullvad api](https://github.com/mullvad/mullvad-wg.sh/blob/ce91d41ae72b5166d87cb5de6956d22bc613af81/mullvad-wg.sh#L59) with each of the node's public keys you want to link.

(I also am really not an expert in this, so please take it with a grain of salt)

averywinters:
It seems useful to me to allow any generic WireGuard-only peer as an exit node, not just Mullvad servers. That way the open-source control server doesn't have to be tied to one VPN provider like the closed-source version.

averywinters:
It might be possible to support this for any WireGuard server peer by accepting peer config files like those generated by [Mullvad's WireGuard config file generator][0] like described in [this guide][1], or by just asking the user to provide the generated fields we need at the CLI when adding the peer.

[0]: https://mullvad.net/en/account/wireguard-config?platform=linux
[1]: https://mullvad.net/en/help/easy-wireguard-mullvad-setup-linux/

Nemo157:
[...]

PR #2892, opened 2025-11-14
Wireguard only peers

iridated:
Implements #1545 as discussed in that feature

**Checklist**
- [x] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md


kradalby:
Hello, thanks, just to manage expectations, I'm gonna have to look at this every ones in a while, because it is very overwhelming, and I will still prioritise the current roadmap and the work I have to do. 

At a first glance, it does look promising, but I am very concerned about the maintenance burden over time. That said I am positive to ultimately getting it in.

Cerothen:
I would greatly appreciate WG only peers since it would make it more straightforward for including third party VPN providers (like tailscale now has mullivad). Which would be great if I want to have one of my devices show an exit node somewhere more easily.

kradalby:
Just checking back in as this PR is in my mind, I am just not sure what to do or how we should tackle it yet.

In addition to maintenance, my genuine main concern with this PR is that I suspect _most users_ do not understand what "WireGuard peers" will mean, even if both [I and @iridated](https://github.com/juanfont/headscale/issues/1545#issuecomment-3441636847) have written up what it actually allows.
I am concerned about the support burden of people misunderstanding and posting issues or taking up space on Slack.

That said, I am still fairly committed to eventually get it in, but we need to split it up and do it bit by bit, and we need to find a way to ensure we can support it over time. 

I think it is unrealistic to get it in before 0.31.0, which means we aim to redo the API/gRPC before hand making a lot of the 6000 lines in this PR obsolete.

I think a meaningful first step would be to factor out all of the WireGuard server-side code into a new PR that is rebased on top of the current main, so we can start to review and test the data structures. I appreciate that not having the CLI/API makes it harder to test, but for now, lets try to just get the amount of code into a size that is reviewable, and make the automated tests run by keeping it rebased on top of main.

Speaking of keeping it on top of main. As I expect this effort to take some time, a reasonable measure for maintainability is going to be to keep it up to date on top of main, which will also measure community commitment. If we do not manage this, the effort will likely go stale and I suspect it will sit there until I eventually have time or want to prioritise it. Currently it is not going to make any roadmap for a long time.

iridated:
Thanks for circling back to have a look! The end of year was a bit hectic but I finally had time to write up my thoughts (and hopefully rebase on top of main soon)

>  I am concerned about the support burden of people misunderstanding and posting issues or taking up space on Slack.

That makes sense, it did seem like there was some confusion around what this feature actually makes possible (and what it cannot do). I'm not sure what a good solution here is. I'm happy to punt the CLI commands and documentation into an "Expert" section if you think that would be better, or otherwise try to have an FAQ with common (non-)usecases?

> I think it is unrealistic to get it in before 0.31.0, which means we aim to redo the API/gRPC before hand making a lot of the 6000 lines in this PR obsolete.

Can you elaborate on where I can find this code/what needs to be updated? It looks like the latest release was 0.28 and I had a quick look at `main` where I couldn't find any major changes to this system.

> I think a meaningful first step would be to factor out all of the WireGuard server-side code into a new PR that is rebased on top of the current main, so we can start to review and test the data structures. I appreciate that not having the CLI/API makes it harder to test, but for now, lets try to just get the amount of code into a size that is reviewable, and make the automated tests run by keeping it rebased on top of main.

I am a little worried getting rid of the CLI will make it much harder to catch bugs since we can't test it in a realistic scenario - is the idea here just to let the automated tests run and worry about compatibility with clients etc in the future? Happy to factor this out if you think it will make it easier to review but I'm worried splitting this out over multiple PRs will just make things more messy. I appreciate it's a lot to review at once, does github not allow for partial reviews so you could go through the files more slowly?

Lite5h4dow:
Hello. Is there anyway I can help with this?

smartin015:
I've been using this in an experimental setup since around the start of the year - it's generally rock solid but I did notice an issue just now with IP assignment of new nodes. It appears that new nodes are assigned an incrementing IP address without any checking of whether the wireguard-only node is already assigned to that address. This is apparently fine for nodes that aren't using a WG exit node, but can lead to failure to reach the node if an exit node is enabled.

PR #2922, opened 2025-11-28
cli: remove node move command

kradalby:
This commit removes support for moving nodes between users.

The reason for this is that allowing such move is going against a intended design about identity in Tailscale, where your peer has an identity based on their machine, and user id. By allowing the node to be moved, we are breaking this contract, a user might have a peer, and it cannot be trusting who owns this peer as it cant change.

This feature should not have been added, and now that Tags are being implemented, essentially representing service account, allowing better seggeration anyways, we will remove it.

One of the problems associated with this would be similar to the email takeover on an OIDC platform.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ release.
   - If you are running a version older than 0.25.0, you must upgrade to 0.25.1 first, then upgrade to this release
   - See the [upgrade path documentation](https://headscale.net/stable/about/faq/#what-is-the-recommended-update-path-can-i-skip-multiple-versions-while-updating) for detailed guidance
   - In version 0.29, all migrations before 0.28.0 will also be removed
+- Remove ability to move nodes between users [#2922](https://github.com/juanfont/headscale/pull/2922)
+  - The `headscale nodes move` CLI command has been removed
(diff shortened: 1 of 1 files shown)


liorsl:
Just my 2 cents here, but unless using OIDC the owner of the device doesn't have any control over the user that the device is assigned to in the first place. I understand Headscale is an OSS project but I just want to mention not having this command anymore will make my (and I am sure other users as well) experience just a lot harder. I have personally used the aforementioned move command quite a lot. 
Of course, I appreciate the work you guys do for the community, as a developer I understand the time spent here. I don't want to sound entitled but just to voice my own personal opinion

kradalby:
> One of the problems associated with this would be similar to the email takeover on an OIDC platform.

It might have been a mistake to include this part, the point of this statement isnt to imply that this is being removed because of OIDC and such, it is just to highlight a similar attack surface. 

We are removing it because of OIDC or the similar, but because it is a feature that does not align with the security model that Tailscale has, and particularly as a reimplementation, it is so easy to introduce security issues, and this is a step in the direction of minimising the possibility to do that.

Instead of moving a device, the correct path is to just reauthenticate it, which is relatively trivial.

PR #2932, opened 2025-12-04
feat: add DERPHomeParams settings

### LINKED ISSUE #2918: [Feature] Can derper be set with priority?
### Use case

First of all, I am deeply grateful to you for developing such a user-friendly tool for remote networking.
In some cases, some servers that are not very useful are used as transit servers due to their low latency. However, I actually have a server with higher bandwidth, and I prefer to use this server instead of the server with low latency. But I cannot remove the server with low latency because I need it as a backup server. So this is what I am struggling with.
If I can set priorities in derper.yaml, I believe this would be a great solution. Would this be considered?
Thank you!

### Description

I would like to request support for configurable server priority in derper.yaml.
﻿
Currently, DERP clients always choose the server with the lowest latency. In some cases, this leads clients to use a low-latency but low-bandwidth server that should only serve as a backup.
﻿
By allowing administrators to assign a priority value to each DERP server, clients could prefer higher-priority servers regardless of small latency differences, while keeping lower-priority servers available as fallbacks. This would give administrators more control over routing behavior and improve reliability and performance in multi-server environments.

### Contribution

_No response_

SamyDjemai:
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Fixes #2918 (hopefully)

### DIFF
--- a/hscontrol/derp/derp.go
+++ b/hscontrol/derp/derp.go
@@ -76,16 +76,24 @@ func loadDERPMapFromURL(addr url.URL) (*tailcfg.DERPMap, error) {
 func mergeDERPMaps(derpMaps []*tailcfg.DERPMap) *tailcfg.DERPMap {
 	result := tailcfg.DERPMap{
 		OmitDefaultRegions: false,
+		HomeParams:         &tailcfg.DERPHomeParams{},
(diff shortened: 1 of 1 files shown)


evlic:
This seems like a very simple and risk-controllable change, so why not merge?

Issue #2902, opened 2025-11-18
Implement tailcfg.PingRequest


kradalby:
Implement [`tailcfg.PingRequest`](https://github.com/tailscale/tailscale/blob/9048ea25db8064b3833b7a2fcbe4b421e4a820dc/tailcfg/tailcfg.go#L1844).

PingRequests allows the control server (Headscale) to send special requests to the client and get a response.

They can be range from simple "are you alive" requests to "asking the client to report back something" ([c2n requests](https://github.com/tailscale/tailscale/blob/main/tailcfg/c2ntypes.go)). 

This is needed to fix the following:
- https://github.com/juanfont/headscale/issues/1850 (c2n and internal endpoints)
- https://github.com/juanfont/headscale/issues/2129 (ping client to understand if it down)

PR #2936, opened 2025-12-06
fix: compilation error in auth_test.go - CreatePreAuthKey expects pointer

roganlynch:
Fixes a compilation error in `hscontrol/auth_test.go` where `CreatePreAuthKey` was being called with a value instead of a pointer.

## Changes
- Updated test at line 3237 to pass `*types.UserID` (pointer) instead of `types.UserID` (value)
- Creates a `userID` variable and passes its address to match the function signature

## Error Fixed
```
cannot use types.UserID(user.ID) (value of type types.UserID) as
*types.UserID value in argument to app.state.CreatePreAuthKey
```

## Context
This is part of the work discussed in #2902 to add ping functionality to headscale.

---


nblock:
Should be fixed in 15c84b34e031d828962e2ed41f7398869bdd4932

PR #2938, opened 2025-12-06
feat(mapper): add SendDirectUpdate method for direct MapResponse delivery

roganlynch:
Adds `SendDirectUpdate` method to the mapper batcher interface for direct MapResponse delivery.

## Changes
- Extends the `Batcher` interface with `SendDirectUpdate(ctx context.Context, node *types.Node, update *tailcfg.MapResponse) error`
- Implements the method to deliver MapResponse updates directly to specific nodes without queuing
- Provides foundation for sending targeted ping requests to nodes

## Context
This is a prerequisite for the ping functionality, allowing the control plane to send MapResponses containing ping requests directly to target nodes. This builds upon the mapper infrastructure and enables immediate delivery of control messages.

---

### DIFF
--- a/hscontrol/mapper/batcher_lockfree.go
+++ b/hscontrol/mapper/batcher_lockfree.go
@@ -449,6 +449,38 @@ func (b *LockFreeBatcher) MapResponseFromChange(id types.NodeID, c change.Change
 	}
 }
 
+// SendDirectUpdate sends a MapResponse directly to a specific node without going through
+// the change broadcast system. This is useful for targeted updates like PingRequests.
+func (b *LockFreeBatcher) SendDirectUpdate(id types.NodeID, resp *tailcfg.MapResponse) error {
+	if resp == nil {
+		return fmt.Errorf("cannot send nil MapResponse to node %d", id)
+	}
+
+	// Look up the node connection
+	nodeConn, exists := b.nodes.Load(id)
+	if !exists {
+		return fmt.Errorf("node %d not found in batcher", id)
+	}
+
+	// Check if node has active connections
+	if !nodeConn.hasActiveConnections() {
+		return fmt.Errorf("node %d has no active connections", id)
+	}
+
+	// Send the MapResponse directly
+	err := nodeConn.send(resp)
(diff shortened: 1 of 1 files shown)


hscontrol/mapper/batcher_lockfree.go
```diff
@@ -449,6 +449,38 @@ func (b *LockFreeBatcher) MapResponseFromChange(id types.NodeID, c change.Change
 	}
 }
 
+// SendDirectUpdate sends a MapResponse directly to a specific node without going through
+// the change broadcast system. This is useful for targeted updates like PingRequests.
+func (b *LockFreeBatcher) SendDirectUpdate(id types.NodeID, resp *tailcfg.MapResponse) error {
```
kradalby, line 454: I think it would be fine to add this method, but, I am unsure if it is needed. In principle, just adding the message to the current batcher pipeline, but saying it is only for a specific node should be sufficient as it should never spend much time there before being sent. 

This will eliminate the need for a lot of additional code. And the positive with not having the ability to send directly is that developers (or sloppy AI) cant get tempted to bypass the batching mechanism.

hscontrol/mapper/batcher_lockfree.go
```diff
@@ -449,6 +449,38 @@ func (b *LockFreeBatcher) MapResponseFromChange(id types.NodeID, c change.Change
...
+	if resp == nil {
+		return fmt.Errorf("cannot send nil MapResponse to node %d", id)
+	}
+
+	// Look up the node connection
```
kradalby, line 459: Please instruct the AI to not add what comments, and only add why comments, where it explain why something is happening, not these mundane things where the code should be sufficient documentation.

roganlynch: Sure.  Apologies for the eyesores...

hscontrol/mapper/batcher_lockfree.go
roganlynch: Sure.  Happy to streamline to drop duplicative code.

PR #2939, opened 2025-12-06
feat(ping): add core PingRequest/Response implementation

roganlynch:
Adds core ping functionality for node health checks and connectivity testing.

## Changes
- **PingManager**: Manages ping request lifecycle, tracking, and response handling
- **PingRequest types**: Support for disco, TSMP, and c2n ping types
- **Noise protocol integration**: Authenticated ping response handler
- **HTTP endpoint**: `/machine/ping-response/{request_id}` for receiving responses
- **Interface extension**: Uses SendDirectUpdate method for direct message delivery

## Implementation Details
The PingManager creates unique ping requests with callback URLs, tracks them with timeouts, and coordinates responses. The Noise protocol ensures only authenticated nodes can submit ping responses.

This builds on the SendDirectUpdate capability from the previous commit and provides the foundation for gRPC ping endpoints and the scheduler.

## Context

## Dependencies
- Requires PR #2938 (SendDirectUpdate method)

---

### DIFF
--- a/hscontrol/ping.go
+++ b/hscontrol/ping.go
@@ -1,0 +1,280 @@
+package hscontrol
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/netip"
+	"sync"
+	"time"
+
+	"github.com/juanfont/headscale/hscontrol/types"
+	"github.com/rs/zerolog/log"
+	"tailscale.com/tailcfg"
+)
+
+// PingRequest tracks a pending ping request to a client
+type PingRequest struct {
+	ID           string
+	NodeID       types.NodeID
+	URL          string
+	Types        string
+	IP           string
+	Payload      []byte
+	CreatedAt    time.Time
+	ResponseChan chan *tailcfg.PingResponse
+	ctx          context.Context
+	cancel       context.CancelFunc
+}
+
+// PingManager manages ping requests to clients
+type PingManager struct {
+	mu       sync.RWMutex
+	requests map[string]*PingRequest
+	baseURL  string
+}
+
+// NewPingManager creates a new ping manager
+func NewPingManager(baseURL string) *PingManager {
+	return &PingManager{
+		requests: make(map[string]*PingRequest),
+		baseURL:  baseURL,
+	}
+}
+
+// CreatePingRequest creates a new ping request with a unique URL
+func (pm *PingManager) CreatePingRequest(nodeID types.NodeID, pingType string, targetIP string, payload []byte) (*PingRequest, error) {
+	// Generate unique request ID
+	idBytes := make([]byte, 16)
+	if _, err := rand.Read(idBytes); err != nil {
+		return nil, fmt.Errorf("generating request ID: %w", err)
+	}
+	requestID := hex.EncodeToString(idBytes)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+
+	req := &PingRequest{
+		ID:           requestID,
+		NodeID:       nodeID,
+		URL:          fmt.Sprintf("%s/machine/ping-response/%s", pm.baseURL, requestID),
+		Types:        pingType,
+		IP:           targetIP,
+		Payload:      payload,
+		CreatedAt:    time.Now(),
+		ResponseChan: make(chan *tailcfg.PingResponse, 1),
+		ctx:          ctx,
+		cancel:       cancel,
+	}
+
+	pm.mu.Lock()
+	pm.requests[requestID] = req
+	pm.mu.Unlock()
+
+	// Clean up after timeout
+	go func() {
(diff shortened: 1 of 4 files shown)


hscontrol/mapper/batcher_lockfree.go
```diff
@@ -449,6 +449,38 @@ func (b *LockFreeBatcher) MapResponseFromChange(id types.NodeID, c change.Change
 	}
 }
 
+// SendDirectUpdate sends a MapResponse directly to a specific node without going through
```
kradalby, line 452: I think my original suggestion was to batch these PR a bit bigger than one for each, so we avoid this where we have the same commits in multiple PRs.

hscontrol/ping.go
```diff
@@ -0,0 +1,280 @@
+package hscontrol
```
kradalby, line 1: I will have to think more about this, I am not sure how we want to structure this yet.

hscontrol/ping.go
```diff
@@ -0,0 +1,280 @@
...
+	ResponseChan chan *tailcfg.PingResponse
+	ctx          context.Context
+	cancel       context.CancelFunc
+}
+
+// PingManager manages ping requests to clients
+type PingManager struct {
+	mu       sync.RWMutex
+	requests map[string]*PingRequest
+	baseURL  string
```
kradalby, line 38: I do not think we actually need a base URL, I think the client ignores the part: https://github.com/tailscale/tailscale/blob/8eda947530cebbe3dc7882ace4d9f2829b0448da/ipn/ipnlocal/web_client.go#L183-L185

Would be good to test, because it would simplify more.

hscontrol/ping.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+}
+
+// CreatePingRequest creates a new ping request with a unique URL
+func (pm *PingManager) CreatePingRequest(nodeID types.NodeID, pingType string, targetIP string, payload []byte) (*PingRequest, error) {
+	// Generate unique request ID
+	idBytes := make([]byte, 16)
+	if _, err := rand.Read(idBytes); err != nil {
+		return nil, fmt.Errorf("generating request ID: %w", err)
+	}
```
kradalby, line 55: we already have a few random ID generators implemented, use one of those.

hscontrol/ping.go
```diff
@@ -0,0 +1,280 @@
...
+	select {
+	case req.ResponseChan <- response:
+		return nil
+	case <-req.ctx.Done():
+		return fmt.Errorf("request timeout")
+	}
+}
+
+// CreateKeepAlivePing creates a simple keep-alive ping request
+func (pm *PingManager) CreateKeepAlivePing(nodeID types.NodeID) (*tailcfg.PingRequest, error) {
```
kradalby, line 116: What is this for?

hscontrol/ping.go
```diff
@@ -0,0 +1,280 @@
...
+		PingRequest: pingReq,
+		KeepAlive:   false, // This is a real update, not just a keepalive
+	}
+
+	return h.mapBatcher.SendDirectUpdate(nodeID, resp)
+}
+
+// CheckNodeOnline checks if a node is online by sending a health check ping.
+// It creates a ping request, sends it via the MapBatcher, and waits for a response.
+func (h *Headscale) CheckNodeOnline(nodeID types.NodeID, targetIP string) (*tailcfg.PingResponse, error) {
```
kradalby, line 251: I feel like this manager in general has too much "exposed api", so a user have access to too much internal channels and other stuff from the PingResponse structure. 

This function is more of a sensible wrapper, where the thing you can access is the "what do I want, I want to know if a node is healthy", without having to do a large multistep building block thing. Of course we need them available, but I suspect we want to have them internally in a package, and only expose specific functions, like "Is it still here" (health sounds misguided, because the node being available doesnt have to mean healthy).

I would think that the only exposed things should be pinging (are you there) and various c2n implementations we add over time. The rest are internals.

hscontrol/ping.go
roganlynch: I can test this and get back to you.

hscontrol/ping.go
roganlynch: Roger.

hscontrol/ping.go
roganlynch: Good Question. Probably vestigial/superfluous. Will investigate and remove as needs be.

hscontrol/ping.go
roganlynch: Roger. So we're starting small - I'll see what can be dropped/simplified.

hscontrol/mapper/batcher_lockfree.go
roganlynch: Apologies - NooB - Went the opposite extreme after the monolithic PR

kradalby:
Superseded by #3183 which was merged. Closing to keep the PR list clean. Thanks!

PR #2940, opened 2025-12-06
feat: add ping scheduler for node liveness checks

roganlynch:
Adds a ping scheduler for automated node liveness checks.

## Changes
- **PingScheduler**: Manages periodic ping operations for node health monitoring
- **Configurable intervals**: Supports customizable ping schedules per node
- **Automatic cleanup**: Handles timeout and response cleanup
- **Integration with PingManager**: Uses the core ping infrastructure for execution

## Implementation Details
The scheduler maintains a registry of nodes requiring periodic health checks and coordinates ping request timing. It integrates with the PingManager to execute pings and process responses, providing automated liveness monitoring for the network.

## Context

## Dependencies
- Requires PR #2939 (core ping implementation)

---

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -66,6 +66,35 @@ prefixes:
   # - random: assigns the next free IP from a pseudo-random IP generator (crypto/rand).
   allocation: sequential
 
+# Node liveness checking configuration.
+# Headscale can periodically ping nodes to verify they are still reachable
+# and detect network failures faster than the default TCP timeout (~16 minutes).
+# This helps with scenarios like subnet router failover and detecting
+# disconnected clients.
+node_liveness:
+  # Enable the periodic ping scheduler for node liveness checks.
+  # When disabled, headscale relies on TCP timeouts to detect offline nodes.
+  # Default: false
+  enable_scheduler: false
+
+  # How often to check all nodes for liveness.
+  # Nodes will be pinged at distributed intervals within this window.
+  # Default: 2m
+  ping_interval: 2m
+
+  # Maximum random jitter to add to ping scheduling.
+  # This prevents all pings from happening at the same time and
+  # distributes the load across the interval window.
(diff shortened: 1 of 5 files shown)


config-example.yaml
```diff
@@ -66,6 +66,35 @@ prefixes:
   # - random: assigns the next free IP from a pseudo-random IP generator (crypto/rand).
   allocation: sequential
 
+# Node liveness checking configuration.
+# Headscale can periodically ping nodes to verify they are still reachable
+# and detect network failures faster than the default TCP timeout (~16 minutes).
+# This helps with scenarios like subnet router failover and detecting
+# disconnected clients.
+node_liveness:
```
kradalby, line 74: I think we will not add this as configuration. Checking if nodes are available will be a server internal thing, and not exposed to users. It will only be used for subnet routers that are in a failover configuration to failover faster. We should not use it for any other node as it will cause a large resource usage on the clients. 

So, any checking of availability should be set up automatically by headscale when nodes are high availability and detecting if they go away is important.

roganlynch: I had misunderstood  - I thought the purpose here was to be able to actively health check nodes on a policy basis - no simply for HA (Quasi-Keepalived)  purposes.

You're then planning to internally tag nodes as exit/router nodes,  determine a precedence,  set a flag, do periodic pings, and update the precedence accordingly?

So probably expose only a 'Ping Routers' flag in the config?

LMK if that's sort of what you're thinking.

kradalby: > So probably expose only a 'Ping Routers' flag in the config?

I think in general we will just always enable it for routers and not leave it user configurable. Subnet routers (at least ha ones) would be expected to not be on battery so the extra compute should not be a problem.

Lets close this one for now too, and then we focus on the internal implementation. I have kept the PingReq on the next release, so I will have more time to give it attention.

PR #2949, opened 2025-12-08
feat: Disable key expiry for nodes

### LINKED ISSUE #2681: Feature Request: Config-level switch to disable node-key expiration (for long-lived IoT fleets)
### Use case

We operate hundreds of battery-powered IoT devices that may remain offline for months or even years (e.g. in storage, remote sites without WAN, manual shutdown).

Each unit is provisioned in-factory with a reusable PAK and then shipped; once in the field we have no physical or remote means to re-run authentication.

When the node finally comes back online we must be able to SSH immediately for service, recovery or firmware updates.

### Description

**Current situation**
Headscale tracks `node.expiry`. Once the expiry passes the node stops forwarding traffic.

The only workaround is to run `headscale nodes expire --reset …` (or set `expiry = NULL` in SQLite) per node, either after enrollment or post-expiry.

Automating that reset via cron/API is fragile: a missed run or a future schema change could strand thousands of devices.

**Requested behavior**

Tailscale has a similar feature: [Disabling key expiry](https://tailscale.com/kb/1028/key-expiry#disabling-key-expiry)

_Add a single configuration flag - global or tag-scoped - that disables key expiry entirely for matching nodes._

```
# config.yaml
...
node_key_expiration:
  disabled: true            # global switch
  # or more granular
  exempt_tags:
    - tag:battery           # never expire devices with this tag
```
If `disabled: true`, Headscale should skip scheduling expiry for every new node and ignore `expiry` checks during map generation.

If _exempt_tags_ is supplied, only nodes carrying one of those tags are exempt; others follow normal 180 d rotation.

**Impact**

Zero behavior change for existing users: the feature defaults to `disabled = false`.

Simplifies large-scale IoT deployments by removing a hidden operational pitfall.

This would keep security knobs (manual nodes expire …, ACL tags) intact for cases where an operator really wants to revoke a stale device.

Thank you for considering, that would really have a huge impact for us!

### Contribution

_No response_

marcotuna:
Fixes #2681

Adds the ability to disable key expiry for nodes via CLI, similar to Tailscale's implementation (https://tailscale.com/kb/1028/key-expiry#disabling-key-expiry).
This is a simpler implementation that focuses only on providing the CLI argument to expire nodes, and therefore does not fully cover the implementation proposed in #2681.

### Usage

# Disable key expiry (node will never expire)
headscale nodes expire --identifier 1 --disable

### Implementation

- Added `disable_expiry` field to `ExpireNodeRequest` proto
- Added `--disable` flag to `headscale nodes expire` CLI command
- When disabled, node expiry is set to NULL and `IsExpired()` returns false

### Checklist

- [x] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/cmd/headscale/cli/nodes.go
+++ b/cmd/headscale/cli/nodes.go
@@ -55,6 +55,7 @@ func init() {
 
 	expireNodeCmd.Flags().Uint64P("identifier", "i", 0, "Node identifier (ID)")
 	expireNodeCmd.Flags().StringP("expiry", "e", "", "Set expire to (RFC3339 format, e.g. 2025-08-27T10:00:00Z), or leave empty to expire immediately.")
+	expireNodeCmd.Flags().BoolP("disable", "d", false, "Disable key expiry (node will never expire)")
 
 	err = expireNodeCmd.MarkFlagRequired("identifier")
 	if err != nil {
@@ -260,9 +261,11 @@ var listNodeRoutesCmd = &cobra.Command{
 }
 
 var expireNodeCmd = &cobra.Command{
-	Use:     "expire",
-	Short:   "Expire (log out) a node in your network",
-	Long:    "Expiring a node will keep the node in the database and force it to reauthenticate.",
+	Use:   "expire",
+	Short: "Expire (log out) a node in your network",
+	Long: `Expiring a node will keep the node in the database and force it to reauthenticate.
+
+Use --disable to disable key expiry (node will never expire).`,
 	Aliases: []string{"logout", "exp", "e"},
 	Run: func(cmd *cobra.Command, args []string) {
 		output, _ := cmd.Flags().GetString("output")
@@ -274,6 +277,49 @@ var expireNodeCmd = &cobra.Command{
 				fmt.Sprintf("Error converting ID to integer: %s", err),
 				output,
 			)
+
+			return
+		}
+
+		disableExpiry, err := cmd.Flags().GetBool("disable")
+		if err != nil {
+			ErrorOutput(
+				err,
+				fmt.Sprintf("Error getting disable flag: %s", err),
+				output,
+			)
+
+			return
+		}
+
+		ctx, client, conn, cancel := newHeadscaleCLIWithConfig()
+		defer cancel()
+		defer conn.Close()
+
+		// Handle disable expiry - node will never expire.
+		if disableExpiry {
+			request := &v1.ExpireNodeRequest{
+				NodeId:        identifier,
+				DisableExpiry: true,
+			}
+
+			response, err := client.ExpireNode(ctx, request)
+			if err != nil {
+				ErrorOutput(
+					err,
+					fmt.Sprintf(
+						"Cannot disable node expiry: %s\n",
+						status.Convert(err).Message(),
+					),
+					output,
+				)
+
+				return
+			}
+
+			SuccessOutput(response.GetNode(), "Node expiry disabled", output)
+
+			return
 		}
 
 		expiry, err := cmd.Flags().GetString("expiry")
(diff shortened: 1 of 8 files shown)


IamTaoChen:
maybe `--never` is better?

cmd/headscale/cli/nodes.go
```diff
@@ -257,7 +258,7 @@ var listNodeRoutesCmd = &cobra.Command{
 var expireNodeCmd = &cobra.Command{
 	Use:     "expire",
 	Short:   "Expire (log out) a node in your network",
-	Long:    "Expiring a node will keep the node in the database and force it to reauthenticate.",
+	Long:    "Expiring a node will keep the node in the database and force it to reauthenticate.\nUse --disable to disable key expiry (node will never expire).",
```
kradalby: can you use a "```" to use a multi line string instead?

hscontrol/db/node.go
```diff
@@ -311,6 +311,17 @@ func NodeSetExpiry(tx *gorm.DB,
...
+func (hsdb *HSDatabase) NodeDisableExpiry(nodeID types.NodeID) error {
+	return hsdb.Write(func(tx *gorm.DB) error {
+		return NodeDisableExpiry(tx, nodeID)
+	})
+}
+
+// NodeDisableExpiry clears the expiry for a node, making it never expire.
+func NodeDisableExpiry(tx *gorm.DB, nodeID types.NodeID) error {
+	return tx.Model(&types.Node{}).Where("id = ?", nodeID).Update("expiry", nil).Error
+}
```
kradalby: Do we need a specific method for this? could we just reuse set expiry?

hscontrol/state/state.go
```diff
@@ -654,6 +654,36 @@ func (s *State) SetNodeExpiry(nodeID types.NodeID, expiry time.Time) (types.Node
 	return s.persistNodeToDB(n)
 }
 
+// DisableNodeExpiry clears the expiration time for a node, making it never expire.
+func (s *State) DisableNodeExpiry(nodeID types.NodeID) (types.NodeView, change.ChangeSet, error) {
+	// Update NodeStore to clear expiry
+	n, ok := s.nodeStore.UpdateNode(nodeID, func(node *types.Node) {
+		node.Expiry = nil
+	})
```
kradalby: Same here, can we reuse SetExpiry?

cmd/headscale/cli/nodes.go
marcotuna: Updated to use ` instead following previous implementations. e.g https://github.com/juanfont/headscale/blob/main/cmd/headscale/cli/root.go#L136-L139

hscontrol/db/node.go
marcotuna: Indeed, I ended up removing this functions and ended up reusing the NodeSetExpiry

hscontrol/state/state.go
marcotuna: Removed in favor of SetNodeExpiry

hscontrol/state/state.go
```diff
@@ -637,22 +637,40 @@ func (s *State) ListEphemeralNodes() views.Slice[types.NodeView] {
...
 // SetNodeExpiry updates the expiration time for a node.
-func (s *State) SetNodeExpiry(nodeID types.NodeID, expiry time.Time) (types.NodeView, change.ChangeSet, error) {
+// If expiry is nil, the node's expiry is disabled (node will never expire).
+func (s *State) SetNodeExpiry(nodeID types.NodeID, expiry *time.Time) (types.NodeView, change.ChangeSet, error) {
 	// Update NodeStore before database to ensure consistency. The NodeStore update is
 	// blocking and will be the source of truth for the batcher. The database update must
 	// make the exact same change. If the database update fails, the NodeStore change will
 	// remain, but since we return an error, no change notification will be sent to the
 	// batcher, preventing inconsistent state propagation.
 	expiryPtr := expiry
```
kradalby: this isnt necessary since it is a ptr now.

kradalby (commented):
Great, can you rebase and fix the conflicts?
It would be great if you get your changes into a single commit that is rebased (no merge commit)

kradalby:
Sorry, the force push was tripping me up and broke the flow, will open a new PR with you as co-author.

marcotuna:
Hey @kradalby 
No worries and thanks for proceeding with this!

PR #2952, opened 2025-12-09
all: remove deadcode

kradalby:
more cleanup


nblock:
I'm not sure if this is related but there are many CI warnings from the integration tests: `No files were found with the provided path: control_logs/*/*.tar. No artifacts will be uploaded.`

PR #2954, opened 2025-12-11
.github/workflows: prebuilt integration test artifacts

kradalby:
This PR restructures the integration tests and prebuilds all common assets used in all tests:

- Headscale and Tailscale HEAD image
- `hi` binary that is used to run tests
- go cache is warmed up for compilation of the test

This essentially means we spend 6-10 minutes building assets before any tests starts, when that is done, all tests can just sprint through.

It looks like we are saving 3-9 minutes per test, and since we are limited to running max 20 concurrent tests across the repo, that means we had a lot of double work. 

There is currently 113 checks, so we have to do five runs of 20, and the saving should be quite noticeable! I think the "worst case" saving would be 20+min and "best case" probably towards an hour!

Thanks to @tomhjp for a lot of good tips!

### DIFF
--- a/integration/tsic/tsic.go
+++ b/integration/tsic/tsic.go
@@ -304,6 +304,15 @@ func New(
 		// Check if a pre-built image is available via environment variable
 		prebuiltImage := os.Getenv("HEADSCALE_INTEGRATION_TAILSCALE_IMAGE")
 
+		// If custom build tags are required (e.g., for websocket DERP), we cannot use
+		// the pre-built image as it won't have the necessary code compiled in.
+		hasBuildTags := len(tsic.buildConfig.tags) > 0
+		if hasBuildTags && prebuiltImage != "" {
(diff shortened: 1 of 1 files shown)


Dockerfile.integration
```diff
@@ -2,26 +2,42 @@
...
+# Copy source and build - invalidated on any source change
 COPY . .
 
 # Build debug binary with debug symbols for delve
 RUN CGO_ENABLED=0 GOOS=linux go build -gcflags="all=-N -l" -o /go/bin/headscale ./cmd/headscale
 
+# Runtime stage
+FROM debian:trixie-slim
+
+RUN apt-get update && apt-get install --no-install-recommends --yes \
```
nblock: Trixie supports update + install in one step:

`RUN apt-get --update install --no-install-recommends --yes …`

juanfont (approved):

PR #2956, opened 2025-12-12
capver: generate for 1.92

kradalby:


hscontrol/capver/capver_generated.go
```diff
@@ -5,55 +5,78 @@ package capver
...
 }
 
 // SupportedMajorMinorVersions is the number of major.minor Tailscale versions supported.
-const SupportedMajorMinorVersions = 9
+const SupportedMajorMinorVersions = 10
 
 // MinSupportedCapabilityVersion represents the minimum capability version
 // supported by this Headscale instance (latest 10 minor versions)
-const MinSupportedCapabilityVersion tailcfg.CapabilityVersion = 106
+const MinSupportedCapabilityVersion tailcfg.CapabilityVersion = 104
```
nblock: Could you set the `Minimum supported Tailscale client version` in the changelog to: v1.72.0 in case we really support it?

kradalby: no I'm waiting to bump it back up again, the docker release should go out today.

Issue #2389, opened 2025-01-28
[Bug] Tags applied to nodes are not applied in ACL until headscale service is restarted after adding the tag


SlackingVeteran:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When a tag is applied or removed from node, ACL is not applied for the node based on tags until headscale service is restarted

### Expected Behavior

When tag is applied or removed from nodes the changes should be reflected on user's account that has access to tag through ACL.

If user has access to the tag and the tag is removed from node then user should no longer have access to the node without restarting headscale service
And if user has access to the tag and the tag is **added** to the node then user should gain access to the node without restarting headscale service

### Steps To Reproduce

I have a `tag:shared` tag in ACL as follows:
```
{
  "groups": {
    "group:all-members": [
      "user1",
      "user2"
    ]
  },
  "tagOwners": {
    "tag:shared": [
      "group:all-members"
    ]
  },
  "hosts": {},
  "acls": [
    {
      "action": "accept",
      "src": [
        "user1"
      ],
      "dst": [
        "user1:*"
      ]
    },
    {
      "action": "accept",
      "src": [
        "user2"
      ],
      "dst": [
        "user2:*"
      ]
    },
    {
      "action": "accept",
      "src": [
        "group:all-members"
      ],
      "dst": [
        "tag:shared:*"
      ]
    }
  ],
  "ssh": []
}
```

Then I applied `tag:shared` to one of the node owned by `user1` with `headscale nodes tags with `headscale nodes tag --tags tag:shared -i 1` and shows up in `headscale nodes list -o json` as:
```
[
        {
                "id": 1,
                "machine_key": "mkey:<redacted>",
                "node_key": "nodekey:<redacted>",
                "disco_key": "discokey:<redacted>",
                "ip_addresses": [
                        "100.64.0.1",
                        "fd7a:115c:a1e0::1"
                ],
                "name": "USER1-DT",
                "user": {
                        "id": 1,
                        "name": "user1",
                        "created_at": {
                                "seconds": 1738102564,
                                "nanos": 803193762
                        },
                        "display_name": "FirstName LastName",
                        "email": "user1@example.com",
                        "provider_id": "<redacted>",
                        "provider": "oidc"
                },
                "last_seen": {
                        "seconds": 1738104866,
                        "nanos": 780535645
                },
                "expiry": {
                        "seconds": 1753654564,
[...]

PR #2961, opened 2025-12-13
tags: ensure tag changes propagate

kradalby:
This commit rebuilds the system for how we propagate
changes in headscale to make sure we send the appropriate
map response based on the changes that has occured instead
of sending a full map response when we do not know what to
do or to be lazy.

This closes gaps where we did not updates peers correctly
when a node was assigned or taken away from a node, or it
was changed in the Policy, and required the nodes to be
informed about changes to the network map.

Additional information in commits

Fixes https://github.com/juanfont/headscale/issues/2389

claude was used in this PR.

### DIFF
--- a/integration/acl_test.go
+++ b/integration/acl_test.go
@@ -77,11 +77,8 @@ func aclScenario(
 			// tailscaled to stop configuring the wgengine, causing it
 			// to not configure DNS.
 			tsic.WithNetfilter("off"),
-			tsic.WithDockerEntrypoint([]string{
-				"/bin/sh",
-				"-c",
-				"/bin/sleep 3 ; apk add python3 curl ; update-ca-certificates ; python3 -m http.server --bind :: 80 & tailscaled --tun=tsdev",
-			}),
+			tsic.WithPackages("curl"),
+			tsic.WithWebserver(80),
 			tsic.WithDockerWorkdir("/"),
 		},
 		hsic.WithACLPolicy(policy),
@@ -311,6 +308,7 @@ func TestACLHostsInNetMapTable(t *testing.T) {
 				[]tsic.Option{},
 				hsic.WithACLPolicy(&testCase.policy),
 			)
+
 			require.NoError(t, err)
 			defer scenario.ShutdownAssertNoPanics(t)
 
@@ -759,6 +757,7 @@ func TestACLNamedHostsCanReach(t *testing.T) {
 
 			test1fqdn, err := test1.FQDN()
 			require.NoError(t, err)
+
 			test1ip4URL := fmt.Sprintf("http://%s/etc/hostname", test1ip4.String())
 			test1ip6URL := fmt.Sprintf("http://[%s]/etc/hostname", test1ip6.String())
 			test1fqdnURL := fmt.Sprintf("http://%s/etc/hostname", test1fqdn)
@@ -770,6 +769,7 @@ func TestACLNamedHostsCanReach(t *testing.T) {
 
 			test2fqdn, err := test2.FQDN()
 			require.NoError(t, err)
+
 			test2ip4URL := fmt.Sprintf("http://%s/etc/hostname", test2ip4.String())
 			test2ip6URL := fmt.Sprintf("http://[%s]/etc/hostname", test2ip6.String())
 			test2fqdnURL := fmt.Sprintf("http://%s/etc/hostname", test2fqdn)
@@ -781,6 +781,7 @@ func TestACLNamedHostsCanReach(t *testing.T) {
 
 			test3fqdn, err := test3.FQDN()
 			require.NoError(t, err)
+
 			test3ip4URL := fmt.Sprintf("http://%s/etc/hostname", test3ip4.String())
 			test3ip6URL := fmt.Sprintf("http://[%s]/etc/hostname", test3ip6.String())
 			test3fqdnURL := fmt.Sprintf("http://%s/etc/hostname", test3fqdn)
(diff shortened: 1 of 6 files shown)


hscontrol/mapper/mapper.go
```diff
@@ -143,6 +143,30 @@ func (m *mapper) fullMapResponse(
...
+
+	for _, node := range allNodes.All() {
+		// Skip self and current peers
+		if node.ID() == nodeID || peerIDs[node.ID()] {
+			continue
+		}
+
+		removedPeerIDs = append(removedPeerIDs, node.ID())
+	}
+
```
kradalby: We do not need to remove nodes here, we are sending a full response and WithPeers will send all the current nodes, the newest state, and the client will replace all of them.

hscontrol/policy/v2/policy.go
```diff
@@ -541,6 +538,37 @@ func (pm *PolicyManager) SetNodes(nodes views.Slice[types.NodeView]) (bool, erro
...
+// nodesPolicyChanged checks if any policy-affecting node properties have changed.
+// This includes node additions, removals, and changes to tags, user ownership, or IPs.
+// Returns true early on first detected change for efficiency.
+func (pm *PolicyManager) nodesPolicyChanged(newNodes views.Slice[types.NodeView]) bool {
+	// Different count means nodes were added or removed
+	if pm.nodes.Len() != newNodes.Len() {
+		return true
+	}
+
+	// Build map of old nodes for O(1) lookup
```
kradalby: stop doing comments explaining code that is self explanatory, only explain code where a  decision was made and a WHY is needed .

hscontrol/policy/v2/policy.go
```diff
@@ -541,6 +538,37 @@ func (pm *PolicyManager) SetNodes(nodes views.Slice[types.NodeView]) (bool, erro
 	return false, nil
 }
 
+// nodesPolicyChanged checks if any policy-affecting node properties have changed.
+// This includes node additions, removals, and changes to tags, user ownership, or IPs.
+// Returns true early on first detected change for efficiency.
+func (pm *PolicyManager) nodesPolicyChanged(newNodes views.Slice[types.NodeView]) bool {
```
kradalby: I dont like the name, it kind of implies the policy has already changed, and it hasnt, this function implies that the policy will or has to change or be recalculated.

hscontrol/types/node.go
```diff
@@ -973,6 +973,25 @@ func (nv NodeView) HasNetworkChanges(other NodeView) bool {
...
 
+// HasPolicyChange returns true if the node has changes that affect policy evaluation.
+// This includes user ownership, tags, and IPs - properties used in ACL rule matching.
+// Used to detect when policy filters need recompilation after node updates.
+func (nv NodeView) HasPolicyChange(other NodeView) bool {
+	if nv.UserID() != other.UserID() {
+		return true
+	}
+
+	if !slices.Equal(nv.Tags().AsSlice(), other.Tags().AsSlice()) {
```
kradalby: Does views of slices not have an equal? do we need to copy/cast it to AsSlice?

integration/acl_test.go
```diff
@@ -2255,3 +2255,520 @@ func TestACLPolicyPropagationOverTime(t *testing.T) {
 
 	t.Log("All 5 iterations completed successfully - ACL propagation is working correctly")
 }
+
+// TestACLTagPropagation validates that tag changes propagate immediately
+// to ACLs without requiring a Headscale restart.
+// This is the primary test for GitHub issue #2389.
```
kradalby, line 2295: This test should also validate that nodes get updated with nodes new identity in the case of a node being assigned tags with SetTags.

Nodes need to know:
- If their peer now is owned by a tag, not a user
- If themselves need a new packet filter based on their new identity (tags changed)
- If peers need a new packet filter based on an identity change (tags changed)

kradalby: We need to ensure we are validating all of these parts. 
in addition, the packetfilter is used to trim away nodes with no access, so we need to validate where the change will fully remove the node from the peer list, and when it will still be there, but remove a single port, like port 80. In that case, ping should still work, but not curl.

hscontrol/mapper/mapper.go
```diff
@@ -143,7 +144,7 @@ func (m *mapper) fullMapResponse(
 ) (*tailcfg.MapResponse, error) {
 	peers := m.state.ListPeers(nodeID)
 
-	return m.NewMapResponseBuilder(nodeID).
+	builder := m.NewMapResponseBuilder(nodeID).
```
kradalby: This seems unnecessary, we dont touch the variable in a new way.

hscontrol/mapper/builder.go
```diff
@@ -33,6 +33,7 @@ const (
 	removeResponseDebug debugType = "remove"
 	changeResponseDebug debugType = "change"
 	derpResponseDebug   debugType = "derp"
```
kradalby: Are we still using these types (and the functions for them) now that we use the new response to builder?

hscontrol/types/change/change.go
```diff
@@ -1,241 +1,92 @@
...
-	NodeEndpoint    Change = 26
-	NodeDERP        Change = 27
-
-	// User changes.
-	UserNewOrUpdate Change = 51
-	UserRemove      Change = 52
-)
+// NodeAddedResponse returns a Response for when a node is added or updated.
+// The OriginNode field enables self-update detection by the mapper.
+func NodeAddedResponse(id types.NodeID) Response {
```
kradalby: cannot everything in this and response.go go in the same file?

hscontrol/types/change/response.go
```diff
@@ -0,0 +1,328 @@
...
+	"slices"
+	"time"
+
+	"github.com/juanfont/headscale/hscontrol/types"
+	"tailscale.com/tailcfg"
+)
+
+// Response declares what should be included in a MapResponse.
+// The mapper uses this to build the response without guessing.
+type Response struct {
```
kradalby: I am not sure about the Response name, it kind of feels weird to call it that so far up in the chain. Maybe we can just call it Change? It should be free now that we have remove the Change type.

nblock (approved):

almereyda:
After returning here from the v0.28.0 release notes, I'm finding

> when a node was assigned or taken away from a node

To later visitors who didn't observe this grow: the first "node" meant "tag".

This PR helps with situations

> when a tag was assigned or taken away from a node

PR #2976, opened 2025-12-19
hostname: normalize before validation

### LINKED ISSUE #2926: [Bug] Normalization of hostnames

### Current Behavior

First off, love Headscale, it works great!

I think I found a bug in hostname handling.

When a client reports a hostname like "My-PC!", it fails ValidateHostname() and gets replaced with "invalid-xyz". Headscales code has a NormaliseHostname() function that would sanitize this to "my-pc".

The affected code paths are:
- ApplyHostnameFromHostInfo() *hscontrol/types/node.go*
- EnsureHostname() in *hscontrol/util/util.go*

Both use ValidateHostname() (reject if invalid), instead of NormaliseHostname() (sanitize and accept).

This seems like an oversight? The NormaliseHostname function exists but isn't being used.

### Expected Behavior

Expected: Hostname "My-PC!" → sanitized to "my-pc"
Actual: Hostname "My-PC!" → rejected, becomes "invalid-abc123"

### Steps To Reproduce

1. Register a node with hostname containing invalid characters:
### On client with hostname "My-PC!"
tailscale up --login-server https://your-headscale-server

2. List nodes on server:
headscale nodes list

3. Expected: Node appears as my-pc

4. Actual: Node appears as invalid-abc123

### Environment

```markdown
- Headscale version: v0.27.1
- Deployment: Docker
- OS: Ubuntu 24.04.3 LTS
- Docker: 29.0.2
```

### Runtime environment

### Debug information

This is a code logic issue, so dont believe debug log is useful.

griffer:
Apply NormaliseHostname() before ValidateHostname() in ApplyHostnameFromHostInfo() and EnsureHostname() to sanitize hostnames with minor invalid characters instead of replacing them with generic identifiers like "invalid-abc123".

This allows hostnames like "My-PC!" to become "my-pc" rather than being rejected.

Fixes #2926

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/types/node.go
+++ b/hscontrol/types/node.go
@@ -585,15 +585,23 @@ func (node *Node) ApplyHostnameFromHostInfo(hostInfo *tailcfg.Hostinfo) {
 		return
 	}
 
-	newHostname := strings.ToLower(hostInfo.Hostname)
-	if err := util.ValidateHostname(newHostname); err != nil {
+	newHostname, err := util.NormaliseHostname(hostInfo.Hostname)
+	if err != nil {
 		log.Warn().
 			Str("node.id", node.ID.String()).
 			Str("current_hostname", node.Hostname).
-			Str("rejected_hostname", hostInfo.Hostname).
+			Str("original_hostname", hostInfo.Hostname).
 			Err(err).
-			Msg("Rejecting invalid hostname update from hostinfo")
+			Msg("Hostname normalization failed, keeping current hostname")
 		return
+	}
+
+	if hostInfo.Hostname != newHostname {
+		log.Info().
+			Str("node.id", node.ID.String()).
+			Str("original", hostInfo.Hostname).
+			Str("sanitized", newHostname).
+			Msg("Hostname sanitized during update")
 	}
 
 	if node.Hostname != newHostname {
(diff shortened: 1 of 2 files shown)


hscontrol/types/node.go
```diff
@@ -585,17 +585,25 @@ func (node *Node) ApplyHostnameFromHostInfo(hostInfo *tailcfg.Hostinfo) {
 		return
 	}
 
-	newHostname := strings.ToLower(hostInfo.Hostname)
-	if err := util.ValidateHostname(newHostname); err != nil {
+	newHostname, err := util.NormaliseHostname(hostInfo.Hostname)
```
kradalby, line 588: I am not oppose to relaxing this again, but I will be a lot more conservative. 
The reason we made it very strict is to untangle the problems we had with invalid names sneaking in, and then also getting in all over the system without one place to control.

When we open up, we need to do : 
- Update/make a Normalise function that aligns with Tailscale SaaS behaviour
- Make sure that the "chain" of normalising, validating and checking for collision is done in the correct order, aligning with upstream behaviour. 

This means that we need to do the exercise of testing these variants against the SaaS control plane to get this right. We often get into the trap on patching things over and over and still getting different behaviour, which ultimately causes support burden.

griffer: @kradalby Thanks for the explanation, it's very much appreciated. As i wrote in my initial bug report, i was not sure if this was a bug or by design. I thought there probably were a more complex reason why things were done like they are.

aalmenar: > * Update/make a Normalise function that aligns with Tailscale SaaS behaviour

Tailscale does not publish its server side normalization rules, so what would be the preferred way to achieve this?

PR #2992, opened 2026-01-07
preauthkey: allow tags-only keys without user

### LINKED ISSUE #2977: [Bug] Preauthkey always requires a user

### Current Behavior

Creating a preauthkey with tags fails as it currently always requires a user:

```bash
$ headscale preauthkeys create --tags tag:foo,tag:bar
Error: required flag(s) "user" not set
```

In SaaS, one can create an auth key with tags defined. It seems to be sufficient to have those tags defined in the policy in the `tagOwners` section (note that *no owner* is defined on those tags):

```json5
{
	"tagOwners": {
		"tag:bar": [],
		"tag:foo": [],
	},
	"acls": [
		{
			"action": "accept",
			"src":    ["*"],
			"dst":    ["*:*"],
		}
}
```

In Headscale, if the preauthkey is assigned to a user:

```bash
 headscale preauthkeys list -u 1
ID | Key/Prefix                  | Reusable | Ephemeral | Used  | Expiration          | Created             | Tags   
1  | hskey-auth-CsLiZAFSkM2z-*** | false    | false     | false | 2025-12-20 17:01:54 | 2025-12-20 16:01:54 | tag:bar
   |                             |          |           |       |                     |                     | tag:foo
```

… it can't be used for non-interactive authentication:

```bash
$ tailscale up --login-server https://headscale.example.com --auth-key hskey-auth-CsLiZA... --advertise-tags=tag:bar,tag:foo
backend error: handling register with auth key: creating new node: requested tags [tag:bar tag:foo] are invalid or not permitted
```

It works with interactive auth (remove the `--auth-key` argument) or when no tags are advertised by the node (remove the `--advertise-tags`) argument.

### Expected Behavior

Be able to create preauthkeys that are either assigned to a user or have tags. All other `headscale preauthkeys` subcommands such as list also require to pass a user.

### Steps To Reproduce

Try to create a preauthkey with tag but without a user

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.28.0-beta.1
- Tailscale version: 1.92.3
```

### Runtime environment

### Debug information

none

kradalby:
Allow creating PreAuthKeys with tags but without a user. Tagged nodes don't
require user ownership since tags define their identity.

Changes:
- Switch preauthkey API from user-based to ID-based operations
- Allow CreatePreAuthKey without user when tags are provided
- Handle tags-only keys correctly during node registration
- Fix CLI argument validation for expire/delete commands

Fixes #2977

### DIFF
--- a/cmd/headscale/cli/preauthkeys.go
+++ b/cmd/headscale/cli/preauthkeys.go
@@ -78,7 +78,7 @@ var listPreAuthKeys = &cobra.Command{
 				"Used",
 				"Expiration",
 				"Created",
-				"Tags",
+				"Owner",
 			},
 		}
 		for _, key := range response.GetPreAuthKeys() {
@@ -87,13 +87,14 @@ var listPreAuthKeys = &cobra.Command{
 				expiration = ColourTime(key.GetExpiration().AsTime())
 			}
 
-			aclTags := ""
-
-			for _, tag := range key.GetAclTags() {
-				aclTags += "\n" + tag
+			var owner string
+			if len(key.GetAclTags()) > 0 {
+				owner = strings.Join(key.GetAclTags(), "\n")
+			} else if key.GetUser() != nil {
+				owner = key.GetUser().GetName()
+			} else {
+				owner = "-"
 			}
-
-			aclTags = strings.TrimLeft(aclTags, "\n")
 
 			tableData = append(tableData, []string{
 				strconv.FormatUint(key.GetId(), 10),
@@ -103,7 +104,7 @@ var listPreAuthKeys = &cobra.Command{
 				strconv.FormatBool(key.GetUsed()),
 				expiration,
 				key.GetCreatedAt().AsTime().Format("2006-01-02 15:04:05"),
-				aclTags,
+				owner,
 			})
 
 		}


hscontrol/state/state.go
```diff
@@ -1430,11 +1430,29 @@ func (s *State) HandleNodeFromPreAuthKey(
 		return types.NodeView{}, change.Change{}, err
 	}
 
+	// Helper to get username for logging (handles nil User for tags-only keys)
+	pakUsername := func() string {
+		if pak.User != nil {
+			return pak.User.Username()
+		}
+
+		return "<tags-only>"
```
kradalby: This should be `tagged-devices` like in `types/user.go`

cmd/headscale/cli/preauthkeys.go
nblock: Could you please update the description here?

nblock: Could you please update the description here?

cmd/headscale/cli/preauthkeys.go
```diff
@@ -142,11 +125,7 @@ var createPreAuthKeyCmd = &cobra.Command{
...
 
-		user, err := cmd.Flags().GetUint64("user")
-		if err != nil {
-			ErrorOutput(err, fmt.Sprintf("Error getting user: %s", err), output)
-		}
-
+		user, _ := cmd.Flags().GetUint64("user")
 		reusable, _ := cmd.Flags().GetBool("reusable")
 		ephemeral, _ := cmd.Flags().GetBool("ephemeral")
 		tags, _ := cmd.Flags().GetStringSlice("tags")
```
nblock, line 132: One can create preauthkeys with arbitrary tags: `headscale preauthkeys create --tags tag:this-is-not-referenced-anywhere`

Should this be possible or should we check `tagOwners` for a list of known tags?

hscontrol/state/state.go
kradalby: Fixed - now uses `types.TaggedDevices.Name` instead of the hardcoded string.

cmd/headscale/cli/preauthkeys.go
kradalby: Updated to "List all preauthkeys"

kradalby: Updated to "Creates a new preauthkey"

cmd/headscale/cli/preauthkeys.go
kradalby: This is intentional - preauthkeys are admin-level operations, so they can use any valid tag format (`tag:xxx`). The `tagOwners` policy governs which users/nodes can *request* tags via `--advertise-tags`, not which tags admins can assign via preauthkeys.

This allows flexibility in workflows (e.g., creating keys before updating policy) and matches how tagged preauthkeys already work. The tag format validation is done, but restricting to tagOwners-defined tags would be a separate policy decision.

If you think this should be restricted, we could open a follow-up issue to discuss the behavior?

nblock: I think this is fine, thx!

nblock:
I noticed `headscale preauthkeys list` no longer has a way to (optionally) filter for a user. All preauthkeys are displayed and since there is no "user" column one can no longer match preauthkeys to their user.

PR #2996, opened 2026-01-09
hscontrol: fix tag updates not propagating to node self view

### LINKED ISSUE #2978: [Bug] Tag update is not sent to node

### Current Behavior

Given this policy:

```json5
{
  "tagOwners": {
      "tag:bar": ["alice@"],
      "tag:foo": ["alice@"]
  },
  "acls": [
    {
      "action": "accept",
      "src": ["*"],
      "dst": ["*:*"]
    }
  ]
}
```

and a node tagged with `tag:foo`:

```bash
$ headscale nodes list -t
ID | Hostname | Name | MachineKey | NodeKey | User           | IP addresses                  | Ephemeral | Last seen           | Expiration          | Connected | Expired | ForcedTags | InvalidTags | ValidTags
1  | n2       | n2   | [M0Gk+]    | [Y0bYt] | tagged-devices | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2025-12-21 08:40:16 | 0001-01-01 00:00:00 | online    | no      | tag:foo    |             |     
```

it knows about its assigned tag:

```bash
$ tailscale status --json | jq .Self.Tags
[
  "tag:foo"
]
```

The tag can be modified on Headscale:

```bash
$ headscale nodes tag -i 1 -t "tag:bar"
Node updated
```

… and its visible in the nodes list:

```bash
$ headscale nodes list -t
ID | Hostname | Name | MachineKey | NodeKey | User           | IP addresses                  | Ephemeral | Last seen           | Expiration          | Connected | Expired | ForcedTags | InvalidTags | ValidTags
1  | n2       | n2   | [M0Gk+]    | [Y0bYt] | tagged-devices | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2025-12-21 08:40:16 | 0001-01-01 00:00:00 | online    | no      | tag:bar    |             |       
```

But its not visible by the node itself:

```bash
tailscale status --json | jq .Self.Tags
[
  "tag:foo"
]
```

Its only visible when tailscaled is restarted on the node.

### Expected Behavior

The node should have its list of tags updated as soon as its modified on Headscale.

### Steps To Reproduce

See above

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.28.0-beta1
- Tailscale version: 1.92.3
```

### Runtime environment

### Debug information

none

kradalby:
## Summary

When SetNodeTags changed a node's tags, the node's self view wasn't updated. The bug manifested as: the first SetNodeTags call updates the server but the client's self view doesn't update until a second call with the same tag.

Root cause: Three issues combined to prevent self-updates:

1. SetNodeTags returned PolicyChange which doesn't set OriginNode, so the mapper's self-update check failed.

2. The Change.Merge function didn't preserve OriginNode, so when changes were batched together, OriginNode was lost.

3. generateMapResponse checked OriginNode only in buildFromChange(), but PolicyChange uses RequiresRuntimePeerComputation which bypasses that code path entirely and calls policyChangeResponse() instead.

The fix addresses all three:
- state.go: Set OriginNode on the returned change
- change.go: Preserve OriginNode (and TargetNode) during merge
- batcher.go: Check OriginNode BEFORE RequiresRuntimePeerComputation

## Test plan

- [x] Add dedicated reproduction test `TestTagsIssue2978ReproTagReplacement` that validates the fix
- [x] Update 8 existing tests to verify node self view propagation after SetNodeTags
- [x] Add unit tests for Change.Merge preserving OriginNode/TargetNode

Fixes #2978

---
claude was used in this pr

### DIFF
--- a/hscontrol/mapper/mapper.go
+++ b/hscontrol/mapper/mapper.go
@@ -187,19 +187,28 @@ func (m *mapper) selfMapResponse(
 // - PeersChanged for remaining peers (their AllowedIPs may have changed due to policy)
 // - Updated PacketFilters
 // - Updated SSHPolicy (SSH rules may reference users/groups that changed)
+// - Optionally, the node's own self info (when includeSelf is true)
 // This avoids the issue where an empty Peers slice is interpreted by Tailscale
 // clients as "no change" rather than "no peers".
+// When includeSelf is true, the node's self info is included so that a node
+// whose own attributes changed (e.g., tags via admin API) sees its updated
+// self info along with the new packet filters.
 func (m *mapper) policyChangeResponse(
 	nodeID types.NodeID,
 	capVer tailcfg.CapabilityVersion,
 	removedPeers []tailcfg.NodeID,
 	currentPeers views.Slice[types.NodeView],
+	includeSelf bool,
 ) (*tailcfg.MapResponse, error) {
 	builder := m.NewMapResponseBuilder(nodeID).
 		WithDebugType(policyResponseDebug).
 		WithCapabilityVersion(capVer).
 		WithPacketFilters().
 		WithSSHPolicy()
+
+	if includeSelf {
+		builder = builder.WithSelfNode()
+	}
 
 	if len(removedPeers) > 0 {
 		// Convert tailcfg.NodeID to types.NodeID for WithPeersRemoved
--- a/hscontrol/mapper/batcher.go
+++ b/hscontrol/mapper/batcher.go
@@ -92,6 +92,11 @@ func generateMapResponse(nc nodeConnection, mapper *mapper, r change.Change) (*t
 		return nil, nil //nolint:nilnil // No response needed for other nodes when self-only
 	}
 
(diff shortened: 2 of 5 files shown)


hscontrol/mapper/mapper.go
```diff
@@ -187,20 +187,28 @@ func (m *mapper) selfMapResponse(
 // - PeersChanged for remaining peers (their AllowedIPs may have changed due to policy)
 // - Updated PacketFilters
 // - Updated SSHPolicy (SSH rules may reference users/groups that changed)
+// - Optionally, the node's own self info (when includeSelf is true)
 // This avoids the issue where an empty Peers slice is interpreted by Tailscale
 // clients as "no change" rather than "no peers".
+// When includeSelf is true, the node's self info is also included, which is
+// needed when the receiving node is the one whose tags changed (issue #2978).
```
kradalby: this comment feels specific and not like it will age well referring to the specific issue.

Issue #1803, opened 2024-03-01
Support ACL Tests


EtaoinWu:
## Why

ACL `tests` is a tailscale feature that allows ACL creators to do a sanity check on an ACL. It is a quality of life feature for access management.

```javascript
// in ACL.hujson
{
  // ... other config
  "tests": [
    {
      "src": "alice-pc",
      "proto": "tcp",
      "accept": [ "jump-server:22", "tag:printer:80" ],
      "deny": [ "bob-laptop:443" ],
    },
  ],
}
```

Headscale currently parses but [ignores](https://github.com/juanfont/headscale/blob/7a920ee701f6c1cc5152075bfcd7dae6f6d604c6/hscontrol/policy/acls_types.go#L40) ACL tests in an ACL config hujson or yaml file.

## Description

Implement ACL testing as described in tailscale's [KB](https://tailscale.com/kb/1337/acl-syntax#tests). Note that the struct `ACLTest` misses some fields as defined by tailscale.

wrbbz:
@kradalby , sorry for tagging you directly
Why is this issue closed as "not planned"? Is it somehow out of scope? Or just due to the inactivity?

It would be a nice to have feature. As it could help to create and maintain an ACLs

kradalby:
I opened it again, yes inactivity. But will like to point out that there is no space for this feature in the roadmap for a while. If someone wants to take a stab at it, this is a pretty good standalone part of the code base.

kradalby:
A new beta is out: [v0.29.0-beta.1](https://github.com/juanfont/headscale/releases/tag/v0.29.0-beta.1).

This feature is included in the release.

If you discover problems with it, open a new issue with full debug information: headscale config, policy, headscale logs at trace level, plus `tailscale debug netmap` and `tailscale status --json` from an affected client. See the [debugging and troubleshooting guide](https://headscale.net/development/ref/debug/) for collection steps and follow the issue template.

PR #3005, opened 2026-01-12
ACL testing (#1803)

Janhouse:
Fixing https://github.com/juanfont/headscale/issues/1803

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,6 +188,10 @@ sequentially through each stable release, selecting the latest patch version ava
 - Log ACME/autocert errors for easier debugging [#2933](https://github.com/juanfont/headscale/pull/2933)
 - Improve CLI list output formatting [#2951](https://github.com/juanfont/headscale/pull/2951)
 - Use Debian 13 distroless base images for containers [#2944](https://github.com/juanfont/headscale/pull/2944)
+- Add ACL testing functionality via CLI, API, and embedded policy tests [#3005](https://github.com/juanfont/headscale/pull/3005)
+  - New `headscale policy test` command to verify ACL rules
+  - New `POST /api/v1/policy/test` endpoint for third-party UI integration
+  - Support for `tests` section in policy files with automatic validation on policy updates
 - Fix ACL policy not applied to new OIDC nodes until client restart [#2890](https://github.com/juanfont/headscale/pull/2890)
 - Fix autogroup:self preventing visibility of nodes matched by other ACL rules [#2882](https://github.com/juanfont/headscale/pull/2882)
 - Fix nodes being rejected after pre-authentication key expiration [#2917](https://github.com/juanfont/headscale/pull/2917)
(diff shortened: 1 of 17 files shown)


Janhouse:
@kradalby said that maybe someone could give it a shot in #1803, so I did that. Golang is not my primary language so I did use Claude Code and tested with tests and on my running instance.
There are no integration tests though was not sure if they are needed here since it is mostly internal.

kradalby:
> @kradalby said that maybe someone could give it a shot in https://github.com/juanfont/headscale/issues/1803, so I did that. Golang is not my primary language so I did use Claude Code and tested with tests and on my running instance.

Nice feature to get in, its quite a large change, and we need to be careful about it as we do not want to provide users with a false sense of security or confidence, but definitely something to work towards. 

> There are no integration tests though was not sure if they are needed here since it is mostly internal.

I agree with this assessment, it should live more or less in the policy package and not need integration. 

That said, this will be an important one which we need test exhaustively. 

Another important question is, should we implement this exhaustively? Should we support everything from "day one", and should implementing new things in the policy be blocked on supporting acl tests?


It's quite a large change, and I do have a lot to do, but will try to look at it in the upcoming weeks, I already have quite a backlog of other large PRs queued up.

Janhouse:
> Another important question is, should we implement this exhaustively? Should we support everything from "day one", and should implementing new things in the policy be blocked on supporting acl tests?

I think it does not matter, this should not block adding new things in the policy. It uses the same filter rule resolution logic.

So if new autogroups or something like that are added to types.go with proper Resolve() implementation, it should still work fine. Of course if a new dimension is added to FilterRule then we would have to update it somewhat. But I don't expect that happening anytime soon.  Not even sure if Tailscale itself has something other than source, destination and protocol.

Of course more testing is needed.

I also added the UI part in headplane, see the screenshots in https://github.com/tale/headplane/pull/425

kradalby:
Next release will focus on `grants` and be policy oriented, so I think it is reasonable to work on this pr and that in tandem, so will get to it at that point.

kradalby:
There is some good potential here, as this release will be focused on policy stuff (mainly Grants), I think this would be good work to pursue, can you rebase it of the current main?

Janhouse:
> can you rebase it of the current main?

Rebased

kradalby:
I just had a refresher on this as I hope to get to it within a couple of weeks.

In general the test engine looks reasonable, but I do not think we should add the CLI/API part.
The main goal of the tests is to keep them running every time to prevent and understand undesired behaviour and not one-offs, so we should not encourage that with the CLI.

The other part is that the CLI/API looks quite fragile and large to maintain, and it just does not seem worth it.

The cycle for checking a test should be to add it to the ACL and try to load the acl, we can add a check mode so the whole thing is evaluate (we might have that but can make it more safe for tests).

Initial thoughts, I'll start poking at it at some point.

Janhouse:
I guess it can be improved over time but IMHO the API part was what gave the actual value.
Without well integrated UI it is not very useful.

PR #3011, opened 2026-01-12
Added migration for forced tags

### LINKED ISSUE #3006: [Bug] Migration from 0.27 to 0.28 only considers ForcedTags

### Current Behavior

Nodes with tags are migrated to `tagged-devices`  if they have `ForcedTags` set. This seems to be only the case when tags for a node were modified using the CLI/API.

- Register n1: `tailscale up --login-server https://headscale.example.com --advertise-tags=tag:bar`
- Register n2: `tailscale up --login-server https://headscale.example.com --advertise-tags=tag:foo`
- Set tags for n2: `headscale nodes tag -i 2 -t tag:bar`

Nodes list in 0.27.1:

```
ID | Hostname | Name | MachineKey | NodeKey | User  | IP addresses                  | Ephemeral | Last seen           | Expiration | Connected | Expired | ForcedTags | InvalidTags | ValidTags
1  | n1       | n1   | [2nA4Z]    | [nonFv] | alice | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2026-01-12 11:03:38 | N/A        | online    | no      |            |             | tag:bar  
2  | n2       | n2   | [6bOhO]    | [IeGXp] | alice | 100.64.0.2, fd7a:115c:a1e0::2 | false     | 2026-01-12 11:03:55 | N/A        | online    | no      | tag:bar    |             | tag:foo  
```

Nodes list after upgrade to a dev build of 0.28, 72fcb93e):

```
ID | Hostname | Name | MachineKey | NodeKey | User           | IP addresses                  | Ephemeral | Last seen           | Expiration | Connected | Expired | ForcedTags | InvalidTags | ValidTags
1  | n1       | n1   | [2nA4Z]    | [nonFv] | alice          | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2026-01-12 11:05:04 | N/A        | online    | no      |            |             |          
2  | n2       | n2   | [6bOhO]    | [IeGXp] | tagged-devices | 100.64.0.2, fd7a:115c:a1e0::2 | false     | 2026-01-12 11:05:04 | N/A        | online    | no      | tag:bar    |             |          
```

### Expected Behavior

Nodes with tags (probably: `ValidTags`, `ForcedTags`) should be migrated to `tagged-devices` during upgrade to 0.28.

Also observe that the tags in the `ValidTags` field are lost during upgrade.

### Steps To Reproduce

1. Register two tagged devices with 0.27.1
2. Set an additional tag for one of the nodes
3. Migrate to 0.28 (used 72fcb93e)
4. Observe that only one node is migrated as `tagged-devices` and some tags are lost

### Environment

```markdown
- OS: Debian 13
- Headscale version: main, 72fcb93e
- Tailscale version: 0.92.5
```

### Runtime environment

### Debug information

None.

shouryagautam:
fixes #3006

### DIFF
--- a/hscontrol/db/db.go
+++ b/hscontrol/db/db.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"net/netip"
-	"os"
 	"path/filepath"
 	"slices"
 	"strconv"
@@ -50,40 +49,6 @@ type HSDatabase struct {
 	DB       *gorm.DB
 	cfg      *types.Config
 	regCache *zcache.Cache[types.RegistrationID, types.RegisterNode]
-}
-
-// loadPolicyBytes loads policy from file or database based on configuration.
-// This is used during migrations when HSDatabase is not yet fully initialized.
-func loadPolicyBytes(tx *gorm.DB, cfg *types.Config) ([]byte, error) {
-	switch cfg.Policy.Mode {
-	case types.PolicyModeFile:
-		if cfg.Policy.Path == "" {
-			return nil, nil
-		}
-
-		absPath := util.AbsolutePathFromConfigPath(cfg.Policy.Path)
-
-		return os.ReadFile(absPath)
-
-	case types.PolicyModeDB:
-		p, err := GetPolicy(tx)
-		if err != nil {
-			if errors.Is(err, types.ErrPolicyNotFound) {
-				return nil, nil
-			}
-
-			return nil, err
-		}
-
-		if p.Data == "" {
-			return nil, nil
-		}
-
-		return []byte(p.Data), nil
-
-	default:
-		return nil, nil
-	}
 }
 
 // NewHeadscaleDatabase creates a new database connection and runs migrations.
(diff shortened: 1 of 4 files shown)


hscontrol/db/db.go
```diff
@@ -592,6 +592,89 @@ AND auth_key_id NOT IN (
...
+							}
+						}
+
+						// Merge RequestTags with existing tags
+						mergedTags := existingTags
+						for _, tag := range hostInfo.RequestTags {
+							if !slices.Contains(mergedTags, tag) {
+								mergedTags = append(mergedTags, tag)
+							}
+						}
```
kradalby: We cannot take these blindly, we need to validate them against the policy to ensure they are:
- The tag exists and can be used
- The owner of the node aligns with the tag owner.

shouryagautam: Ohkay, I will try checking the two conditions before appending

hscontrol/db/db.go
```diff
@@ -592,6 +593,178 @@ AND auth_key_id NOT IN (
...
+						return nil
+					}
+
+					if len(pol.TagOwners) == 0 {
+						log.Info().Msg("No tagOwners defined in policy, skipping RequestTags migration")
+						return nil
+					}
+
+					// Helper function to check if a user can have a tag
+					canUserHaveTag := func(username string, tag string) bool {
```
kradalby: I think that it is better to load up the proper policy manager machinery here so we do things exactly the same and not try to reimplement it. There is a lot of pitfalls where there are subtle ways to look up ownership that might be missed and this would be really hard to test. At the same time, there is logic for handling policies from file and database which both needs to be highlighted. 

I'll give this a go and show you a commit for it.

kradalby:
I'm pushing a version that uses the current machinery and handles the policy as a file aswell. Feel free to take a look!

Issue #3009, opened 2026-01-12
[Feature] Remove support bare wildcard in SSH dst


nblock:
### Use case

The [changelog for 0.26.0](https://github.com/juanfont/headscale/blob/72fcb93ef3ae99a5f9fe769a51ea0d50ef702b97/CHANGELOG.md?plain=1#L321) describes an inconsistency with SSH dest:

> There is one large inconsistency left, we allow `*` as a destination as we
currently do not support `autogroup:self`, `autogroup:member` and
`autogroup:tagged`. The support for `*` will be removed when we have support for
the autogroups.

Since autogroups are now available, should probably align the behavior with SaaS, [as per docs](https://tailscale.com/kb/1337/policy-syntax#dst-1):

> Specifies the destination (where the connection goes). The destination can be a tag, autogroup:self (if the source contains only users or groups), or a single named user (if the source contains only the same named user). The reason for these limitations is Tailscale does not let a user start a Tailscale SSH session on a user-owned device, unless the source is a device owned by the same user.
>
> You cannot specify a port for the destination because the only allowed port is 22. You cannot specify * as the destination.

### Description

Implement the the `dst` section of `ssh` policy rules like SaaS and remove the current bare wildcard `*` workaround.

### Contribution

- [ ] I can write the design doc for this feature
- [ ] I can contribute this feature

PR #3018, opened 2026-01-16
policy: validate SSH src/dst combinations per Tailscale rules

### LINKED ISSUE #3010: [Bug] Tailscale SSH from tagged device to user device should be denied

### Current Behavior

0.28 dev (tested with 72fcb93e) currently allows to `tailscale ssh` from a tagged device into a user-owned device. This should not be allowed as per [Tailscale docs](https://tailscale.com/kb/1337/policy-syntax#dst-1):

> Specifies the destination (where the connection goes). The destination can be a tag, autogroup:self (if the source contains only users or groups), or a single named user (if the source contains only the same named user). The reason for these limitations is Tailscale does not let a user start a Tailscale SSH session on a user-owned device, unless the source is a device owned by the same user.

### Expected Behavior

Refuse `tailscale ssh` from a tagged-device to a user-owned device, probably during policy loading time?

The policy below is refused by SaaS with the following error message: `Error: users in dst are only allowed from the same user`.

See also #3009.

### Steps To Reproduce

Given this policy:

```json
{
  "tagOwners": {
    "tag:bar": ["alice@"]
  },
  "acls": [
    {
      "action": "accept",
      "src": ["*"],
      "dst": ["*:22"]
    }
  ],
  "ssh": [
    {
      "action": "accept",
      "src": ["alice@"],
      "dst": ["tag:bar"],
      "users": ["root"]
    },
    {
      "action": "accept",
      "src": ["tag:bar"],
      "dst": ["alice@"],
      "users": ["root"]
    }
  ]
}
```

and the following list of nodes:

```
ID | Hostname | Name | MachineKey | NodeKey | User           | IP addresses                  | Ephemeral | Last seen           | Expiration | Connected | Expired | ForcedTags | InvalidTags | ValidTags
1  | n1       | n1   | [t+vzS]    | [Qvwvf] | tagged-devices | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2026-01-12 12:24:13 | N/A        | online    | no      | tag:bar    |             |          
2  | n2       | n2   | [Cywwx]    | [CILBB] | alice          | 100.64.0.2, fd7a:115c:a1e0::2 | false     | 2026-01-12 12:24:13 | N/A        | online    | no      |            |             |          
```

Allows ssh root login in all directions:

```console
root@n1:~# tailscale ssh root@100.64.0.2 hostname
n2
```
and
```console
root@n2:~# tailscale ssh root@100.64.0.1 hostname
n1
```

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.28 dev, 72fcb93e
- Tailscale version: 1.92.5
```

### Runtime environment

### Debug information

None.

kradalby:
Enforce Tailscale's SSH policy validation rules at policy load time:

- When dst contains a username, src must contain only the same username
- Tag/group/autogroup sources cannot SSH to user-owned devices
- Runtime filtering for `*` dst with tag sources (workaround until #3009)

Fixes #3010

### DIFF
--- a/hscontrol/policy/v2/filter.go
+++ b/hscontrol/policy/v2/filter.go
@@ -406,15 +406,7 @@ func (pol *Policy) compileSSHPolicy(
 		if len(otherDests) > 0 {
 			// TODO(kradalby): Remove this srcHasTags check when #3009 is completed
 			// and * is removed from SSH destinations entirely.
-			// Check if source contains any tags
-			srcHasTags := false
-
-			for _, src := range rule.Sources {
-				if _, ok := src.(*Tag); ok {
-					srcHasTags = true
-					break
-				}
-			}
+			srcHasTags := rule.Sources.ContainsTag()
 
 			// Build destination set for other destinations
 			var dest netipx.IPSetBuilder
--- a/hscontrol/policy/v2/filter_test.go
+++ b/hscontrol/policy/v2/filter_test.go
@@ -1782,10 +1782,28 @@ func TestSSHWithAsterixAndTagSource(t *testing.T) {
 	require.NoError(t, err)
 
 	// Tagged node should be able to receive SSH from other tagged nodes
-	if sshPolicy2 != nil {
-		require.GreaterOrEqual(t, len(sshPolicy2.Rules), 1,
-			"tagged node should have SSH rules allowing access from other tagged sources")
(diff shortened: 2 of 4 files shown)


kradalby:
Closing in favour of #3028

PR #3023, opened 2026-01-18
Added policy validation before nixos-rebuild switch

### LINKED ISSUE #2898: NixOS module: check policy as part of switch
Not sure if this is possible, but I think it is:

When I am running `nixes-rebuild switch` to change to a new configuration, before accepting a new Policy, we should run a `headscale policy validate` and reject the whole switch if the Policy isn't valid. 

This would allow admins to detect issues with the policy before the headscale instance is restarted. 

We could have a flag to bypass this validation for edge cases.

shouryagautam:
fixes #2898

### DIFF
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -20,6 +20,30 @@ let
   settingsFormat = pkgs.formats.yaml { };
   cliConfigFile = settingsFormat.generate "headscale.yaml" cliConfig;
 
+  shouldValidatePolicy =
+    cfg.settings.policy.mode == "file"
+    && cfg.settings.policy.path != null
+    && cfg.settings.policy.validatePolicy;
+
+  validatedPolicyFile =
+    if shouldValidatePolicy then
+      pkgs.runCommand "headscale-policy.hujson" { } ''
+        cp "${cfg.settings.policy.path}" "$out"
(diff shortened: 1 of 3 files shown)


nix/module.nix
```diff
@@ -653,6 +663,15 @@ in
...
     };
 
+    system.activationScripts.headscale-policy-check = lib.mkIf (
+      cfg.settings.policy.mode == "file"
+      && cfg.settings.policy.path != null
+      && cfg.settings.policy.validatePolicy
+    ) ''
+      # Validate headscale policy file
+      ${lib.getExe cfg.package} policy check -f "${cfg.settings.policy.path}"
+    '';
```
kradalby: I am not sure how this will work exactly, but I think this approach will not catch the issue as early as we want. I think this will let the whole Nix derivation build, and then only when it is activated it will fail. 

Ideally we find a way to test the configuration as a "post build" hook, so that the headscale build or a post-dependency of it fail, by using the newly built binary to validate as part of "building" the policy. That way we can fail early, even before trying to activate the system.

Maybe a way to achieve this is to make the policy its own derivation?

shouryagautam: I have tried adding policy validation as a derivation dependency

PR #3036, opened 2026-01-23
policy: add comprehensive Tailscale ACL compatibility tests

kradalby:
EXPERIMENT:

This PR contains a 9000~ line generated test file generated from having Claude agents run various ACL configuration against Tailscale SaaS and then observing the output Packet filters, and turn them into test cases.
They helped us discover several differences, we are addressing some of them in this PR, and some we will evaluate for later.

The tests can be treated as a "black box" test, they do not need to be reviewed and are not for humans to review. We use them for regression validation over time.

Add extensive test coverage verifying Headscale's ACL policy behavior
matches Tailscale's coordination server. Tests cover:

- Source/destination resolution for users, groups, tags, hosts, IPs
- autogroup:member, autogroup:tagged, autogroup:self behavior
- Filter rule deduplication and merging semantics
- Multi-rule interaction patterns
- Error case validation

Key behavioral differences documented:
- Headscale creates separate filter entries per ACL rule; Tailscale
  merges rules with identical sources
- Headscale deduplicates Dsts within a rule; Tailscale does not
- Headscale does not validate autogroup:self source restrictions for
  ACL rules (only SSH rules); Tailscale rejects invalid sources

Tests are based on real Tailscale coordination server responses
captured from a test environment with 5 nodes (1 user-owned, 4 tagged).

claude was used in this pr.

### DIFF
--- a/integration/route_test.go
+++ b/integration/route_test.go
@@ -1473,7 +1473,7 @@ func TestSubnetRouteACL(t *testing.T) {
 	wantClientFilter := []filter.Match{
 		{
 			IPProto: views.SliceOf([]ipproto.Proto{
-				ipproto.TCP, ipproto.UDP,
+				ipproto.TCP, ipproto.UDP, ipproto.ICMPv4, ipproto.ICMPv6,
 			}),
 			Srcs: []netip.Prefix{
 				netip.MustParsePrefix("100.64.0.1/32"),
@@ -1505,10 +1505,12 @@ func TestSubnetRouteACL(t *testing.T) {
 	}, 10*time.Second, 200*time.Millisecond, "Waiting for client packet filter to update")
 
 	// Wait for packet filter updates to propagate to subnet router netmap
+	// The two ACL rules (group:admins -> group:admins:* and group:admins -> 10.33.0.0/16:*)
+	// are merged into one filter rule since they share the same SrcIPs and IPProto.
 	wantSubnetFilter := []filter.Match{
 		{
 			IPProto: views.SliceOf([]ipproto.Proto{
-				ipproto.TCP, ipproto.UDP,
+				ipproto.TCP, ipproto.UDP, ipproto.ICMPv4, ipproto.ICMPv6,
 			}),
 			Srcs: []netip.Prefix{
 				netip.MustParsePrefix("100.64.0.1/32"),
@@ -1525,20 +1527,6 @@ func TestSubnetRouteACL(t *testing.T) {
 					Net:   netip.MustParsePrefix("fd7a:115c:a1e0::1/128"),
 					Ports: allPorts,
 				},
-			},
-			Caps: []filter.CapMatch{},
-		},
-		{
-			IPProto: views.SliceOf([]ipproto.Proto{
-				ipproto.TCP, ipproto.UDP,
-			}),
-			Srcs: []netip.Prefix{
-				netip.MustParsePrefix("100.64.0.1/32"),
-				netip.MustParsePrefix("100.64.0.2/32"),
-				netip.MustParsePrefix("fd7a:115c:a1e0::1/128"),
-				netip.MustParsePrefix("fd7a:115c:a1e0::2/128"),
-			},
-			Dsts: []filter.NetPortRange{
 				{
 					Net:   netip.MustParsePrefix("10.33.0.0/16"),
 					Ports: allPorts,


hscontrol/policy/v2/types.go
```diff
@@ -1679,6 +1686,51 @@ func validateSSHSrcDstCombination(sources SSHSrcAliases, destinations SSHDstAlia
 	return nil
 }
 
+// validateACLSrcDstCombination validates that ACL source/destination combinations
+// follow Tailscale's security model:
+// - autogroup:self destinations require ALL sources to be users, groups, autogroup:member, or wildcard (*)
+// - Tags, autogroup:tagged, hosts, and raw IPs are NOT valid sources for autogroup:self
+// - Wildcard (*) is allowed because autogroup:self evaluation narrows it per-node to the node's own IPs.
+func validateACLSrcDstCombination(sources Aliases, destinations []AliasWithPorts) error {
```
kradalby, line 1694: This sounds like it is a breaking change, and needs a changelog entry.

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## 0.29.0 (202x-xx-xx)
 
+### Tailscale ACL compatibility improvements
+
+Tests generated using Tailscale's mts.go test harness to run virtual nodes, set ACL policies
+via tscli, and capture PacketFilter from each node's netmap. Several bugs were found and fixed.
+[#3036](https://github.com/juanfont/headscale/pull/3036)
```
kradalby, line 10: This description is uncorrect.

mts is not a test harness. 

It should say something like, we have systematically generated an extensive test of test cases by using tailscale clients and the official SaaS to understand how the packet filter should be generated. 
We discovered a few differences, but overall our implementation was very close.

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+
+### BREAKING
+
+- **ACL Policy**: Wildcard (`*`) in ACL sources and destinations now resolves to Tailscale's CGNAT range (`100.64.0.0/10`) and ULA range (`fd7a:115c:a1e0::/48`) instead of all IPs (`0.0.0.0/0` and `::/0`) [#3036](https://github.com/juanfont/headscale/pull/3036)
+  - This better matches Tailscale's security model where `*` means "any node in the tailnet" rather than "any IP address"
+  - Policies relying on wildcard to match non-Tailscale IPs will need to use explicit CIDR ranges instead
+
+### Changes
+
+- **ACL Policy**: Validate autogroup:self source restrictions matching Tailscale behavior - tags, hosts, and IPs are rejected as sources for autogroup:self destinations [#3036](https://github.com/juanfont/headscale/pull/3036)
```
kradalby: Sounds breaking.

hscontrol/policy/v2/types.go
```diff
@@ -1344,42 +1349,44 @@ func (p Protocol) Description() string {
...
+		return []int{ProtocolUDP}, false
+	case ProtoNameGRE:
+		return []int{ProtocolGRE}, true
+	case ProtoNameESP:
+		return []int{ProtocolESP}, true
+	case ProtoNameAH:
+		return []int{ProtocolAH}, true
+	case ProtoNameSCTP:
+		return []int{ProtocolSCTP}, false
+	case ProtoNameICMP:
```
kradalby: ProtoName -> ProtocolName

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
 
+### Tailscale ACL compatibility improvements
+
+Tests generated using Tailscale's mts.go test harness to run virtual nodes, set ACL policies
+via tscli, and capture PacketFilter from each node's netmap. Several bugs were found and fixed.
+[#3036](https://github.com/juanfont/headscale/pull/3036)
+
+### BREAKING
+
+- **ACL Policy**: Wildcard (`*`) in ACL sources and destinations now resolves to Tailscale's CGNAT range (`100.64.0.0/10`) and ULA range (`fd7a:115c:a1e0::/48`) instead of all IPs (`0.0.0.0/0` and `::/0`) [#3036](https://github.com/juanfont/headscale/pull/3036)
```
kradalby, line 14: This will break all users that has non standard ip ranges in `prefixes.ipv4` and `prefixes.ipv6`. This is already not supported (as per comment in config-example.yaml) and a warning, but it will break them and they will have to explicitly set their CIDR instead of using `*`. I think this is the right direction as we are moving away from that, but we need to explain it.

hscontrol/policy/matcher/matcher.go
```diff
@@ -93,11 +93,24 @@ func (m *Match) DestsOverlapsPrefixes(prefixes ...netip.Prefix) bool {
...
+		return true
+	}
+
+	// Check if dests contains all prefixes of TheInternet (superset check)
+	theInternet := util.TheInternet()
+	for _, prefix := range theInternet.Prefixes() {
+		if !m.dests.ContainsPrefix(prefix) {
+			return false
+		}
+	}
```
kradalby, line 113: Is this strictly necessary?

kradalby, line 113: Is this strictly necessary?

Issue #2990, opened 2026-01-05
[Bug] autogroup:self seems to cover other ACL rules


danieldietzler:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

```json
{
  "groups": {
    "group:admin": [...]
  },
  "acls": [
    {
      "action": "accept",
      "src": ["group:admin"],
      "dst": ["*:*"]
    },
    {
      "action": "accept",
      "src": ["autogroup:member"],
      "dst": ["autogroup:self:*"]
    },
  ],
}
```
does not give me access to any tagged devices, even though I should have access to any device on the tailnet because of the first ACL rule.

### Expected Behavior

When part of the admin group, I should keep access to all devices, even with the extra rule
```json
    {
      "action": "accept",
      "src": ["autogroup:member"],
      "dst": ["autogroup:self:*"]
    },
```

### Steps To Reproduce

Use the aforementioned rules, replace the `...` in the admin group with your user.

### Environment

```markdown
- OS: Debian 13.2
- Headscale version: v0.28.0-beta.1
- Tailscale version: 1.92.3
```

### Runtime environment

- [x] Headscale is behind a (reverse) proxy
- [x] Headscale runs in a container

### Debug information

Same symptoms as in #2838

nblock:
Confirmed on 72fcb93ef3ae99a5f9fe769a51ea0d50ef702b97. 

Nodes n1 and n2 can communicate with each other while connections to n3 are blocked with this policy:

```json
{
  "groups": {
    "group:admin": ["alice@"]
  },
  "tagOwners": {
    "tag:foo": [
      "group:admin"
    ]
  },
  "acls": [
    {
      "action": "accept",
      "src": ["group:admin"],
      "dst": ["*:*"]
    },
    {
      "action": "accept",
      "src": ["autogroup:member"],
      "dst": ["autogroup:self:*"]
    }
  ]
}
```

```
$ headscale nodes list -t
ID | Hostname | Name | MachineKey | NodeKey | User           | IP addresses                  | Ephemeral | Last seen           | Expiration | Connected | Expired | ForcedTags | InvalidTags | ValidTags
1  | n1       | n1   | [utXwe]    | [3cter] | alice          | 100.64.0.1, fd7a:115c:a1e0::1 | false     | 2026-01-11 09:43:09 | N/A        | online    | no      |            |             |          
[...]

PR #3044, opened 2026-01-26
policy/v2: include nodes with empty filters in BuildPeerMap

kradalby:
Remove the `len(filter) == 0` skip condition so all nodes are included in nodeMatchers. This ensures symmetric visibility works correctly when using autogroup:self with multiple user groups.

Updates #2990


kradalby:
confirmed working by @danieldietzler on gh issue and Nathaniel on Discord.

Issue #3055, opened 2026-02-04
Add more explicit warning against using a non standard IP prefix


kradalby:
Headscale has these historic configuration options for changing the prefixes from which clients are assigned IP addresses, `prefixes.v4` and `prefixes.v6`. 

This was added at a time, very early on, where we thought that the Tailscale client would accept any range and we never checked if this was actually the case. It turns out that there are a lot of assumptions and implicit and explicit checks in the client that really depend on the use of the CGNAT range for IPv4 and ULA range for IPv6. 

Even as we discovered this, we did not remove the option, but added a comment that it is _not_ supported:
https://github.com/juanfont/headscale/blob/a09b0d1d6919058de6ed58557032bcd7a671bb22/config-example.yaml#L58

This has not deterred all users from using this option, and as they change their configuration and the set of feature they use _or_ the Tailscale client changes, this might break. Not only does it break, but it causes a class of really hard to debug issues that putting limits on a system help us avoid, which is likely why Tailscale only uses a single range, and does not provide any option to configure it. 

While I will still not suggest that we _remove_ the option, we need to be more explicit about how this is **not** supported and how this might break now or in the future. 

We should add a very explicit warning to discourage the use of any IP range that is outside of the default ranges. The option can still be useful for using _a smaller_ (e.g. subset) of the default range. 

We could do a more explicit warning like:

```
################################################################
###      __          __              _             _         ###
###      \ \        / /             (_)           | |        ###
###       \ \  /\  / /_ _ _ __ _ __  _ _ __   __ _| |        ###
###        \ \/  \/ / _` | '__| '_ \| | '_ \ / _` | |        ###
###         \  /\  / (_| | |  | | | | | | | | (_| |_|        ###
###          \/  \/ \__,_|_|  |_| |_|_|_| |_|\__, (_)        ###
###                                           __/ |          ###
###                                          |___/           ###
################################################################
###                                                          ###
###  You have overriden any of the Headscale `prefixes`,     ###
###  which is not supported. Please revert them back to      ###
###  the standard CGNAT and ULA range as described in the    ###
###  example configuration.                                  ###
###                                                          ###
###  Running Headscale with this configuration will cause    ###
###  undefined behaviour as the Tailscale client is NOT      ###
###  made to run on any other ranges and there are explicit  ###
###  and implicit parts of the code that will cause trouble. ###
###                                                          ###
###  Any issue raised using a range outside of the supported ###
###  range will be labelled as wontfix and closed.           ###
[...]

PR #3062, opened 2026-02-05
docs(config): add explicit prefixes warning

LouisLau-art:
Fixes juanfont/headscale#3055.

Adds a much more explicit warning block above `prefixes` in `config-example.yaml`:

- Clarifies that overriding `prefixes` to non-default ranges is not supported.
- Calls out that Tailscale clients assume CGNAT/ULA and may break now or in future.
- Notes that using a *subset* of the defaults is fine.
- Links to the relevant `tsaddr` checks in the Tailscale client.

### DIFF
--- a/hscontrol/types/config.go
+++ b/hscontrol/types/config.go
@@ -856,7 +856,9 @@ func prefixV4() (*netip.Prefix, error) {
 	ipSet, _ := builder.IPSet()
 	if !ipSet.ContainsPrefix(prefixV4) {
 		log.Warn().
-			Msgf("Prefix %s is not in the %s range. This is an unsupported configuration.",
+			Msgf("`prefixes.v4` is set to %s, which is outside the supported %s range. "+
+				"Headscale only supports prefixes within this range (including subsets); "+
+				"custom ranges are unsupported and may break client connectivity.",
 				prefixV4Str, tsaddr.CGNATRange())
 	}
 
@@ -881,7 +883,9 @@ func prefixV6() (*netip.Prefix, error) {
 
 	if !ipSet.ContainsPrefix(prefixV6) {
 		log.Warn().
-			Msgf("Prefix %s is not in the %s range. This is an unsupported configuration.",
+			Msgf("`prefixes.v6` is set to %s, which is outside the supported %s range. "+
+				"Headscale only supports prefixes within this range (including subsets); "+
+				"custom ranges are unsupported and may break client connectivity.",
 				prefixV6Str, tsaddr.TailscaleULARange())
 	}
 


nblock:
Thx, let's consider this when a related code change is merged.

LouisLau-art:
Understood, thanks. I’ll keep this open and can rebase or adjust quickly once the related code change lands.

kradalby:
@LouisLau-art do you want to give the code change a go?
I think there already is a warning in the code, so should be to replace it/improve it.

LouisLau-art:
Yes — I’ll take the code-change follow-up on this PR.
I’ll update/improve the existing warning in code (not only docs), then push an update for review.

LouisLau-art:
Implemented the code-side follow-up in this PR (commit `62409aa9`).

What changed:
- Updated runtime warnings in `prefixV4()` and `prefixV6()` (`hscontrol/types/config.go`)
- Warnings now explicitly reference `prefixes.v4` / `prefixes.v6`
- Clarified that only prefixes within CGNAT/ULA ranges are supported (including subsets)
- Added impact wording that custom ranges are unsupported and may break client connectivity

Validation:
- `go test ./hscontrol/types -count=1`

config-example.yaml
```diff
@@ -48,14 +48,29 @@ noise:
...
-# client - i.e., subnets of 100.64.0.0/10 and fd7a:115c:a1e0::/48.
-# See below:
-# IPv6: https://github.com/tailscale/tailscale/blob/22ebb25e833264f58d7c3f534a8b166894a89536/net/tsaddr/tsaddr.go#LL81C52-L81C71
-# IPv4: https://github.com/tailscale/tailscale/blob/22ebb25e833264f58d7c3f534a8b166894a89536/net/tsaddr/tsaddr.go#L33
-# Any other range is NOT supported, and it will cause unexpected issues.
+# Each prefix consists of either an IPv4 or IPv6 address, and the associated prefix
+# length, delimited by a slash.
+#
+# -------------------------------------------------------------------------------
+# WARNING: OVERRIDING `prefixes` IS NOT SUPPORTED (except for taking a subset).
```
kradalby, line 55: I think we need to change the wording a bit here to reflect that it is supported to change it, just not to something outside of the CGNAT/ULA range, I think we could use the message as proposed in the issue as it explains the case as well.

hscontrol/types/config.go
```diff
@@ -856,7 +856,9 @@ func prefixV4() (*netip.Prefix, error) {
 	ipSet, _ := builder.IPSet()
 	if !ipSet.ContainsPrefix(prefixV4) {
 		log.Warn().
-			Msgf("Prefix %s is not in the %s range. This is an unsupported configuration.",
+			Msgf("`prefixes.v4` is set to %s, which is outside the supported %s range. "+
+				"Headscale only supports prefixes within this range (including subsets); "+
+				"custom ranges are unsupported and may break client connectivity.",
```
kradalby, line 861: I think we should print the full formatted warning from the issue, we could make a reusable function that prints the whole thing from the issue and make sure it formats nicely when we print it. 

We should only print it once if either is changed, so could make this one if and print.

Bonus points would be to create a "print large warning" function that prints the warning banner and formats the text nicely inside it so we could reuse it for other things where we feel like we want to show a large warning explicit.

TanayK07:
Hey @LouisLau-art — I saw this PR and the review feedback from @kradalby, and since it looked like it had gone quiet for a bit, I went ahead and took a crack at it in #3131. I tried to address the review comments here (reusable banner function, single warning even if both prefixes are non-standard, updated config-example wording).

Totally understand if you'd rather pick this back up yourself though — happy to close mine if so. Just wanted to make sure the issue got some momentum since it's on the v0.29.0 milestone. Cheers!

PR #3078, opened 2026-02-11
fix(auth): honor ephemeral registration for interactive auth

### LINKED ISSUE #2719: [Bug] Unable to register ephemeral nodes using OIDC

### Current Behavior

I am trying to create ephemeral node (Windows 10) by using OIDC.
As per Tailscale documentation that should be possible in 2 ways:

1. using preauth ephemeral keys
2. starting tailscaled with option "--state=mem:"

I would like to avoid preshared key in order to use 2FA provided by OIDC (Keycloak), the other problem with 1. is that I already have to have user registered to create preauth key (on first login Keycloak is set up to enable and configure 2FA).
I tried option 2. by starting tailscaled on windows with option "--state=mem:" but when I log in to headscale newly registered node is not ephemeral.

### Expected Behavior

After starting tailscaled with option "--state=mem:" on a node and registering a node with headscale, node should be registered as ephemeral.

### Steps To Reproduce

- start headscale server
- start tailscaled on client machine:  tailscaled --state=mem:
- run on client machine: tailscale login --login-server https://ts.example.org --accept-routes
- go to URL returned by previous command and authenticate
- headscale nodes list shows: 
- 26 | NODE-030    | node-030    | [5UOSB]    | [fVQNN] | test-user | 100.64.0.9, fd7a:115c:a1e0::9 | **false**     | 2025-08-06 06:49:56 | 2026-02-02 06:44:34 | online   | no

### Environment

```markdown
- OS: Debian 12
- Headscale version: 0.26.1
- Tailscale version: 1.86.2
```

### Runtime environment

### Debug information

I know that there is disparity between headscale and Tailscale features, so my question is if this possibility is missing from headscale, is there  some configuration needed, or is it a bug ?

LouisLau-art:
## Summary

Fixes `tailscaled --state=mem:` registrations that go through interactive auth (OIDC/web), so nodes are persisted as ephemeral and handled consistently.

Closes #2719.

## What changed

- Add `nodes.ephemeral` as a node-level persisted flag.
- Propagate `RegisterRequest.Ephemeral` into interactive registration cache entries.
- Carry that flag through auth callback (`HandleNodeFromAuthPath`) into created/updated nodes.
- Keep pre-auth key behavior unchanged by mirroring `PreAuthKey.Ephemeral` into `nodes.ephemeral`.
- Update ephemeral lookups/counting to use node-level `IsEphemeral()` checks.
- Add DB migration to add/backfill `nodes.ephemeral` from existing ephemeral pre-auth keys.
- Extend tests:
  - interactive workflow with `Ephemeral: true` now asserts node is ephemeral;
  - DB ephemeral listing includes nodes with `nodes.ephemeral = true` and no pre-auth key.

## Verification

- `go test ./hscontrol -run TestAuthenticationFlows -count=1`
- `go test ./hscontrol/db -run TestListEphemeralNodes -count=1`
- `go test ./hscontrol/state -run TestEphemeralNodeLogoutRaceCondition -count=1`
- `go test ./hscontrol/db -run 'TestSQLiteMigrationAndDataValidation|TestSQLiteAllTestdataMigrations' -count=1`
- `go test ./hscontrol/types -run TestNodeRegisterMethodToV1Enum -count=1`

### DIFF
--- a/hscontrol/auth_test.go
+++ b/hscontrol/auth_test.go
@@ -342,6 +342,35 @@ func TestAuthenticationFlows(t *testing.T) {
 			validateCompleteResponse: true,
 			expectedAuthURLPattern:   "/register/",
 		},
+		{
+			name: "full_interactive_workflow_ephemeral_node",
+			setupFunc: func(t *testing.T, app *Headscale) (string, error) {
+				return "", nil
+			},
+			request: func(_ string) tailcfg.RegisterRequest {
+				return tailcfg.RegisterRequest{
+					NodeKey:   nodeKey1.Public(),
+					Ephemeral: true,
+					Hostinfo: &tailcfg.Hostinfo{
+						Hostname: "interactive-ephemeral-node",
+					},
+					Expiry: time.Now().Add(24 * time.Hour),
+				}
+			},
+			machineKey:              func() key.MachinePublic { return machineKey1.Public() },
+			requiresInteractiveFlow: true,
+			interactiveSteps: []interactiveStep{
+				{stepType: stepTypeInitialRequest, expectAuthURL: true, expectCacheEntry: true},
+				{stepType: stepTypeAuthCompletion, callAuthPath: true, expectCacheEntry: false},
+			},
+			validateCompleteResponse: true,
+			expectedAuthURLPattern:   "/register/",
+			validate: func(t *testing.T, _ *tailcfg.RegisterResponse, app *Headscale) {
+				node, found := app.state.GetNodeByNodeKey(nodeKey1.Public())
+				require.True(t, found)
+				assert.True(t, node.IsEphemeral(), "interactive node with state=mem should be marked ephemeral")
+			},
+		},
 		// TEST: Interactive workflow with no Auth struct in request
 		// WHAT: Tests interactive flow when request has no Auth field (nil)
 		// INPUT: Register request with Auth field set to nil
--- a/hscontrol/db/db.go
+++ b/hscontrol/db/db.go
@@ -352,6 +352,7 @@ AND auth_key_id NOT IN (
   given_name varchar(63),
   user_id integer,
   register_method text,
+  ephemeral numeric DEFAULT false,
   forced_tags text,
   auth_key_id integer,
   last_seen datetime,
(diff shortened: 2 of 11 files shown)


hscontrol/db/db.go
kradalby, line 248: This changes a previous migration, which we cannot do, we can only add migrations.

hscontrol/db/db.go
```diff
@@ -699,6 +700,32 @@ AND auth_key_id NOT IN (
 				},
 				Rollback: func(db *gorm.DB) error { return nil },
 			},
+			{
+				ID: "202602111000-node-ephemeral-column",
```
kradalby, line 704: I think this is the right approach, the `ephemeral` field should be on the node and not the auth key. This is essentially the same change as we did for `tags` in the previous release.

At authentication, the `ephemeral` field should be set on the node, and from there on, the pre auth key should _always_ be ignored.

hscontrol/db/node.go
```diff
@@ -100,7 +100,10 @@ func (hsdb *HSDatabase) ListEphemeralNodes() (types.Nodes, error) {
 	return Read(hsdb.DB, func(rx *gorm.DB) (types.Nodes, error) {
 		nodes := types.Nodes{}
 
-		err := rx.Joins("AuthKey").Where(`"AuthKey"."ephemeral" = true`).Find(&nodes).Error
+		err := rx.
+			Joins("LEFT JOIN pre_auth_keys pak ON pak.id = nodes.auth_key_id").
+			Where("nodes.ephemeral = ? OR pak.ephemeral = ?", true, true).
+			Find(&nodes).Error
```
kradalby, line 106: This change should be reverted and the authkey part should be removed. We should _only_ be looking at the `ephemeral` field on the node, so no join should be necessary.

hscontrol/state/state.go
```diff
@@ -1314,6 +1316,7 @@ func (s *State) createAndSaveNewNode(params newNodeParams) (types.NodeView, erro
 
 		nodeToRegister.AuthKey = params.PreAuthKey
 		nodeToRegister.AuthKeyID = &params.PreAuthKey.ID
+		nodeToRegister.Ephemeral = params.PreAuthKey.Ephemeral
```
kradalby, line 1319: This should be the last place `PreAuthKey.Ephemeral` is used, and after that, the Node field is the "authority".

hscontrol/types/node.go
```diff
@@ -181,7 +182,7 @@ func (node *Node) IsExpired() bool {
 // IsEphemeral returns if the node is registered as an Ephemeral node.
 // https://tailscale.com/kb/1111/ephemeral-nodes/
 func (node *Node) IsEphemeral() bool {
-	return node.AuthKey != nil && node.AuthKey.Ephemeral
+	return node.Ephemeral || (node.AuthKey != nil && node.AuthKey.Ephemeral)
```
kradalby, line 185: This can drop `|| (node.AuthKey != nil && node.AuthKey.Ephemeral)` as it is only used at auth time.

hscontrol/auth_test.go
```diff
@@ -342,6 +342,35 @@ func TestAuthenticationFlows(t *testing.T) {
 			validateCompleteResponse: true,
 			expectedAuthURLPattern:   "/register/",
 		},
+		{
+			name: "full_interactive_workflow_ephemeral_node",
```
kradalby, line 346: We need an integration test for the webauth and oidc auth to use `--state=mem:` to validate the behaviour.

kradalby (commented):
Looks pretty good, some comments and things needed.

If you are assisted by an AI, remember to declare the agent and model.

PR #3089, opened 2026-02-18
hscontrol: enforce that tagged nodes never have user_id

### LINKED ISSUE #3077: Migrating to tagged devices leaves user_id set

### Current Behavior

I just updated from 0.27.1 to 0.28.0 and assigned tags to several devices which lead to migration to "tagged-devices". As there was no device left for legacy-user1 and legacy-user2 (not linked to oidc), i tried to delete them. CLI gives error `Cannot destroy user: user not empty: node(s) found` because in the database the tagged devices are still in relation with the corresponding user_id.

Image

I then tried to reassign the tag with `headscale nodes tag -2 -t tag:name` but the user_id persists and the legacy users cannot be deleted.

### Expected Behavior

Legacy users that are migrated to tagged-devices can be deleted without an error message.

### Steps To Reproduce

Create user and assign devices, then add tag to migrate these users. Try to delete existing user without node assgned.

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.28.0
- Tailscale version: 1.90.4
```

### Runtime environment

### Debug information

-

kradalby:
Tagged nodes are owned by their tags, not a user. Enforce this by setting `user_id = NULL` on tagged nodes — both via a migration for existing data and at every write path going forward.

This makes `DestroyUser` work naturally since tagged nodes no longer reference any user.

Fixes #3077

claude-opus-4 was used in this PR.

### DIFF
--- a/hscontrol/state/state.go
+++ b/hscontrol/state/state.go
@@ -1680,6 +1680,30 @@ func (s *State) createNewNodeFromAuth(
 }
 
 // HandleNodeFromPreAuthKey handles node registration using a pre-authentication key.
+// findExistingNodeForPAK looks up an existing node by machine key,
+// matching the PAK's ownership. For user-owned keys it checks the
+// user's ID; for tagged keys it checks UserID(0) since tagged nodes
+// have no owning user.
+func (s *State) findExistingNodeForPAK(
+	machineKey key.MachinePublic,
+	pak *types.PreAuthKey,
+) (types.NodeView, bool) {
+	if pak.User != nil {
+		node, exists := s.nodeStore.GetNodeByMachineKey(machineKey, types.UserID(pak.User.ID))
+		if exists {
+			return node, true
+		}
+	}
+
+	// Tagged nodes have nil UserID, so they are indexed under UserID(0)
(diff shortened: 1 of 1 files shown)


hscontrol/state/state.go
```diff
@@ -1871,17 +1873,20 @@ func (s *State) HandleNodeFromPreAuthKey(
 
 		// For user-owned keys, check if node exists for a different user
 		// For tags-only keys (pak.User == nil), this check is skipped
-		if pak.User != nil && existsAnyUser && existingNodeAnyUser.Valid() && existingNodeAnyUser.UserID().Get() != pak.User.ID {
-			// Node exists but belongs to a different user
-			// Create a NEW node for the new user (do not transfer)
-			// This allows the same machine to have separate node identities per user
-			oldUser := existingNodeAnyUser.User()
+		if pak.User != nil && existsAnyUser && existingNodeAnyUser.Valid() && !existingNodeAnyUser.IsTagged() && existingNodeAnyUser.UserID().Get() != pak.User.ID {
```
kradalby: This is a really gnarly if statement that is very hard to read, can we break it down?

PR #3106, opened 2026-02-25
fix(policy/v2): use TrimPrefix for leading @ in username resolution

### LINKED ISSUE #3104: [Bug] policy/v2: Username.resolveUser uses TrimSuffix instead of TrimPrefix for @ prefix
## Bug Description

`Username.resolveUser` in `hscontrol/policy/v2/types.go` uses `strings.TrimSuffix` to strip the `@` character from username tokens, but the headscale ACL policy format uses a **leading** `@` (e.g., `@username`), not a trailing one.

This means `resolveUser` never matches any user, causing **all group-based ACL rules to resolve to empty source IP sets**, which makes them silently skipped during filter compilation.

## Impact

- All group-based ACL source rules produce no filter rules
- `/debug/filter` returns `null` (empty filter set)
- `ReduceRoutes` and `BuildPeerMap` see no matchers, so subnet routes are never distributed to group members

## Root Cause

```go
// hscontrol/policy/v2/types.go, line ~240
func (u *Username) resolveUser(users types.Users) (types.User, error) {
    // ...
    uTrimmed := strings.TrimSuffix(u.String(), "@")  // BUG: should be TrimPrefix
    // ...
}
```

The ACL policy format is `"@username"` (leading `@`), but `TrimSuffix` only strips a trailing `@`. The result:
- `u.String()` = `"@user1"`
- `uTrimmed` = `"@user1"` (unchanged — no trailing `@` to strip)
- `user.Name == "@user1"` → **never matches** (user names are stored without `@`)

## Fix

```go
uTrimmed := strings.TrimPrefix(u.String(), "@")
```

## Reproduction

1. Create multiple users and a group-based ACL policy:
   ```json
   {
     "groups": { "group:test": ["@user1"] },
     "acls": [{ "action": "accept", "src": ["group:test"], "dst": ["10.0.0.0/8:*"] }]
   }
   ```
2. Register a subnet router advertising `10.0.0.0/8`
3. Connect as `user1`
4. Observe: router does not appear in netmap, `/debug/filter` returns `null`

## Environment

- headscale version: v0.27.1 (commit 8048f10d)
- The bug appears to be present since the v2 policy engine was introduced

lokkju:
## Summary

- Fix `Username.resolveUser` to strip the **leading** `@` from usernames using `TrimPrefix` instead of `TrimSuffix`
- The ACL policy format uses `@username` (leading `@`), but `TrimSuffix` only strips a trailing `@`, so no group member usernames ever matched, causing all group-based ACL rules to silently compile to empty filter sets

## Root Cause

```go
// Before (bug): trailing @ stripped, but policy uses leading @
uTrimmed := strings.TrimSuffix(u.String(), "@")
// "@laronica.lawson" → "@laronica.lawson" (unchanged, never matches)

// After (fix): strip both leading and trailing @
uTrimmed := strings.TrimPrefix(strings.TrimSuffix(u.String(), "@"), "@")
// "@laronica.lawson" → "laronica.lawson" (matches user.Name)
```

## Impact

This bug caused:
- All group-based ACL source rules to produce empty IP sets (silently skipped)
- `/debug/filter` returning `null` when only group-based rules exist
- `ReduceRoutes` and `BuildPeerMap` seeing no matchers, so subnet routes were never distributed to group members
- The `src: *` workaround bypassed the bug because wildcard sources don't go through `resolveUser`

## Verification

Tested on a live headscale v0.27.1 deployment with:
- A subnet router advertising `10.0.0.0/8` and a 4via6 prefix
- Group-based ACLs (`group:n8n → n8n-svc:5678`)
- Before fix: `/debug/filter` = `null`, router not visible in peer netmap
- After fix: filter rules correctly compiled, router visible with scoped routes, end-to-end connectivity confirmed

All existing policy tests pass.

Fixes #3104

### DIFF
--- a/hscontrol/policy/v2/types.go
+++ b/hscontrol/policy/v2/types.go
@@ -233,11 +233,11 @@ func (u *Username) CanBeAutoApprover() bool {
 func (u *Username) resolveUser(users types.Users) (types.User, error) {
 	var potentialUsers types.Users
 
-	// At parsetime, we require all usernames to contain an "@" character, if the
-	// username token does not naturally do so (like email), the user have to
-	// add it to the end of the username. We strip it here as we do not expect the
-	// usernames to be stored with the "@".
-	uTrimmed := strings.TrimSuffix(u.String(), "@")
+	// At parsetime, we require all usernames to contain an "@" character.
+	// The policy format uses a leading "@" (e.g., "@username"), while email-style
+	// identifiers have it naturally. Strip both leading and trailing "@" since
+	// usernames are stored without the "@" prefix.
+	uTrimmed := strings.TrimPrefix(strings.TrimSuffix(u.String(), "@"), "@")
 
 	for _, user := range users {
 		if user.ProviderIdentifier.Valid && user.ProviderIdentifier.String == uTrimmed {


kradalby:
This is not discussed in an issue and the pr states what it tries to fix, but not why it does it.

Please provide full usecases and why this is needed in an issue, and use a human voice.

lokkju:
It absolutely is discussed in a bug, which was linked in this PR: #3104

PR #3109, opened 2026-02-26
types: avoid NodeView clone in CanAccess

### LINKED ISSUE #3108: [Bug] Performance regression after upgrade 0.26.1 to 0.27.1

### Current Behavior

After upgrading Headscale from 0.26.1 to 0.27.1, CPU usage jumped under the same workload (~80 connected clients, same config/policy). Overall CPU usage went from ~5% to ~60% even when nothing really happened (no active reconnects, etc).

Upgrading to 0.28.0 did not help.

### Expected Behavior

Upgrading from 0.26.1 to 0.27.1 should not cause this CPU/memory regression under the same traffic pattern. The changelog says performance _should_ improve after introducing in-memory state in 0.27.x, but it got worse.

### Steps To Reproduce

1. Run Headscale 0.27.1 with ~80 connected clients.
2. Let normal traffic run.
3. Capture profiles:
    - `go tool pprof "http://localhost:9090/debug/pprof/profile?seconds=30"`
    - `go tool pprof "http://localhost:9090/debug/pprof/allocs?seconds=30"` (profile.MemProfileAllocs needs to be enabled)
5. Inspect:
    - `go tool pprof -http=localhost:8080 <path-to-cpu-profile>`
    - `go tool pprof -http=localhost:8080 <path-to-mem-profile>`

### Environment

```markdown
- OS: Ubuntu 24.04, Kernel 6.8
- Headscale version: 0.27.1
– Go: 1.26.0
- Tailscale version: 1.94.2
- Number of nodes: ~80 connected
```

### Runtime environment

### Debug information

I profiled this and attached CPU + heap allocation flamegraphs.

CPU:
Image

Memory allocation space:
Image

The hotspot is:
```
BuildPeerMap -> NodeView.CanAccess -> NodeView.AsStruct -> Node.Clone / Hostinfo.Clone / NetInfo.Clone / maps.clone
```

This path allocates about **2.31GB in 30 seconds**.

dmeremyanin:
NodeView.CanAccess called node2.AsStruct() on every check. In peer-map construction we run CanAccess in O(n^2) pair scans (often twice per pair), so that per-call clone multiplied into large heap churn.

This PR removes that clone path in `CanAccess` and keeps the same behavior.

I also added a simple benchmark. Local results:
```
❯ go test ./hscontrol/types -run '^$' -bench BenchmarkNodeViewCanAccess -benchmem
goos: darwin
goarch: arm64
pkg: github.com/juanfont/headscale/hscontrol/types
cpu: Apple M4 Pro
BenchmarkNodeViewCanAccess/pointer-14           26586536                45.93 ns/op           48 B/op          2 allocs/op
BenchmarkNodeViewCanAccess/struct_clone-14        513435              2254 ns/op           14944 B/op         10 allocs/op
PASS
ok      github.com/juanfont/headscale/hscontrol/types   2.590s
```

So it's roughly 50x faster in this benchmark, with much lower allocations.

Fixes #3108

---

- [ ] added unit tests
- [x] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/types/node.go
+++ b/hscontrol/types/node.go
@@ -800,11 +800,11 @@ func (nv NodeView) InIPSet(set *netipx.IPSet) bool {
 }
 
 func (nv NodeView) CanAccess(matchers []matcher.Match, node2 NodeView) bool {
-	if !nv.Valid() {
+	if !nv.Valid() || !node2.Valid() {
 		return false
 	}
 
-	return nv.ж.CanAccess(matchers, node2.AsStruct())
+	return nv.ж.CanAccess(matchers, node2.ж)
 }
 
 func (nv NodeView) CanAccessRoute(matchers []matcher.Match, route netip.Prefix) bool {
--- a/hscontrol/types/node_benchmark_test.go
+++ b/hscontrol/types/node_benchmark_test.go
@@ -1,0 +1,73 @@
+package types
+
+import (
+	"fmt"
+	"net/netip"
+	"testing"
+
+	"github.com/juanfont/headscale/hscontrol/policy/matcher"
+	"tailscale.com/tailcfg"
+)
+
+func BenchmarkNodeViewCanAccess(b *testing.B) {
+	addr := func(ip string) *netip.Addr {
+		parsed := netip.MustParseAddr(ip)
+		return &parsed
+	}
+
+	rules := []tailcfg.FilterRule{
+		{
+			SrcIPs: []string{"100.64.0.1/32"},
+			DstPorts: []tailcfg.NetPortRange{
+				{
+					IP:    "100.64.0.2/32",
+					Ports: tailcfg.PortRangeAny,
+				},
+			},
+		},
+	}
+	matchers := matcher.MatchesFromFilterRules(rules)
+
+	derpLatency := make(map[string]float64, 256)
+	for i := range 128 {
+		derpLatency[fmt.Sprintf("%d-v4", i)] = float64(i) / 10
+		derpLatency[fmt.Sprintf("%d-v6", i)] = float64(i) / 10
+	}
+
+	src := Node{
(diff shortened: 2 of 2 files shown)


hscontrol/types/node.go
```diff
@@ -800,11 +800,11 @@ func (nv NodeView) InIPSet(set *netipx.IPSet) bool {
 }
 
 func (nv NodeView) CanAccess(matchers []matcher.Match, node2 NodeView) bool {
-	if !nv.Valid() {
+	if !nv.Valid() || !node2.Valid() {
 		return false
 	}
 
-	return nv.ж.CanAccess(matchers, node2.AsStruct())
+	return nv.ж.CanAccess(matchers, node2.ж)
```
kradalby, line 807: I agree, this looks like an oversight and it would be fine to use the internal here. Good catch.

dmeremyanin: Great, thanks for confirming. Does the PR look good to merge as-is, or would you like any changes first? Also, should I add a changelog entry for this, and do you want to keep the benchmark test or drop it before merge?

kradalby: You can leave them, this is fine.

dmeremyanin:
I'm not sure the CI failure is related to the changes in this PR

kradalby (approved):

dmeremyanin:
@kradalby thanks for merging!

I've been playing around with a few more performance improvements, and in my setup CPU usage is basically close to zero now. I know you're usually pretty careful about performance-related changes in Headscale, but would you be open to reviewing a couple more PRs like this in the future? No new libraries or major core logic changes involved.

Cediddi:
@dmeremyanin; As a user of Headscale, I'd be more than happy if you'd contribute your performance-related PRs. 🙏  Thanks for this PR

PR #2987, opened 2026-01-01
policy: add App Connector support

matanbaruch:
Implement App Connector functionality for Headscale, allowing nodes to advertise as app connectors and receive domain-based routing configuration from the control plane. This addresses issue #1651.

### Core Implementation
- Add `appConnectors` field to Policy struct for defining app connector configurations in ACLs
- Parse app connector configuration including name, connectors (tags or "*"), domains (with wildcard support), and optional routes
- Add validation for app connector configuration (domains, tags, etc.)
- Add `AppConnectorConfigForNode` method to PolicyManager to get matching configurations for nodes advertising as app connectors
- Update mapper to add `tailscale.com/app-connectors` capability to CapMap in MapResponse for nodes advertising as app connectors

### Tests
- Add comprehensive unit tests for app connector functionality (`hscontrol/policy/v2/appconnector_test.go`)
- Add integration tests (`integration/appconnector_test.go`):
  - `TestAppConnectorBasic`: Tests tagged nodes receive app connector configuration
  - `TestAppConnectorNonMatchingTag`: Tests non-matching tags are excluded
  - `TestAppConnectorWildcardConnector`: Tests wildcard connector matching

### Documentation
- Add App Connectors section to `docs/ref/acls.md` with configuration examples
- Add App Connectors to feature list in `docs/about/features.md`
- Add CHANGELOG.md entry

## Example ACL Configuration

```json
{
  "tagOwners": {
    "tag:connector": ["user@example.com"]
  },
  "appConnectors": [
    {
      "name": "Internal Apps",
      "connectors": ["tag:connector"],
      "domains": ["internal.example.com", "*.corp.example.com"],
      "routes": ["10.0.0.0/8"]
    }
  ]
}
```

## How It Works

1. Configure tagged nodes as app connectors in your ACL policy
2. Nodes with the specified tags that advertise themselves as app connectors (`tailscale set --advertise-connector`) will receive the domain configuration
3. When clients query DNS for the configured domains, traffic is automatically routed through the connector nodes

Closes #1651

- [x] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [x] added unit tests
- [x] added integration tests
- [x] updated documentation if needed
- [x] updated CHANGELOG.md

Issue #2782, opened 2025-09-18
[Bug] extra records (JSON) works only when names are in lowercase


reinob:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I have configured my headscale DNS using NextDNS (works OK) as well as an additional extra-records.json file (configured in `extra_records_path`) for resolving names of my home network.

I noticed (using `dig <name> @100.100.100.100`) that regardless of how I type the name when resolving (lower, upper, MiXeD case) headscale only returns a valid response (A/AAAA) when the name is stored in lowercase in the JSON file. If the JSON file contains names like "Printer.fritz.box" it is never matched, so the query goes to the global nameserver.

### Expected Behavior

Regardless of how the name is written in the JSON file, the query should succeed.
(e.g. headscale should do the lookup disregarding case).

### Steps To Reproduce

As above.

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.26.1
- Tailscale version: tested above with Android/1.86.4
```

### Runtime environment

- [ ] Headscale is behind a (reverse) proxy
- [ ] Headscale runs in a container

### Debug information

I don't think anything is necessary here?

reinob:
Maybe if/when I have some time I can learn some rust (currently "read-only") and see where I can change the comparison to be case-insensitive. Whether I'm able to send a proper "commit" I can't say for now — my git is also very rusty :)

Issue #2753, opened 2025-08-30
[Bug] Headscale reads extra DNS records from path too early


nblock:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I use the `dns.extra_records_path` configuration option to load DNS records from a JSON file. This file is created by a script that combines multiple files into one. The relevant line is:

```bash
jq --sort-keys --slurp 'unique | flatten' dns.d/*.json > /var/lib/headscale/extra-records.json
```

There are about 150 DNS records written to `/var/lib/headscale/extra-records.json` with each invocation. From time to time Headscale's filewatcher picks up the file while it is being written. Presumably, the output buffer is flushed by `jq` and a partial JSON file is persisted for a short while. This causes errors about broken JSON in Headscale's logs:

```
[snipped]
{\\n    \\\"name\\\": \\\"acgmw.example.com\\\",\\n    \\\"type\\\": \\\"A\\\",\\n    \\\"value\\\": \\\"100.64.0.1\\\"\\n  },\\n  {\\n    \\\"name\\\": \\\"acgmx.example.com\\\",\\n    \\\"type\\\": \\\"A\\\",\\n    \\\"value\\\": \\\"100.64.0.1\\\"\\n  },\\n  {\\n    \\\"name\\\": \\\"acgmy.example.com\\\",\\n    \\\"type\\\": \\\"A\\\",\\n    \\\"v\": 
unexpected end of JSON input"
```

As a workaround, I use `stdbuf` to enlarge the output buffer for `jq`:

```bash
stdbuf -o 1M jq --sort-keys --slurp 'unique | flatten' dns.d/*.json > /var/lib/headscale/extra-records.json
```

Is there a way to configure the filewatcher such that it picks up the file once no other process is writing to it?

### Expected Behavior

Unsure what is the best solution to this, some ideas:

- Read the records in `dns.extra_records_path` once no other process writes to it
- Delay the read a bit (probably just a workaround)

### Steps To Reproduce

1. Setup Headscale 0.26.1 with `dns.extra_records_path`
1. Create some sample JSON files
    - [sample-1000.json](https://github.com/user-attachments/files/22057864/sample-1000.json)
    - [sample-100.json](https://github.com/user-attachments/files/22057863/sample-100.json)
    - [sample-10.json](https://github.com/user-attachments/files/22057865/sample-10.json)
[...]

PR #3121, opened 2026-03-01
policy: add App Connector support via nodeAttrs

### LINKED ISSUE #1651: FR: support for App Connectors
## Why

Tailscale launches [new App Connectors feature](https://tailscale.com/kb/1281/app-connectors/) to allow some traffic to pass through specific servers. For example, I have access to [example.com](example.com) only through my edge server running Tailscale with the `--advertise-connector` flag.

## Description

It's like exit-node, but for specific domains only, and as far as I understand, configuration must be done on Headscale side.

matanbaruch:
App connector configuration is now delivered through the `nodeAttrs` `app` field, the way Tailscale delivers it, instead of the custom `appConnectors` policy block this PR originally added. Rebased on current main.

Following up on @Matchlighter's suggestion: after #3251 landed nodeAttrs, the only missing piece was valued capabilities. `nodeAttrs` carried key-only caps through `attr` and nothing else.

## What changed

1. `NodeAttrGrant` gets an `app` field (`tailcfg.NodeCapMap`), stamped into the node's CapMap in `compileNodeAttrs` next to the key-only attrs. Payloads are passed through opaquely and the client interprets them.
2. App connectors ride that generic path, so there is no app-connector-specific code left anywhere: `compileNodeAttrs` -> `NodeCapMap` -> `NodesWithChangedCapMap` -> `SelfUpdate`.
3. The `PolicyChange{IncludeSelf: true}` broadcast the first version needed is gone. A reload now sends a targeted self update only to nodes whose config actually changed, which also fixes the reload propagation @dmeremyanin hit.
4. Removed: the `appConnectors` block, the `AppConnector` / `AppConnectorAttr` types, `AppConnectorConfigForNode` and its wrappers, the `builder.go` injection, and the load-time domain validation.

```json
{
  "tagOwners": {
    "tag:connector": ["admin@"]
  },
  "nodeAttrs": [
    {
      "target": ["tag:connector"],
      "app": {
        "tailscale.com/app-connectors": [
          {
            "name": "Internal Apps",
            "connectors": ["tag:connector"],
            "domains": ["internal.example.com", "*.corp.example.com"],
            "routes": ["10.0.0.0/8"]
          }
        ]
      }
    }
  ]
}
```

## What works and what does not

Works today: the capability reaches every node the `target` selects, hardcoded `routes` are advertised as soon as the connector gets its config, and policy reloads propagate.

Does not work yet: Headscale does not tell clients to send DNS queries for the configured domains to the connector, so dynamic domain discovery does not happen on its own. That is the DNS side of the feature and it belongs with the v0.31 work. `docs/ref/policy.md` says so in a warning admonition, `features.md` marks App Connectors unchecked with the delivery part checked underneath, so nobody reads this as done.

Two deliberate behavior notes:

1. The cap ships on `target` match regardless of `tailscale set --advertise-connector`. This matches Tailscale, where the client gates on its local connector flag and on the payload's `connectors` field.
2. No load-time domain validation. Payloads are pass through, so a malformed domain surfaces client side.

## Cap name validation

Grant `app` cap names go through `validateCapabilityName`, which rejects URL-formatted names and `tailscale.com/*` names that are not in `tailscaleCapAllowlist`. The `app` field on nodeAttrs needs the same guard, otherwise a policy can ship any `tailscale.com` capability to a node with nothing behind it:

```json
{ "target": ["*"], "app": { "tailscale.com/cap/funnel": [] } }
```

So there is a `validateNodeCapabilityName` with its own allowlist, currently holding only `tailscale.com/app-connectors`. It is a separate map because `tailscaleCapAllowlist` is keyed by `tailcfg.PeerCapability` and its comment is about peer caps. Names in an operator's own domain, like `example.com/app/my-cap`, still pass through. The reasoning is the one already in `nodeAttrUnsupportedCaps`: rejecting at load beats letting a node advertise something that does not work.

If you would rather have one shared allowlist, or no gate at all on the nodeAttrs side, say so and I will change it.

## Tests

Unit (`hscontrol/policy/v2/appconnector_test.go`), on the shared nodeAttrs fixtures:

1. `TestAppConnectorViaNodeAttrs`: targeted node gets both configs, non-targeted nodes get nothing, `routes` round trips through `appctype.AppConnectorAttr`.
2. `TestAppConnectorWildcardTarget`: `*` reaches every node.
3. `TestAppConnectorChangeTracking`: the cap participates in `NodesWithChangedCapMap`, which is what makes the `IncludeSelf` removal safe.
4. `TestValidateNodeCapabilityName` and `TestAppConnectorUnsupportedCapRejected`: the allowlist, both at the helper and at policy load.

Integration (`integration/appconnector_test.go`): `TestAppConnectorBasic`, `TestAppConnectorNonMatchingTag`, `TestAppConnectorWildcardConnector`. Configs are marshalled through `appctype.AppConnectorAttr` so the tests use the same wire shape a Tailscale hosted control plane emits.

Updates #1651

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -243,6 +243,9 @@ jobs:
           - TestAPIAuthenticationBypassCurl
           - TestRemoteCLIAuthenticationBypass
           - TestCLIWithConfigAuthenticationBypass
+          - TestAppConnectorBasic
+          - TestAppConnectorNonMatchingTag
+          - TestAppConnectorWildcardConnector
           - TestAuthKeyLogoutAndReloginSameUser
           - TestAuthKeyLogoutAndReloginNewUser
           - TestAuthKeyLogoutAndReloginSameUserExpiredKey
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ keys remain all-access.
 
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
+- `nodeAttrs` accepts an `app` field for valued capabilities, delivering [App Connector](./docs/ref/policy.md#app-connectors) configuration via `tailscale.com/app-connectors` [#3121](https://github.com/juanfont/headscale/pull/3121)
 
 ## 0.29.2 (2026-07-01)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ keys remain all-access.
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
 - Headscale now requires Go 1.27 to build
+- `nodeAttrs` accepts an `app` field for valued capabilities, delivering [App Connector](./docs/ref/policy.md#app-connectors) configuration via `tailscale.com/app-connectors` [#3121](https://github.com/juanfont/headscale/pull/3121)
 
 ## 0.29.3 (2026-07-29)
 
--- a/docs/about/features.md
+++ b/docs/about/features.md
@@ -40,6 +40,9 @@ provides on overview of Headscale's feature and compatibility with the Tailscale
     - [x] Basic registration
     - [x] Update user profile from identity provider
     - [ ] OIDC groups cannot be used in ACLs
+- [ ] [App Connectors](https://tailscale.com/kb/1281/app-connectors) ([#1651](https://github.com/juanfont/headscale/issues/1651))
+    - [x] [Connector configuration delivered via node attributes](../ref/policy.md#app-connectors)
+    - [ ] Routing client DNS queries for configured domains to the connector
 - [ ] [Funnel](https://tailscale.com/docs/features/tailscale-funnel) ([#1040](https://github.com/juanfont/headscale/issues/1040))
 - [ ] [Serve](https://tailscale.com/docs/features/tailscale-serve) ([#1234](https://github.com/juanfont/headscale/issues/1921))
 - [ ] [Network flow logs](https://tailscale.com/docs/features/logging/network-flow-logs) ([#1687](https://github.com/juanfont/headscale/issues/1687))
--- a/docs/ref/policy.md
+++ b/docs/ref/policy.md
@@ -226,6 +226,11 @@ configuration and attributes. At least the following node attributes are current
 }
 ```
 
+The `attr` field above carries key-only capabilities. A `nodeAttrs` entry may instead (or additionally) carry an `app`
+field for *valued* capabilities — a capability name mapped to a list of JSON payloads, matching Tailscale's `app` field.
+Headscale passes these payloads through unmodified. [App Connectors](#app-connectors) are delivered this way via the
+`tailscale.com/app-connectors` capability.
+
 ## Network-wide policy options
 
 The following options are applied for the entire tailnet. Consider [node attributes](#node-attributes) for a more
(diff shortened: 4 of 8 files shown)


matanbaruch:
This re-opened by a request from @kradalby https://github.com/juanfont/headscale/pull/2987#issuecomment-3941578426

matanbaruch:
@kradalby The golangci-lint failure was a `wsl_v5` issue — missing blank line before an assignment in `integration/appconnector_test.go:356`. Fixed in e4180cd7. Could you please approve and re-run the CI checks? Thanks!

matanbaruch:
@kradalby looks like it failed on a test that fails on most of the integrations run.
Can this get approved?

kradalby:
The test by itself is fine, but approval is not just granted, we will have to review it when we have time.

This one will likely require more testing aswell to make sure we get it right so we don't have to fix it up later.

matanbaruch:
> The test by itself is fine, but approval is not just granted, we will have to review it when we have time.
> 
> This one will likely require more testing aswell to make sure we get it right so we don't have to fix it up later.

I don't think the failure of this test related to my PR.

Let me know when you have time to review it.

heyanlong:
This is a feature I’ve been looking forward to for a long time. Really glad to see it finally implemented.
﻿
The design looks clean and the implementation feels straightforward. It should make the workflow much smoother for users who rely on this capability.
﻿
Thanks for the great work!

Lvmoo:
That's right—this is a feature I've been looking forward to for a long time. I hope it can be merged and released as soon as possible.

gerilya:
Is it possible to get an ETA for review? I realize you guys are busy and understanding the priorities would help to decide whether it's worth building binaries from a branch for the benefit of anyone who needs it. Thank you

kradalby:
Could happen before, but realistically with the amount of testing I would expect this to need, I doubt we would get to it before the DNS focused release, which is [`v0.31.0`](https://github.com/juanfont/headscale/milestone/16).

gerilya:
> Could happen before, but realistically with the amount of testing I would expect this to need, I doubt we would get to it before the DNS focused release, which is [`v0.31.0`](https://github.com/juanfont/headscale/milestone/16).

Thank you for the prompt response

dmeremyanin:
I tested this PR in my setup. Here are my observations:
- It's probably worth mentioning that `autoApprovers` also need to be configured if you want advertised routes to be approved automatically, since App Connector effectively works as a dynamic subnet router under the hood.
- I couldn't get the full dynamic flow working in my setup. My local Tailscale client doesn't seem to send DNS requests to the app connector node, so no routes are advertised.
- It does start working if I manually send a DoH request to the app connector node for the domain via PeerAPI, for example: `http://100.90.183.244:64526/dns-query?q=ipinfo.io`
- It works as expected if routes are hardcoded in the `appConnectors` section - those routes are advertised immediately. At that point, though, I'm not entirely sure how this differs in practice from regular subnet route advertisement.
- In my testing, adding a new domain to the ACL and running `headscale reload` did not propagate the updated domain list to the app connector node. The app connector node only picked up the updated domain list after restarting the tailscaled service.
- In my tests, `tailscale appc-routes --all` on the app connector node always returned nothing.

It seems like the critical piece here is the DoH flow from clients. I'm happy to help debug/test further, but I still don't understand how Tailscale is expected to route DNS resolution for configured domains through a specific node in the tailnet.

---

This is the ACL config I tested with:

```json
{
    "acls": [
        {
            "action": "accept",
            "dst": [
                "autogroup:internet:*"
            ],
            "src": [
                "group:admin"
            ]
        }
    ],
    "appConnectors": [
        {
            "connectors": [
                "tag:router"
            ],
            "domains": [
                "ipinfo.io"
            ]
        }
    ],
    "autoApprovers": {
        "routes": {
            "0.0.0.0/0": [
                "tag:router"
            ]
        }
    },
    "groups": {
        "group:admin": [
            "admin@",
        ]
    },
    "tagOwners": {
        "tag:router": [
            "router@"
        ]
    }
}
```

`tag:router` is the app connector node, and `admin` is my local machine.

matanbaruch:
@dmeremyanin Thanks for the thorough testing, your findings align with known issues being tracked across milestones.

The autoApprovers/route approval behavior you observed is related to bugs being actively fixed in v0.29.0 (PR #3175 by kradalby and PR #3179 — both fix route approval logic in the same policy code this PR builds on).

The DNS routing gap you identified (clients not sending DNS queries to the connector node) is exactly why kradalby placed this PR in the [v0.31.0 "DNS Work" milestone](https://github.com/juanfont/headscale/milestone/16). There are foundational DNS bugs (#2899, #2782, #2753) that need to be resolved first, and then the DNS-to-connector routing can be built on solid ground.

This PR intentionally covers the policy/capability layer. The DNS integration will follow as part of the v0.31.0 DNS work. I'll fix the policy reload propagation issue in the meantime.

Matchlighter:
With `0.29` adding `nodeAttrs` a bit more thoroughly, would AppConnector support now be "just" a matter of adding support for `nodeAttrs.app` (with the aforementioned DNS caveat)? Seems like that may be a cleaner solution as it can then match how Tailscale configures app connectors and would reduces the burden on Headscale to document the custom `appConnectors` policy entry.

matanbaruch:
@Matchlighter makes sense, gave it a shot - took a stab at moving it onto `nodeAttrs.app` with the `tailscale.com/app-connectors` cap instead of the custom `appConnectors` block, roughly how Tailscale does it. Had to add valued-cap support to nodeAttrs first (it only did key-only `attr` before), and from there app connectors mostly fall out of the existing machinery.

One thing I liked: it rides the normal `NodeCapMap -> NodesWithChangedCapMap -> SelfUpdate` path now, so I think the `IncludeSelf` workaround can go.

Still need to shake it out, but a couple of things I'm not sure about: the cap ends up shipping on `target` match regardless of `--advertise-connector` (figured the client gates locally anyway?), and I dropped the load-time domain validation in favor of pass-through. Open to feedback on both. The DNS routes + auto-approval part I left alone, still on `autoApprovers`.

matanbaruch:
Rebased onto current main and reworked onto `nodeAttrs`. Not asking for a review pass yet, the DNS side still belongs with v0.31.

What changed since the last push:

1. App connector config rides the `app` field on `nodeAttrs` now (@Matchlighter's suggestion) instead of the custom `appConnectors` block. There is no app-connector-specific code left in the delivery path.
2. `PolicyChange{IncludeSelf: true}` is gone. A reload sends a targeted self update only to nodes whose config actually changed, which covers the reload propagation @dmeremyanin hit.
3. Added a gate on nodeAttrs `app` capability names, mirroring what grants already do through `validateCapabilityName`. The allowlist holds `tailscale.com/app-connectors` only. Say the word if you would rather have one shared allowlist, or no gate on the nodeAttrs side.
4. Docs and `features.md` state plainly that the DNS side is missing, so App Connectors is not marked as done anywhere.

All Garnix checks are green. The GitHub Actions workflows have not run on this head, looks like they need approval for a fork PR.

Matchlighter:
I gave the `NodeAttrGrant` here a run and it seems to work as expected! I ended up going another direction on the "App Connectors" front, so haven't tested that aspect specifically.

If you're interested (shameless plug to get a tester or two!), "I" (I designed, Codex did a lot of the code since I'm horrible at Go) created https://github.com/Matchlighter/ts-domain-gateway.

PR #3122, opened 2026-03-02
types,state: add configurable default node key expiry

### LINKED ISSUE #1711: Authenicating with auth-key does not set node expiry
## Bug description

Possibly related to https://github.com/juanfont/headscale/issues/1579

With adding new nodes using authentication keys, the expiry of the node is set to `0` on the authentication request, resulting in `headscale nodes list` outputting entries like this where expiry is `0001-01-01 00:00:00`.

```
45 | machine-1                  | machine-1-qgg83fyw         | [RiUAc]    | [pDq5E] | machine-1                       | 100.64.0.4, fd7a:115c:a1e0::4   | false     | 2024-02-02 19:49:17 | 0001-01-01 00:00:00 | offline   | no
```

I'm unsure if there's logic somewhere else that handles this, but it doesn't quite feel right. I would have thought there to be some default expiry, such as when logging in with a GUI client and OIDC where it sets the expiry to 2 weeks. 

My _"I have no idea what I'm doing"_ patch looks like this to resolve it for my preference. 

https://github.com/dustinblackman/headscale/commit/ad745bee3f98d4f52590ca8374ffbb7cb450e51a

Should there be a config and functionality to set a default expiry when an authentication request does not have one set?

## Environment

- Version of headscale used: `v0.23.0-alpha2`
- Version of tailscale client: `1.58.2`
- OS (e.g. Linux, Mac, Cygwin, WSL, etc.) and version: Debian 11
- Kernel version: Not relevant. 
- The relevant config parameters you used: Not relevant.
- Log output: `headscale nodes list` output shown above.

- [ ] Headscale is behind a (reverse) proxy - No
- [ ] Headscale runs in a container - No

## To Reproduce

1. On your headscale machine, run the following to create a user, and a preauthkey.

```bash
headscale users create machine-1
headscale preauthkeys create -u machine-1 -e 1h --tags tag:myawesomemachine
```

2. On your machine (dubbed `machine-1`), login.

```bash
sudo tailscale up --login-server https://my-headscale-server.xyz --accept-dns=false --auth-key KEY-FROM-STEP-2
```

3. Back on your headscale machine, look at the node expiry.

```bash
headscale nodes list | grep machine-1
```

kradalby:
Add `node.expiry` config (default: 180d) as the unified default key
expiry for all non-tagged nodes regardless of registration method
(auth key, CLI, web auth). Tagged nodes remain exempt and never expire.

OIDC expiry falls back to `node.expiry` when `oidc.expiry` is not
explicitly set, unifying the expiry model. The old
`ephemeral_node_inactivity_timeout` key is preserved for backwards
compatibility under `node.ephemeral.inactivity_timeout`.

Fixes #1711

### DIFF
--- a/hscontrol/capver/capver_generated.go
+++ b/hscontrol/capver/capver_generated.go
@@ -41,6 +41,7 @@ var tailscaleToCapVer = map[string]tailcfg.CapabilityVersion{
 	"v1.90": 130,
 	"v1.92": 131,
 	"v1.94": 131,
+	"v1.96": 133,
 }
 
 var capVerToTailscaleVer = map[tailcfg.CapabilityVersion]string{
@@ -75,6 +76,7 @@ var capVerToTailscaleVer = map[tailcfg.CapabilityVersion]string{
 	125: "v1.88",
 	130: "v1.90",
 	131: "v1.92",
+	133: "v1.96",
 }
 
 // SupportedMajorMinorVersions is the number of major.minor Tailscale versions supported.
@@ -82,4 +84,4 @@ const SupportedMajorMinorVersions = 10
 
 // MinSupportedCapabilityVersion represents the minimum capability version
 // supported by this Headscale instance (latest 10 minor versions)
-const MinSupportedCapabilityVersion tailcfg.CapabilityVersion = 106
+const MinSupportedCapabilityVersion tailcfg.CapabilityVersion = 109
--- a/hscontrol/capver/capver_test_data.go
+++ b/hscontrol/capver/capver_test_data.go
@@ -9,10 +9,9 @@ var tailscaleLatestMajorMinorTests = []struct {
 	stripV   bool
 	expected []string
 }{
-	{3, false, []string{"v1.90", "v1.92", "v1.94"}},
-	{2, true, []string{"1.92", "1.94"}},
+	{3, false, []string{"v1.92", "v1.94", "v1.96"}},
+	{2, true, []string{"1.94", "1.96"}},
 	{10, true, []string{
-		"1.76",
 		"1.78",
 		"1.80",
 		"1.82",
@@ -22,6 +21,7 @@ var tailscaleLatestMajorMinorTests = []struct {
 		"1.90",
 		"1.92",
 		"1.94",
+		"1.96",
 	}},
 	{0, false, nil},
 }
@@ -30,7 +30,7 @@ var capVerMinimumTailscaleVersionTests = []struct {
 	input    tailcfg.CapabilityVersion
 	expected string
 }{
-	{106, "v1.74"},
+	{109, "v1.78"},
 	{32, "v1.24"},
 	{41, "v1.30"},
 	{46, "v1.32"},


hscontrol/state/state.go
```diff
@@ -1431,6 +1431,16 @@ func (s *State) applyAuthNodeUpdate(params authNodeUpdateParams) (types.NodeView
 			}
 		}
 		// Tagged → Tagged: keep existing expiry (nil) - no action needed
+
+		// Apply default node expiry for non-tagged nodes when the
+		// resolved expiry is still nil or zero (e.g., CLI registration
+		// where the client did not request a specific expiry).
+		if !node.IsTagged() &&
+			(node.Expiry == nil || node.Expiry.IsZero()) &&
+			s.cfg.Node.Expiry > 0 {
```
kradalby: Lets make this an easier to read variable condition.

hscontrol/state/state.go
```diff
@@ -1553,6 +1563,17 @@ func (s *State) createAndSaveNewNode(params newNodeParams) (types.NodeView, erro
 		}
 	}
 
+	// Apply default node expiry for non-tagged nodes when the client
+	// did not request a specific expiry. This aligns auth key and CLI
+	// registrations with OIDC behaviour (180 day default).
+	// Tagged nodes are exempt — they never expire.
+	if !nodeToRegister.IsTagged() &&
+		(nodeToRegister.Expiry == nil || nodeToRegister.Expiry.IsZero()) &&
+		s.cfg.Node.Expiry > 0 {
```
kradalby: Lets make this an easier to read variable condition.

hscontrol/types/config.go
```diff
@@ -395,6 +416,12 @@ func LoadConfig(path string, isFile bool) error {
 	viper.SetDefault("randomize_client_port", false)
 	viper.SetDefault("taildrop.enabled", true)
 
+	viper.SetDefault("node.expiry", "180d")
+	viper.SetDefault("node.ephemeral.inactivity_timeout", "120s")
+
+	// Backwards compatibility: old top-level key still works.
+	// If the old key is explicitly set and the new key is not,
+	// the old value takes precedence.
 	viper.SetDefault("ephemeral_node_inactivity_timeout", "120s")
```
kradalby: Mark it deprecated so people get a warning that it will be removed.

hscontrol/types/config.go
```diff
@@ -418,6 +445,78 @@ func LoadConfig(path string, isFile bool) error {
...
+	// New key takes precedence if explicitly set in config.
+	if viper.IsSet("node.ephemeral.inactivity_timeout") &&
+		viper.GetString("node.ephemeral.inactivity_timeout") != "" {
+		return viper.GetDuration("node.ephemeral.inactivity_timeout")
+	}
+
+	// Fall back to old key for backwards compatibility.
+	if viper.IsSet("ephemeral_node_inactivity_timeout") {
+		return viper.GetDuration("ephemeral_node_inactivity_timeout")
+	}
```
kradalby, line 456: Doesnt viper support mapping the value onto the new value?

config-example.yaml
```diff
@@ -145,8 +145,23 @@ derp:
...
+  # (auth key, CLI, web auth). Tagged nodes are exempt and never expire.
+  #
+  # This is the base default. OIDC can override this via oidc.expiry.
+  # If a client explicitly requests a specific expiry, the client value is used.
+  #
+  # Setting the value to "0" means no default expiry (nodes never expire unless
+  # explicitly expired via `headscale nodes expire`).
+  #
+  # Default: 180d (matches Tailscale SaaS behaviour)
+  expiry: 180d
```
kradalby: actually, we should document default saas behaviour, but not use it. No expiry should be default as it has been a painpoint for saas.

hscontrol/types/config.go
```diff
@@ -385,7 +404,8 @@ func LoadConfig(path string, isFile bool) error {
 
 	viper.SetDefault("oidc.scope", []string{oidc.ScopeOpenID, "profile", "email"})
 	viper.SetDefault("oidc.only_start_if_oidc_is_available", true)
-	viper.SetDefault("oidc.expiry", "180d")
+	// Note: oidc.expiry has NO viper default. If not explicitly set in config,
+	// it falls back to node.expiry via resolveOIDCExpiry().
```
kradalby: drop this comment

hscontrol/types/config.go
kradalby: This is not addressed.

Issue #3125, opened 2026-03-05
Tracking: Flaky integration and unit tests


kradalby:
Tracking issue for flaky tests identified by systematically analyzing CI results across ~75 merged PRs.

Supersedes #2299 and #2935.

## Flaky Tests

Failure rate is approximate, based on ~40 PRs that ran integration tests.

### Critical (~20-42% failure rate)

- [ ] `TestAutoApproveMultiNetwork` (all subtests) — `integration/route_test.go:2037` — ~42% failure rate. ACL/route auto-approve changes don't propagate within EventuallyWithT timeouts on CI runners.
- [ ] `TestACLTagPropagation` — `integration/acl_test.go:2296` — ~20% failure rate. Tag change propagation too slow for CI. Subtests `multiple-tags-partial-removal` and `remove-tag-revokes-access` are worst offenders.

### High (~7-12% failure rate)

- [ ] `TestACLTagPropagationPortSpecific` — `integration/acl_test.go:2852` — ~12%. Port-specific ACL changes via tags don't propagate in time.
- [ ] `TestACLHostsInNetMapTable` — `integration/acl_test.go:101` — ~7%. ACL propagation timing.
- [ ] `TestPingAllByIPManyUpDown` — `integration/general_test.go:1391` — ~7%. Ping failures during up/down cycling.

### Medium (~5% failure rate)

- [ ] `TestAuthWebFlowAuthenticationPingAll` — `integration/auth_web_flow_test.go:18`
- [ ] `TestOIDCExpiryAfterRestart` — `integration/auth_oidc_test.go:1352`
- [ ] `TestEphemeral2006DeletedTooQuickly` — `integration/general_test.go:213`

### Low (~2% failure rate)

- [ ] `TestNodeOnlineStatus` — `integration/general_test.go:1266`
- [ ] `TestAuthWebFlowLogoutAndReloginSameUser` — `integration/auth_web_flow_test.go:60`
- [ ] `TestTagsUserLoginReauthWithEmptyTagsRemovesAllTags` — `integration/tags_test.go:2827`
- [ ] `TestPingAllByIP` — `integration/general_test.go:28`
- [ ] `TestPingAllByIPPublicDERP` — `integration/general_test.go:90`
- [ ] `TestSSHAutogroupSelf` — `integration/ssh_test.go:546`
- [ ] `TestEphemeralInAlternateTimezone` — `integration/general_test.go:131`
[...]

PR #3126, opened 2026-03-05
integration: reduce flaky test failures on CI

kradalby:
Reduce flakiness in integration tests that fail on CI due to resource-constrained runners.

- Scale EventuallyWithT timeouts using PeerSyncTimeout (60s local, 120s CI)
- Use CurlFailFast for assertions expecting connection failure
- Use relaxed ping settings (2s timeout, DERP acceptable) instead of defaults (200ms, direct-only)
- Increase polling intervals for expensive operations (curl, traceroute)

Updates #3125

### DIFF
--- a/integration/general_test.go
+++ b/integration/general_test.go
@@ -1473,16 +1473,14 @@ func TestPingAllByIPManyUpDown(t *testing.T) {
 		}
 		t.Logf("All nodes brought up at %s", time.Now().Format(TimestampFormat))
 
-		// After bringing up all nodes, verify batcher shows all reconnected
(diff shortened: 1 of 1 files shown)


kradalby:
Closing after https://github.com/juanfont/headscale/pull/3130, https://github.com/juanfont/headscale/pull/3143, https://github.com/juanfont/headscale/pull/3144, https://github.com/juanfont/headscale/pull/3134, https://github.com/juanfont/headscale/pull/3145, I think we are in a better place now, with less invasive changes and more internal improvements in headscale. 

We should now work on various other things and collect new statistics for flaky tests from this point to figure out what we need to work on.

PR #3130, opened 2026-03-07
Fix stale netmap sessions after broken client writes

### LINKED ISSUE #3129: [Bug] Newly connected nodes cannot use subnet routers after a netmap stream flap

### Current Behavior

After a transient connection issue affecting a subnet router, Headscale logs warnings like
```json
{"level":"warn","error":"connection 10b0c6925e5a6b8f: timeout sending to channel (likely stale connection)","node.id":122,"chan":"0x182ad4e013b0","conn.id":"10b0c6925e5a6b8f","connection_index":0,"time":1772763603,"message":"send: connection send failed"}
```

Node 122 is the subnet router.

From that point on, nodes that connect **after the event** cannot use that subnet router. Existing nodes that already knew about the router continue to work normally. The problem persists until the subnet router reconnects or Headscale is restarted.

In addition, restarting Headscale does not exit cleanly in this state: **the process hangs during shutdown** and only terminates if killed with SIGKILL.

### Expected Behavior

A temporary network flap or broken netmap stream should not leave a subnet router in a permanently degraded state. Newly connected nodes should be able to use the subnet router without requiring a Headscale restart or a router reconnect.

Headscale should also shut down cleanly instead of hanging during restart.

### Steps To Reproduce

I don't have a fully deterministic reproducer, but the issue appears with this pattern:
1. Run Headscale with at least one subnet router advertising approved routes.
2. Ensure existing clients can reach the routed subnet through that router.
3. Trigger a transient connection problem affecting the subnet router's netmap stream.
4. Observe Headscale log warnings like: `timeout sending to channel (likely stale connection)`.
5. Connect a new node to the tailnet after that event.
6. Observe that the new node cannot use the subnet router, while older nodes still can.
7. Reconnect the subnet router or restart Headscale.
8. Observe that newly connected nodes start working through the subnet router again.
9. In the restart case, observe that Headscale may hang during shutdown and require SIGKILL.

### Environment

```markdown
- OS: Ubunut 24.04
- Headscale version: 0.28.0
- Tailscale version: 1.94.2
```

### Runtime environment

### Debug information

Subnet router logs:
```
magicsock: disco: ignoring CallMeMaybe from d:39944a37b731e697; [GL3bW] is unknown
magicsock: disco: ignoring CallMeMaybe from d:8aed5a55372b8ddd; [+tCFc] is unknown
```

--

Based on local debugging, the issue appears to be caused by two pieces of state getting out of sync:
- Headscale still considers the subnet router online and serving routes.
- The live netmap delivery path from Headscale to that router is no longer usable.

Observed sequence:

1. The subnet router's long-poll / netmap session flaps and stops draining updates from its session channel.
2. Headscale attempts to send an update to that router, the enqueue times out in `connectionEntry.send`, and the failed channel is removed in `multiChannelNodeConn.send`.
3. The router entry remains in the batcher for rapid reconnection, and the router remains online in state until the old `mapSession` fully unwinds through `serveLongPoll`.
4. If that old session is stuck in `writeMap()` / `Flush()` while writing to the client, the deferred cleanup in `serveLongPoll` does not run promptly, so the router is not disconnected in state.
5. New nodes connect and their peer updates are queued and broadcast as usual.
6. Those updates are still targeted at the subnet router, but the router no longer has a usable live map channel, so it never receives the new peer information.
7. Existing nodes continue to work because the router already knew about them before the flap.
8. Nodes that connect after the event cannot use the subnet router because the router never learns about those new peers.

This also appears related to the shutdown hang: when the old `mapSession` gets stuck in `writeMap()` / `Flush()`, Headscale can block during shutdown waiting for that client stream to exit.

dmeremyanin:
This fixes the case where newly connected nodes can no longer use a subnet router after a flap, and also prevents shutdown from hanging on a stuck netmap stream.

Changes:
- handle `RemoveNode` correctly when the channel was already removed earlier by stale-connection cleanup
- tear down stale map sessions when stale-send cleanup removes a timed-out connection
- stop stale sessions through an explicit session teardown hook instead of closing the map-response channel from the batcher side

Fixes #3129

- [x] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/poll_test.go
+++ b/hscontrol/poll_test.go
@@ -59,7 +59,10 @@ func (w *delayedSuccessResponseWriter) Write(data []byte) (int, error) {
 			close(w.firstWriteStarted)
 		})
 
-		time.Sleep(w.firstWriteDelay)
+		timer := time.NewTimer(w.firstWriteDelay)
+		defer timer.Stop()
+
+		<-timer.C
 
 		w.firstWriteFinishedOnce.Do(func() {
 			close(w.firstWriteFinished)
@@ -117,6 +120,7 @@ func TestGitHubIssue3129_TransientlyBlockedWriteDoesNotLeaveLiveStaleSession(t *
 
 	reloadedState, err := state.NewState(app.cfg)
 	require.NoError(t, err)
+
 	app.state = reloadedState
 
 	app.mapBatcher = mapper.NewBatcherAndMapper(app.cfg, app.state)
@@ -140,6 +144,7 @@ func TestGitHubIssue3129_TransientlyBlockedWriteDoesNotLeaveLiveStaleSession(t *
 	}, writer, node)
 
 	serveDone := make(chan struct{})
+
 	go func() {
 		session.serveLongPoll()
 		close(serveDone)
@@ -148,11 +153,14 @@ func TestGitHubIssue3129_TransientlyBlockedWriteDoesNotLeaveLiveStaleSession(t *
 	t.Cleanup(func() {
 		dummyCh := make(chan *tailcfg.MapResponse, 1)
 		_ = app.mapBatcher.AddNode(node.ID, dummyCh, tailcfg.CapabilityVersion(100), nil)
+
 		cancel()
+
 		select {
 		case <-serveDone:
 		case <-time.After(2 * time.Second):
 		}
+
 		_ = app.mapBatcher.RemoveNode(node.ID, dummyCh)
 	})
 
@@ -163,6 +171,7 @@ func TestGitHubIssue3129_TransientlyBlockedWriteDoesNotLeaveLiveStaleSession(t *
 	}
 
 	streamsClosed := make(chan struct{})
+
 	go func() {
 		app.clientStreamsOpen.Wait()
 		close(streamsClosed)


hscontrol/mapper/batcher_lockfree.go
```diff
@@ -135,7 +135,6 @@ func (b *LockFreeBatcher) RemoveNode(id types.NodeID, c chan<- *tailcfg.MapRespo
 	removed := nodeConn.removeConnectionByChannel(c)
 	if !removed {
 		nlog.Debug().Caller().Msg("removeNode: channel not found, connection already removed or invalid")
-		return false
```
dmeremyanin, line 138: Removing the return here ensures RemoveNode still evaluates the remaining connection state even when the channel was already removed earlier. Without that the last stale connection can disappear from multiChannelNodeConn, while the node stays logically connected in the batcher because RemoveNode exits before marking it disconnected.

dmeremyanin:
I tested the patched version and the root issue still appears to be unfixed. After a network flap affecting a subnet router:
- Headscale process still hang during shutdown
- subnet router still does not receive peer updates for nodes that connect after the flap

So the current patch doesnt seem to fully resolve the stale-session path. I'm continuing to look into it

kradalby:
Thank you!

It would be really great if we somehow could trap this condition in a test.

dmeremyanin:
I added a focused reproducer for this in `hscontrol/poll_test.go` (Codex helped me here), and removed the earlier writeMap timeout commit for now, since it seems unrelated to the root issue here, though it may still be worth adding separately later.

The updated fix takes a different approach:
- when stale-send cleanup removes a timed-out connection from multiChannelNodeConn, **it now also closes that session's map-response channel**
- this makes the old serveLongPoll exit instead of lingering as an orphaned stale session
- the client can then reconnect normally and receive a fresh initial full map, which should let a subnet router learn about peers it missed while the old session was stale
- RemoveNode is also kept tolerant of the channel already having been removed earlier, since that is now an expected cleanup order

I'm still learning the Headscale codebase, so I may be missing some nuance here, but this seems closer to the actual failure mode than the previous approach. I've deployed the updated version in my installation and will monitor it.

dmeremyanin:
The [last change](https://github.com/juanfont/headscale/pull/3130/changes/0c07f2ffc9234f59f7085d1b284348b5c7d05fc1) was needed because TestBatcherConcurrentClients/LockFree started failing with `send on closed channel`. What was happening:
- AddNode registers the connection and sends initialMap
- at the same time, another path can decide that session is stale and tear it down
- if the batcher closes the map-response channel from its side, that can race with the initialMap send and panic

I'm not completely happy about changing the AddNode interface for this, but I think it's the cleaner fix.

I first tried to fix it by holding a mutex while sending initialMap, but that didn't feel right. That send can block for a while, and keeping the per-node lock held during it means reconnects, cleanup, stale-session removal, and shutdown for that node can all get stuck behind one slow send.

So instead of trying to protect c <- initialMap with a long-held lock, the latest change makes session teardown explicit: the batcher stops the old session through its cancel path, and no longer closes the data channel underneath senders. That fixes the closed-channel panic in the churn test and feels like a better ownership model too (but makes the AddNode interface a bit wider).

hscontrol/poll.go
```diff
@@ -92,6 +89,12 @@ func (m *mapSession) resetKeepAlive() {
 	m.keepAliveTicker.Reset(m.keepAlive)
 }
 
+func (m *mapSession) stopFromBatcher() {
+	if m.cancelChClosed.CompareAndSwap(false, true) {
```
dmeremyanin, line 93: stopFromBatcher only needs to close cancelCh once, so a CAS is enough here.

dmeremyanin:
Tailscale HEAD now appears to require Go 1.26.1, while this branch and main are still on 1.26.0. Should I bump Go to 1.26.1 in this PR?

kradalby:
> Should I bump Go to 1.26.1 in this PR?

You can bump it to unblock yourself, then we could rebase it away if we get it in before you.

dmeremyanin:
Ok, I bumped Go to 1.26.1 here. I also verified that the previously failing Tailscale docker build now passes locally.

hscontrol/poll.go
kradalby: Neat!

kradalby:
I've rebased your branch with the 1.26 commit in main, it is a bit more involved to get it out, lets see how the tests do now.

dmeremyanin:
Thanks. I removed the time.Sleep and updated the formatting to satisfy the lint rules, so the branch should be clean now

PR #3131, opened 2026-03-08
Add prominent warning banner for non-standard IP prefixes

### LINKED ISSUE #3055: Add more explicit warning against using a non standard IP prefix

TanayK07:
Closes #3055
Closes #3062

## Summary

- Adds a reusable `warnBanner()` function that prints a large ASCII-art "Warning!" banner to the log output, matching the format from #3055
- Modifies `prefixV4()` and `prefixV6()` to return a boolean indicating whether the configured prefix is outside the standard range, instead of logging independently
- Consolidates the check in `LoadServerConfig()` so the banner fires **once** even if both prefixes are non-standard
- Updates `config-example.yaml` to clarify that subsets of the default ranges are fine, but ranges outside CGNAT/ULA are not

## What the warning looks like

When a user configures a prefix outside the standard CGNAT or ULA range, they'll see something like this at startup:

```
################################################################
###      __          __              _             _         ###
###      \ \        / /             (_)           | |        ###
###       \ \  /\  / /_ _ _ __ _ __  _ _ __   __ _| |        ###
###        \ \/  \/ / _` | '__| '_ \| | '_ \ / _` | |        ###
###         \  /\  / (_| | |  | | | | | | | | (_| |_|        ###
###          \/  \/ \__,_|_|  |_| |_|_|_| |_|\__, (_)        ###
###                                           __/ |          ###
###                                          |___/           ###
################################################################
###                                                          ###
###  You have overridden the default Headscale IP prefixes   ###
###  with a range outside of the standard CGNAT and/or ULA   ###
###  ranges. This is NOT a supported configuration.          ###
###                                                          ###
###  Using subsets of the default ranges (100.64.0.0/10 for  ###
###  IPv4, fd7a:115c:a1e0::/48 for IPv6) is fine. Using      ###
###  ranges outside of these will cause undefined behaviour   ###
###  as the Tailscale client is NOT designed to operate on    ###
###  any other ranges.                                        ###
###                                                          ###
###  Please revert your prefixes to subsets of the standard  ###
###  ranges as described in the example configuration.        ###
###                                                          ###
###  Any issue raised using a range outside of the supported ###
###  range will be labelled as wontfix and closed.            ###
###                                                          ###
################################################################
```

## Design decisions

- **`warnBanner()` is reusable** — takes `[]string` lines and formats them into the banner box, so it can be used for other critical warnings in the future
- **Warning fires once** — even if both v4 and v6 are non-standard, only one banner is printed (per review feedback on #3062)
- **Subsets are fine** — the check uses `ContainsPrefix` which correctly allows subsets like `100.64.0.0/24` within the CGNAT range

## Test plan

- [x] Standard prefixes (100.64.0.0/10, fd7a:115c:a1e0::/48) → no warning
- [x] Subset of standard range (e.g., 100.64.0.0/24) → no warning
- [x] Non-standard prefix (e.g., 10.0.0.0/8) → warning banner printed
- [x] Both prefixes non-standard → single warning banner
- [x] Existing tests continue to pass

### DIFF
--- a/hscontrol/types/config.go
+++ b/hscontrol/types/config.go
@@ -862,7 +862,7 @@ func warnBanner(lines []string) {
 	b.WriteString("###                                                          ###\n")
 
 	for _, line := range lines {
-		b.WriteString(fmt.Sprintf("###  %-56s  ###\n", line))
+		b.WriteString(fmt.Sprintf("###  %-54s  ###\n", line))
 	}
 
 	b.WriteString("###                                                          ###\n")
@@ -968,8 +968,9 @@ func LoadServerConfig() (*Config, error) {
 			"Please revert your prefixes to subsets of the standard",
 			"ranges as described in the example configuration.",
 			"",
-			"Any issue raised using a range outside of the supported",
-			"range will be labelled as wontfix and closed.",
+			"Any issue raised using a range outside of the",
+			"supported range will be labelled as wontfix",
+			"and closed.",
 		})
 	}
 


kradalby:
Looks reasonable, but the width seem a bit off in the example, can you smooth out the right side?

TanayK07:
Fixed the banner width — the format string was producing 66-char lines instead of 64. Changed `%-56s` to `%-54s` and rewrapped one overflowing line. Right side should now be smoothly aligned. Thanks for the review!

TanayK07:
@kradalby Here's what the banner looks like now — right side is smoothly aligned at 64 chars:

```
################################################################
###      __          __              _             _         ###
###      \ \        / /             (_)           | |        ###
###       \ \  /\  / /_ _ _ __ _ __  _ _ __   __ _| |        ###
###        \ \/  \/ / _` | '__| '_ \| | '_ \ / _` | |        ###
###         \  /\  / (_| | |  | | | | | | | | (_| |_|        ###
###          \/  \/ \__,_|_|  |_| |_|_|_| |_|\__, (_)        ###
###                                           __/ |          ###
###                                          |___/           ###
################################################################
###                                                          ###
###  You have overridden the default Headscale IP prefixes   ###
###  with a range outside of the standard CGNAT and/or ULA   ###
###  ranges. This is NOT a supported configuration.          ###
###                                                          ###
###  Using subsets of the default ranges (100.64.0.0/10 for  ###
###  IPv4, fd7a:115c:a1e0::/48 for IPv6) is fine. Using      ###
###  ranges outside of these will cause undefined behaviour  ###
###  as the Tailscale client is NOT designed to operate on   ###
###  any other ranges.                                       ###
###                                                          ###
###  Please revert your prefixes to subsets of the standard  ###
###  ranges as described in the example configuration.       ###
###                                                          ###
###  Any issue raised using a range outside of the           ###
###  supported range will be labelled as wontfix             ###
###  and closed.                                             ###
###                                                          ###
################################################################
```

kradalby:
@TanayK07 can you have a look at the linter issues?

TanayK07:
@kradalby The CI failures were caused by the branch being 31 commits behind `main`. The lint workflow uses `--new-from-rev` with the PR base SHA, but with `fetch-depth: 2` in the shallow clone, that old base SHA wasn't available — so golangci-lint fell back to scanning the entire codebase and flagged pre-existing issues in unrelated files (`filter.go`, `debug.go`, `metrics.go`, etc.).

I've rebased onto latest `main` to fix this. The test failures (`TestACLTagPropagation`, `TestPingAllByIP`) are also unrelated to this PR — they appear to be flaky integration tests.

CI should be green now (for the lint job at least). Let me know if you'd like any changes to the code itself!

TanayK07:
@kradalby Hey! CI is looking way better after the rebase — golangci-lint and everything else passes now.

The only remaining failure is `sqlite (TestEphemeral2006DeletedTooQuickly)` — it timed out on a ping (dockertest timeout), which looks like a flaky integration test and isn't related to the changes here. Happy to re-trigger if needed.

Let me know if you'd like any code changes!

PR #3132, opened 2026-03-08
fix: persist empty ApprovedRoutes and Endpoints to database

### LINKED ISSUE #3110: [Bug] Route rejection (empty `approved_routes`) is not persisted to the database

### Description
When a route is approved via the `ApproveRoutes` API (gRPC `SetApprovedRoutes`), the data is saved correctly if there is at least one route. However, when all routes are rejected (resulting in an empty `[]netip.Prefix` slice), the changes are applied to the in-memory `NodeStore` but are **not** persisted to the underlying database. 

Restarting `headscale` will simply reload the old routes making the rejection revert.

### Root cause
The bug originates in the `persistNodeToDB` method (`hscontrol/state/state.go`), which relies on GORM's `.Updates(nodePtr)` to persist node changes to the DB:

```go
// hscontrol/state/state.go
err := s.db.DB.Omit("expiry", "AuthKeyID", "AuthKey").Updates(nodePtr).Error
```

According to GORM's documentation, when an entire struct is passed to `Updates`, GORM completely ignores "zero-value" fields (such as `0`, `""`, `false`, and empty slices `[]`). Thus, an empty slice for `ApprovedRoutes` is quietly omitted from the resulting SQL `UPDATE` statement, causing the database to retain its previously approved routes.

### Current Behavior
The API call succeeds, the in-memory `NodeStore` is updated, and the response looks correct, but it fails to clear the database column `approved_routes`.

### Expected Behavior

Rejecting all routes (setting `ApprovedRoutes` to an empty array) should accurately clear the `approved_routes` column for the node in the database.

### Steps To Reproduce

1. Start `headscale` server.
2. Run the command to approve the route:
   `headscale nodes approve-routes -i 5 -r "10.0.0.0/8"`.
   Then run `headscale nodes list-routes` and you can see that the `10.0.0.0/8` range appears in the route table.
3. Remove the route using:
   `headscale nodes approve-routes -i 5 -r ""`.
   After that, run `headscale nodes list-routes` and the `10.0.0.0/8` range is no longer shown in the route table. However, when checking the database, it has not been deleted.
4. Restart Headscale, and the `10.0.0.0/8` route appears again in the Headscale route table.

### Environment

```markdown
- OS: Ubuntu 24.04
- Headscale version: 0.28.0
- Tailscale version: 1.94.1
```

### Runtime environment

### Debug information

None

TanayK07:
## Summary

Fixes #3110

When an admin rejects all routes for a node (`headscale nodes approve-routes -i 5 -r ""`), the empty `ApprovedRoutes` slice is correctly set in the in-memory NodeStore but never written to the database. On restart, the old routes reappear.

**Root cause**: `persistNodeToDB` uses GORM's `.Updates(struct)`, which silently skips zero-value fields. An empty `[]netip.Prefix` is a zero value, so it's dropped.

**Fix**: After the `.Updates()` call in `persistNodeToDB`, explicitly persist empty `ApprovedRoutes` and `Endpoints` slices by writing `"[]"` — the same pattern already used in `db.SetApprovedRoutes` (node.go:242-244).

This is done in `persistNodeToDB` rather than individual callers so it fixes the issue for all code paths, and also addresses a latent bug where `Endpoints` could similarly fail to persist when empty.

## Test plan

- [x] `go build ./...` compiles
- [x] `go test ./hscontrol/state/` passes (including ephemeral tests)
- [ ] Manual: set approved_routes to empty, restart headscale, confirm routes stay empty

### DIFF
--- a/hscontrol/state/state.go
+++ b/hscontrol/state/state.go
@@ -456,24 +456,13 @@ func (s *State) persistNodeToDB(node types.NodeView) (types.NodeView, change.Cha
 	//   auth_key_id to NULL via ON DELETE SET NULL. Without this, Updates() would fail with a
 	//   foreign key constraint error when trying to reference a deleted PreAuthKey.
 	// See also: https://github.com/juanfont/headscale/issues/2862
-	err := s.db.DB.Omit("expiry", "AuthKeyID", "AuthKey").Updates(nodePtr).Error
+	// Use Select("*") to force GORM to persist zero-value fields (e.g. empty slices
+	// like ApprovedRoutes set to [] when rejecting all routes). Without this, GORM's
+	// Updates(struct) silently skips zero-value fields and they are never persisted.
+	// See: https://github.com/juanfont/headscale/issues/3110
+	err := s.db.DB.Omit("expiry", "AuthKeyID", "AuthKey").Select("*").Updates(nodePtr).Error
 	if err != nil {
 		return types.NodeView{}, change.Change{}, fmt.Errorf("saving node: %w", err)
-	}
-
-	// GORM's Updates(struct) skips zero-value fields, which means empty slices
-	// (like ApprovedRoutes set to [] when rejecting all routes) are silently
-	// dropped. Explicitly persist empty slices to ensure they reach the database.
-	// See: https://github.com/juanfont/headscale/issues/3110
-	if len(nodePtr.ApprovedRoutes) == 0 {
-		if err := s.db.DB.Model(&types.Node{}).Where("id = ?", nodePtr.ID).Update("approved_routes", "[]").Error; err != nil {
-			return types.NodeView{}, change.Change{}, fmt.Errorf("persisting empty approved_routes: %w", err)
-		}
-	}
-	if len(nodePtr.Endpoints) == 0 {
-		if err := s.db.DB.Model(&types.Node{}).Where("id = ?", nodePtr.ID).Update("endpoints", "[]").Error; err != nil {
-			return types.NodeView{}, change.Change{}, fmt.Errorf("persisting empty endpoints: %w", err)
-		}
 	}
 
 	// Check if policy manager needs updating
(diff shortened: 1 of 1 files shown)


hscontrol/state/state.go
```diff
@@ -455,6 +455,21 @@ func (s *State) persistNodeToDB(node types.NodeView) (types.NodeView, change.Cha
...
+	if len(nodePtr.ApprovedRoutes) == 0 {
+		if err := s.db.DB.Model(&types.Node{}).Where("id = ?", nodePtr.ID).Update("approved_routes", "[]").Error; err != nil {
+			return types.NodeView{}, change.Change{}, fmt.Errorf("persisting empty approved_routes: %w", err)
+		}
+	}
+	if len(nodePtr.Endpoints) == 0 {
+		if err := s.db.DB.Model(&types.Node{}).Where("id = ?", nodePtr.ID).Update("endpoints", "[]").Error; err != nil {
+			return types.NodeView{}, change.Change{}, fmt.Errorf("persisting empty endpoints: %w", err)
+		}
+	}
```
kradalby: This happens outside of the transaction, so it is not safe. 

The amount of code to work this out and how this looks makes me think that we are doing something wrong and that GORM cannot be this stupid. There must be a better way to do it, but I have not had time to look into it.

TanayK07:
Hi @kradalby — I've rebased onto latest main (no conflicts). The reworked fix uses `Select("*")` chained on the existing `.Updates()` call to force GORM to include zero-value fields (like empty slices). This is the [documented GORM approach](https://gorm.io/docs/update.html#Update-Selected-Fields) for this — `Omit()` still takes precedence for the protected fields (`expiry`, `AuthKeyID`, `AuthKey`).

The change is a single line in `persistNodeToDB`:
```go
err := s.db.DB.Omit("expiry", "AuthKeyID", "AuthKey").Select("*").Updates(nodePtr).Error
```

CI workflows are waiting on approval to runm. Could you approve them when you get a chance if that's okay?

nblock:
> Would you be able to take another look when you get a chance? Also, could you approve the CI workflows so the checks can run? Thanks!

Please stop to annoy/ping maintainers. Thank you.

TanayK07:
Apologies

nblock:
There's no need to delete/close the PR. Feel free to re-open and just wait until maintainers have a chance to look at it.

PR #3147, opened 2026-03-15
Fix list-routes -i flag not filtering by node ID under approve-routes

### LINKED ISSUE #2927: [Bug] headscale nodes approve-routes list-routes -i flag doesn't filter by node ID

### Current Behavior

After updating to the latest version of Headscale, there appears to be a regression in the CLI for route management commands that breaks existing automation and scripts.

When using the `-i <node_id>` flag with route listing commands, only the routes for the specified node should be returned, similar to how the previous `headscale routes list -i <node_id>` command worked.

The previous CLI command `headscale routes list -i <node_id> -o json` correctly filtered output to only show routes for the specified node ID.

### Expected Behavior

The new command `headscale nodes approve-routes list-routes -i <node_id> -o json` ignores the -i flag entirely and returns route information for all nodes in the network, not just the specified node.

### Steps To Reproduce

Have a Headscale instance with multiple nodes (subnet routers)
Run: `headscale nodes approve-routes list-routes -i 12 -o json`
The output contains route information for ALL nodes, not just node ID 12

### Environment

```markdown
- OS: MacOS Sequoia 15.7.2
- Headscale version: v0.27.0-debug
- Tailscale version: 1.80.2
- Deployment method: Podman container
```

### Runtime environment

### Debug information

**Impact on Automation**
This issue specifically affects:

Ansible role for subnet router configuration
Scripts that need to verify specific node route configurations
Monitoring systems that check route status per node

The workaround of fetching all nodes and filtering client-side is not feasible for production environments with many nodes due to:

Performance overhead
Security concerns (exposing all node data)
Breaking existing automation that expects node-specific output

Note: This is a CLI-specific issue and does not appear to affect the actual routing functionality - routes work correctly once configured. The issue is purely with the management interface returning unfiltered results.

olegnazarov23:
## Summary

Fixes the `-i` flag being ignored when running `headscale nodes approve-routes list-routes -i <id>`, which returned routes for all nodes instead of the specified one.

## Problem

`list-routes` was registered only as a direct subcommand of `nodes`, not `approve-routes`. Running `headscale nodes approve-routes list-routes -i 12` caused cobra to parse `-i` as belonging to `approve-routes` rather than `list-routes`, so `list-routes` saw `identifier=0` and returned all nodes.

## Fix

- Changed `approve-routes` `-i` flag from `Flags()` to `PersistentFlags()` so subcommands inherit it
- Registered `list-routes` as a subcommand of `approve-routes`
- `list-routes` remains accessible directly via `headscale nodes list-routes -i <id>` for backward compatibility

Both command paths now work correctly:
- `headscale nodes list-routes -i 12`
- `headscale nodes approve-routes list-routes -i 12`

Fixes #2927

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,7 @@ connected" routers that maintain their control session but cannot route packets.
 - Remove deprecated `namespace`/`ns` command aliases for `users` and `machine`/`machines` aliases for `nodes` [#3093](https://github.com/juanfont/headscale/pull/3093)
 - **User deletion**: Fix `DestroyUser` deleting all pre-auth keys in the database instead of only the target user's keys [#3155](https://github.com/juanfont/headscale/pull/3155)
 - `headscale policy check` evaluates the `tests` block when invoked with `--bypass-grpc-and-access-database-directly`; without the flag it warns instead of running the tests against empty data [#1803](https://github.com/juanfont/headscale/issues/1803)
+- Fix `headscale nodes approve-routes list-routes -i <id>` so `-i` filters by node; previously the invocation silently cleared all approved routes on the target node [#3147](https://github.com/juanfont/headscale/pull/3147)
 
 #### API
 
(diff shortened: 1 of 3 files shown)


kradalby (approved):

nblock (changes requested):
This would add `headscale nodes approve-routes list-routes` which nests two commands.

PR #3148, opened 2026-03-17
hscontrol/servertest: add in-process control plane test harness

kradalby:
Playground for more tests more complex than unit tests, but not as complex as integration tests.

### DIFF
--- a/hscontrol/servertest/client.go
+++ b/hscontrol/servertest/client.go
@@ -266,6 +266,25 @@ func (c *TestClient) Reconnect(tb testing.TB) {
 		}
 	}
 
+	// Clear stale netmap data so that callers like WaitForPeers
+	// actually wait for the new session's map instead of returning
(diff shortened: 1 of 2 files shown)


hscontrol/app.go
```diff
@@ -1069,6 +1070,51 @@ func (h *Headscale) Change(cs ...change.Change) {
...
+
+// StartEphemeralGC starts the ephemeral node garbage collector.
+// It registers a cleanup function on tb to stop the collector.
+func (h *Headscale) StartEphemeralGC(tb testing.TB) {
+	tb.Helper()
+
+	go h.ephemeralGC.Start()
+
+	tb.Cleanup(func() { h.ephemeralGC.Close() })
+}
```
kradalby, line 1121: these should be called "ForTest" as a suffix as they are not intended to run outside. Maybe should panic outside.

PR #3179, opened 2026-04-10
Fix exit node routes when using autoApprovers

### LINKED ISSUE #3178: [Bug] Communication failure when enabling the Exit Node using autoApprovers

### Current Behavior

Communication failure when enabling the Exit Node using autoApprovers

Source Text) autoApprovers를 사용해 Exit Node를 활성화 했을 때 통신이 안되는 현상

### Expected Behavior

Internet communication must be routed through an exit node

Source Text) Exit Node로 인터넷 통신이 되어야 함

### Steps To Reproduce

https://github.com/user-attachments/assets/e3f0a58e-f11e-4868-9fed-9bbc30e8737d 

It's hard to explain, so I'll show you a video instead.
Source Text) 영상으로 대체하여 올립니다

### Environment

```markdown
- OS: Ubuntu 24.04.4 LTS (GNU/Linux 6.8.0-106-generic x86_64)
- Headscale version: 0.6.2
- Tailscale version: 1.94.1
- connected nodes: 8
- subnet router: 2 nodes
```

### Runtime environment

### Debug information
N/A

utafrali:
Fixes #3178

- [x] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

Ran into this when setting up an exit node with autoApprovers. The route approval logic wasn't correctly handling this case and was rejecting the exit node routes. Fixed the policy evaluator in policy.go to properly check routes when autoApprovers are involved.

Added unit tests to cover the scenario. Also did some cleanup on the debug output formatting while I was at it, switched from WriteString/Sprintf to fmt.Fprintf.

### DIFF
--- a/cmd/headscale/cli/nodes.go
+++ b/cmd/headscale/cli/nodes.go
@@ -175,8 +175,10 @@ Use --disable to disable key expiry (node will never expire).`,
 		now := time.Now()
 
 		expiryTime := now
+
 		if expiry != "" {
 			var err error
+
 			expiryTime, err = time.Parse(time.RFC3339, expiry)
 			if err != nil {
 				return fmt.Errorf("parsing expiry time: %w", err)
@@ -397,6 +399,7 @@ func nodesToPtables(
 		}
 
 		var ipBuilder strings.Builder
+
 		for _, addr := range node.GetIpAddresses() {
 			ip, err := netip.ParseAddr(addr)
 			if err == nil {
--- a/cmd/headscale/cli/policy.go
+++ b/cmd/headscale/cli/policy.go
@@ -63,6 +63,7 @@ var getPolicy = &cobra.Command{
 	Aliases: []string{"show", "view", "fetch"},
 	RunE: func(cmd *cobra.Command, args []string) error {
 		var policyData string
+
 		if bypass, _ := cmd.Flags().GetBool(bypassFlag); bypass {
 			if !confirmAction(cmd, "DO NOT run this command if an instance of headscale is running, are you sure headscale is not running?") {
 				return errAborted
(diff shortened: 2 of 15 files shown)


kradalby:
Thanks for looking into this and for the tests — the mutex move (`pm.mu.Lock()` before the exit route check) is a good catch.

However, the fallback to `autoApproveMap` for exit routes would diverge from SaaS behavior. We have investigated how the SaaS control server handles this and found that exit route approval **only** goes through the `exitNode` field — `0.0.0.0/0` in the `routes` map is deliberately ignored for exit approval.

The current code in `NodeCanApproveRoute` is actually correct:

```go
if tsaddr.IsExitRoute(route) {
    if pm.exitSet == nil {    // exitSet is built from autoApprovers.exitNode
        return false
    }
    if slices.ContainsFunc(node.IPs(), pm.exitSet.Contains) {
        return true
    }
    return false
}
```

The early return when `exitSet == nil` is intentional — it prevents the function from falling through to `autoApproveMap` (which contains the `routes` map entries). Adding the fallback:

```go
if approvers, ok := pm.autoApproveMap[route]; ok {
    return slices.ContainsFunc(node.IPs(), approvers.Contains)
}
```

would approve exit nodes through `routes: {"0.0.0.0/0": ["tag:exit"]}`, which SaaS does not do. The `routes` map entry for `0.0.0.0/0` serves a different purpose on SaaS — it acts as a wildcard for **subnet** route approval (e.g., auto-approving `172.16.0.0/24` for a `tag:exit` node that also advertises subnets).

The correct policy for exit node auto-approval is:

```json
"autoApprovers": {
    "exitNode": ["tag:exit"]
}
```

We have compatibility tests covering all the combinations (exitNode field, routes map, both, wrong tags, etc.) — happy to share the test scenarios if useful.

Would you be open to extracting just the mutex fix as a separate PR?

PR #3191, opened 2026-04-15
ci: add rolling development tag to container builds

kradalby:
Adds a rolling `development` tag alongside the existing `main-<sha>` tag for container images pushed to both GHCR and Docker Hub. This gives users a stable reference to track the latest main build without needing to look up commit SHAs.

### DIFF
--- a/.github/workflows/container-main.yml
+++ b/.github/workflows/container-main.yml
@@ -59,7 +59,7 @@ jobs:
           nix develop --command -- ko build \
             --bare \
             --platform=linux/amd64,linux/arm64 \
-            --tags=main-${GITHUB_SHA::7} \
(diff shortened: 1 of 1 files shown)


christian-heusel:
Nice, thanks for the addition 🥳 changes look good to me!

nblock:
FWIW: for docs we use `development` or `unstable` but not `nightly`.

kradalby:
yes, I should have elaborated, I am happy to change the name and maybe development is better, but I felt like unstable is more like beta and rc releases, while this is "it can break every time".

Maybe we should do `development`?

nblock:
> Maybe we should do `development`?

sounds good.

nblock (approved):

Issue #3170, opened 2026-04-05
[Bug] Tagged nodes' expiry reset from nil to zero-time on restart (incomplete #2862 fix)


eccgecko:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Key expiry for [tagged](https://tailscale.com/docs/features/tags#best-practices) devices is disabled by default.

After updating headscale to v0.28.0 and its introduction of tags, tagged nodes (registered via tag-only pre-auth keys) have their expiry reset from nil/unset to `0001-01-01 00:00:00` (Go zero-time) whenever tailscaled restarts or the node re-registers with Auth=nil.

e.g.:
Before restart:
  headplane-agent | tagged-devices | tag:headplane-agent | Expiration: N/A

After restart:
  headplane-agent | tagged-devices | tag:headplane-agent | Expiration: 0001-01-01 00:00:00

The node appears to continue to function (IsExpired() treats both nil and zero as "not expired"), but the expiry representation in the database and API changes from null to "0001-01-01T00:00:00Z", which affects downstream consumers (e.g. the headplane web-ui loses he "No expiry" badge because it checks `=== null`).

### Expected Behavior

Tagged nodes should preserve their nil expiry across tailscaled / headscale restarts.

The guard added in 3bd4ecd (fix for #2862) should also cover nodes with nil expiry.

### Steps To Reproduce

1. Create a tag-only pre-auth key:
 headscale preauthkeys create --tags tag:test-agent

2. Register a node using that key (e.g. via tsnet or `tailscale up --authkey=...`)

3. Confirm the node's expiry is N/A (nil)
 
 4. Restart the node's tailscaled (or restart the headscale server)

5. Check the node's expiry again:
   ```
   headscale nodes list
   # Expiration: 0001-01-01 00:00:00
   ```

**Root cause trace:**

When tailscaled restarts with persisted state, it sends `RegisterRequest` with `Auth=nil` and `Expiry=time.Time{}` (zero). In `handleRegister()` (auth.go:67-80):

  ```go
  if req.Auth == nil {
      if node, ok := h.state.GetNodeByNodeKey(req.NodeKey); ok {
          // Guard added in 3bd4ecd for #2862:
          if req.Expiry.IsZero() && node.Expiry().Valid() && !node.IsExpired() {
              return nodeToRegisterResponse(node), nil
          }
          resp, err := h.handleLogout(node, req, machineKey)
[...]

Issue #3111, opened 2026-02-26
[Bug] oidc expiry: 0 sets far-future date (year 2318) instead of NULL


guy0nabuffalo:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When the `oicd: expiry` option is set to `0` the expiry column in db.sqlite is set for 292 years from now.

### Expected Behavior

When the `oicd: expiry` option is set to `0` the expiry column in db.sqlite be set to NULL

### Steps To Reproduce

In config.yaml set
```
oicd:
  expiry: 0
```
Restart headscale
Enroll new machine
Run command `sqlite3 /root/docker/headscale/lib/db.sqlite "SELECT id, hostname, expiry FROM nodes;"`
New machine expiration set for 292 years from now. 
I realize this is basically forever but because it has an expiration, headplane does not add the `no expiry` tag. If the value instead is set to NULL the `no expiry` tag shows up in the headplane UI.

### Environment

```markdown
- OS: Ubuntu 22.04
- Headscale version: 0.28.0
- Tailscale version: 1.94.2
```

### Runtime environment

- [x] Headscale is behind a (reverse) proxy
- [x] Headscale runs in a container

### Debug information

Example output from my db.sqlite using the command: `sqlite3 /root/docker/headscale/lib/db.sqlite "SELECT id, hostname, expiry FROM nodes;"` The nodes that have a blank expiry column I manually set to NULL using the command: `sqlite3 /root/docker/headscale/lib/db.sqlite "UPDATE node SET expiry = NULL WHERE id = <node number>;"`
```
~# sqlite3 /root/docker/headscale/lib/db.sqlite "SELECT id, hostname, expiry FROM nodes;"
1|ubuntu2404|
5|localhost|
6|homeassistant|
7|homeassistant|
8|drewskivo|
9|jellyfin|2318-06-08 10:21:20.470938637-05:00
```

kradalby:
Fixed in #3122 and #3199 — oidc.expiry was removed in favor of node.expiry (which treats 0/unset as no default expiry, never a computed far-future date), and registration now explicitly clears expiry to nil instead of leaving a zero-time pointer.

PR #3199, opened 2026-04-16
state: preserve nil expiry on user owned registration when no default is configured

### LINKED ISSUE #3198: [Bug] User owned re registration with zero expiry stores 0001-01-01 instead of NULL

### Current Behavior

When a user owned node registers or re registers with a pre auth key, if the client does not send an expiry AND `node.expiry` is configured as `0` (no default), the `nodes.expiry` column gets `0001-01-01 00:00:00` (Go zero time) instead of NULL.

### Expected Behavior

The column should be NULL, matching the behaviour for tagged nodes and the value consumers use to distinguish "no expiry" from "expired at year 1". The headplane "no expiry" badge is an example of a consumer that checks `=== null`.

### Steps To Reproduce

1. Configure `node.expiry: 0` in headscale
2. `headscale preauthkeys create` for a user
3. Register a node with that key (client sends zero expiry by default)
4. `sqlite3 db.sqlite "SELECT id, hostname, expiry FROM nodes"` shows `0001-01-01 00:00:00` for that node
5. Expected: NULL

### Environment

```markdown
- OS: Linux
- Headscale version: current master (after 6337a3db)
- Tailscale version: any
```

### Runtime environment

### Debug information

Root cause is in `hscontrol/state/state.go`. Two sites build a non nil pointer to `regReq.Expiry` regardless of whether the value is zero time:

- Line 2132 (re registration with same MachineKey): `node.Expiry = &regReq.Expiry`
- Line 2231 (initial registration): `Expiry: &regReq.Expiry` passed to `createAndSaveNewNode`

Both were introduced in commit 6337a3db (fix for #3111). The `needsDefaultExpiry` guard at `state.go:1644` checks `.IsZero()` but only applies a default when `s.cfg.Node.Expiry > 0`, so when both the request and the config are zero the pointer to zero time survives to the database.

A regression test and fix are in an accompanying PR.

akhilesharora:
- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

When a user owned node registers or re registers with a PreAuthKey and the client sends zero expiry while `node.expiry` is set to `0`, the node's `expiry` column ends up as `0001-01-01 00:00:00` instead of NULL. `hscontrol/state/state.go` builds a pointer to `regReq.Expiry` unconditionally at two sites, even when the value is zero time. Downstream consumers that distinguish "no expiry" from "expired at year 1" (the headplane "no expiry" badge checks `=== null`) see the wrong state.

This change converts an unset `regReq.Expiry` to nil before handing it off, so the `needsDefaultExpiry` guard at `state.go:1644` is the only place that assigns a non nil pointer.

Added a regression test `TestReregistrationZeroExpiryStaysNil` that covers both initial and re registration paths with `node.expiry=0` and zero client expiry. It fails on master and passes with this change. Full `go test ./hscontrol/...` passes.

This is a narrower sibling of #3170 (tagged node on tailscaled restart) on the user owned PreAuthKey path. The regression was introduced alongside the fix for #3111 in commit 6337a3db.

Fixes #3198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -183,6 +183,7 @@ connected" routers that maintain their control session but cannot route packets.
 
 - Remove old migrations for the debian package [#3185](https://github.com/juanfont/headscale/pull/3185)
 - Install `config-example.yaml` as example for the debian package [#3186](https://github.com/juanfont/headscale/pull/3186)
+- **Node Expiry**: Fix user owned re registration with zero client expiry and no default storing `0001-01-01 00:00:00` in the database instead of NULL [#3199](https://github.com/juanfont/headscale/pull/3199)
 
 ## 0.28.0 (2026-02-04)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,7 @@ connected" routers that maintain their control session but cannot route packets.
 - Remove old migrations for the debian package [#3185](https://github.com/juanfont/headscale/pull/3185)
 - Install `config-example.yaml` as example for the debian package [#3186](https://github.com/juanfont/headscale/pull/3186)
 - **Node Expiry**: Fix user owned re registration with zero client expiry and no default storing `0001-01-01 00:00:00` in the database instead of NULL [#3199](https://github.com/juanfont/headscale/pull/3199)
+  - Pre-existing rows with `0001-01-01 00:00:00` are not backfilled; they clear themselves the next time the node re-registers
 
 ## 0.28.0 (2026-02-04)
 
--- a/hscontrol/auth_tags_test.go
+++ b/hscontrol/auth_tags_test.go
@@ -1119,3 +1119,67 @@ func TestReregistrationAppliesDefaultExpiry(t *testing.T) {
 	assert.True(t, node2.Expiry().Get().After(firstExpiry),
 		"re-registration expiry should be later than initial registration expiry")
 }
+
+// TestReregistrationZeroExpiryStaysNil tests that when a user-owned node
+// re-registers with zero client expiry and node.expiry is disabled (0),
+// the node's expiry stays nil rather than being set to a pointer to zero
+// time. Regression test for the else branch introduced in commit 6337a3db
+// which assigned `&regReq.Expiry` (pointer to time.Time{}) instead of nil,
+// causing the database row to hold `0001-01-01 00:00:00` instead of NULL.
+func TestReregistrationZeroExpiryStaysNil(t *testing.T) {
+	t.Parallel()
+
+	// node.expiry = 0 means "no default expiry"
+	app := createTestAppWithNodeExpiry(t, 0)
+
+	user := app.state.CreateUserForTest("node-owner")
+
+	pak, err := app.state.CreatePreAuthKey(user.TypedID(), true, false, nil, nil)
+	require.NoError(t, err)
+
+	machineKey := key.NewMachine()
+	nodeKey := key.NewNode()
+
+	// Initial registration with zero client expiry
+	regReq := tailcfg.RegisterRequest{
+		Auth: &tailcfg.RegisterResponseAuth{
+			AuthKey: pak.Key,
+		},
+		NodeKey: nodeKey.Public(),
+		Hostinfo: &tailcfg.Hostinfo{
+			Hostname: "reregister-zero-expiry",
+		},
+		Expiry: time.Time{},
+	}
+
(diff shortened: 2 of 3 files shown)


hscontrol/state/state.go
```diff
@@ -2216,6 +2218,15 @@ func (s *State) HandleNodeFromPreAuthKey(
 			pakUser = *pak.User
 		}
 
+		// Only pass the client-requested expiry when it is actually set.
+		// A pointer to a zero time.Time gets persisted as "0001-01-01 00:00:00"
+		// rather than NULL, which breaks downstream consumers that distinguish
+		// "no expiry" from "expires at year 1".
+		var reqExpiry *time.Time
```
kradalby, line 2225: From Claude:

```
Same `!IsZero → &x; else → nil` shape shows up twice now — here and in the re-registration `else` branch above. Not worth a helper for two sites, but if a third appears, a small `expiryOrNil(t time.Time) *time.Time` would read nicer than the branch. Fine as-is for now.
```

hscontrol/auth_tags_test.go
```diff
@@ -1119,3 +1119,67 @@ func TestReregistrationAppliesDefaultExpiry(t *testing.T) {
...
 		"re-registration expiry should be later than initial registration expiry")
 }
+
+// TestReregistrationZeroExpiryStaysNil tests that when a user-owned node
+// re-registers with zero client expiry and node.expiry is disabled (0),
+// the node's expiry stays nil rather than being set to a pointer to zero
+// time. Regression test for the else branch introduced in commit 6337a3db
+// which assigned `&regReq.Expiry` (pointer to time.Time{}) instead of nil,
+// causing the database row to hold `0001-01-01 00:00:00` instead of NULL.
+func TestReregistrationZeroExpiryStaysNil(t *testing.T) {
```
kradalby, line 1129: From Claude:

```
Test covers the user-owned path nicely. The initial-registration fix at `state.go:2218` also handles the tags-only PreAuthKey case (tagged node → `Expiry=nil`), but that isn't exercised here. The `!regReq.Expiry.IsZero()` gate at the call site makes it obvious by inspection, so not blocking — but a comment noting that the same fix covers the tagged path would help a future reader.
```

CHANGELOG.md
```diff
@@ -108,6 +108,7 @@ connected" routers that maintain their control session but cannot route packets.
 - Policy validation error messages now include field context (e.g., `src=`, `dst=`) and are more descriptive [#2180](https://github.com/juanfont/headscale/pull/2180)
 - Remove old migrations for the debian package [#3185](https://github.com/juanfont/headscale/pull/3185)
 - Install `config-example.yaml` as example for the debian package [#3186](https://github.com/juanfont/headscale/pull/3186)
+- **Node Expiry**: Fix user owned re registration with zero client expiry and no default storing `0001-01-01 00:00:00` in the database instead of NULL [#3199](https://github.com/juanfont/headscale/pull/3199)
```
kradalby, line 111: From Claude:

```
Pre-existing rows already holding `0001-01-01 00:00:00` aren't backfilled by this change — they keep the bad value until the node re-registers, at which point the fix naturally corrects them. If that turns out to be annoying for users with long-lived nodes, a one-shot `UPDATE nodes SET expiry = NULL WHERE expiry < '1900-01-01'` migration is cheap to add. Worth a line in the CHANGELOG or a followup issue noting this.
```

kradalby (approved):
Looks good, I think we should merge this, I did a little Claude review as well and got back a couple of things.

hscontrol/auth_tags_test.go
akhilesharora: Fixed in [4ab88296](https://github.com/juanfont/headscale/pull/3199/changes/4ab88296a9c5026cabd7f6cfb625cc2a02541000)

CHANGELOG.md
akhilesharora: Fixed in [4ab88296](https://github.com/juanfont/headscale/pull/3199/changes/4ab88296a9c5026cabd7f6cfb625cc2a02541000)

akhilesharora:
> Looks good, I think we should merge this, I did a little Claude review as well and got back a couple of things.

Thanks! Pushed 4ab88296 with the test comment and changelog line

kradalby:
looks good, there is a lint failure, and I think it needs a rebase of main too.

akhilesharora:
> looks good, there is a lint failure, and I think it needs a rebase of main too.

Rebased!

PR #3218, opened 2026-04-26
state: preserve username-based IP allocation

suluck97:
## Summary

Preserve username-based IP allocation on top of the current upstream main branch.

## Changes

- add namespace-specific IPv4 prefix allocation support in the IP allocator
- allocate node IPs based on the registering user's namespace
- load namespace prefix overrides from config
- update allocator tests to match the new behavior

## Notes

This change keeps the existing custom username-based IP allocation behavior after rebasing onto the latest upstream main branch.


kradalby:
Hi, we will not implement any alternative ip allocation strategies. We will eventually try to implement https://tailscale.com/docs/reference/ip-pool, but for now there will be no changes to this.

PR #3223, opened 2026-04-28
policy: surface exit nodes via autogroup:internet

### LINKED ISSUE #3212: [Bug] Exit nodes no longer work in main

### Current Behavior

Listing or setting an exit-node no longer works in main as of 2026-04-18. It doesn't matter if the exit-node is user-assigned or tagged.

### Expected Behavior

The steps in https://headscale.net/stable/ref/routes/#setup-an-exit-node should work.

Policy:

```json
{
  "acls": [
    {
      "action": "accept",
      "src": ["alice@"],
      "dst": ["autogroup:internet:*"]
    }
  ]
}
```

### Steps To Reproduce

- Create a user alice: `headscale user create alice`
- Register an exit node for alice:
  - `sysctl -w net.ipv4.ip_forward=1 net.ipv6.conf.all.forwarding=1`
  - `tailscale up --login-server https://headscale.example.com --advertise-exit-node`
- Approve routes: `headscale nodes approve-routes --identifier 1 -r 0.0.0.0/0`
- Register a regular node for alice
- Try to list or set an exit node

```
$ tailscale exit-node list
no exit nodes found

$ tailscale set --exit-node=100.64.0.1
no node found in netmap with IP 100.64.0.1
```

The above steps work as-is on 0.28.0

### Environment

```markdown
- OS: Debian 13
- Headscale version: v0.0.0-20260418141221-4e1d83ecefeb
- Tailscale version: 1.96.4
```

### Runtime environment

### Additional information

- [Direct access to the exit node is not required in order to use it](https://tailscale.com/docs/features/exit-nodes?q=autogroup%3Ainternet#prerequisites)
  Image
- [Grants: allow using an exit node](https://tailscale.com/docs/reference/examples/grants#allow-using-exit-nodes)

kradalby:
`compileFilterRules` skipped `autogroup:internet` destinations to keep them out of the wire-format `PacketFilter`, but those compiled rules also feed `pm.matchers`. `Node.CanAccess` looks at a matcher whose `DestsIsTheInternet` covers the public internet to surface exit-node peers to ACL sources — with the skip in place no such matcher existed, exit nodes silently dropped out of the source's peer list, and the `ref/routes/#setup-an-exit-node` walkthrough stopped working: `tailscale exit-node list` returned `no exit nodes found` and `tailscale set --exit-node=<ip>` returned `no node found in netmap with IP`.

The skip lands in `filter.go`. Removing it lets `autogroup:internet` flow through normal matcher derivation. To keep the wire-format PacketFilter correct on advertisers, `ReduceFilterRules` now keeps a destination on exit-route advertisers when every prefix sits inside `util.TheInternet()` — Tailscale SaaS sends those rules to exit nodes so the advertiser's kernel filter accepts traffic forwarded by `autogroup:internet` sources.

Verified against a live Tailscale SaaS tailnet on 2026-04-28 via `tscap`. The b17/b18 captures land under `testdata/issue_3212/` as a regression guard. They are isolated from `testdata/routes_results/` because the broader `TestRoutesCompat` machinery assumes a CIDR-prefix wire format that differs from the IPSet-range form SaaS emits for `autogroup:internet`; aligning that wire format is tracked separately.

Fixes #3212

### DIFF
--- a/hscontrol/policy/policyutil/reduce.go
+++ b/hscontrol/policy/policyutil/reduce.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/juanfont/headscale/hscontrol/types"
 	"github.com/juanfont/headscale/hscontrol/util"
+	"go4.org/netipx"
 	"tailscale.com/tailcfg"
 )
 
@@ -18,6 +19,7 @@ import (
 func ReduceFilterRules(node types.NodeView, rules []tailcfg.FilterRule) []tailcfg.FilterRule {
 	ret := []tailcfg.FilterRule{}
 	subnetRoutes := node.SubnetRoutes()
+	hasExitRoutes := node.IsExitNode()
 
 	for _, rule := range rules {
 		// Handle CapGrant rules separately — they use CapGrant[].Dsts
@@ -56,6 +58,14 @@ func ReduceFilterRules(node types.NodeView, rules []tailcfg.FilterRule) []tailcf
 			// AllowedIPs/routing.
 			if slices.ContainsFunc(subnetRoutes, expanded.OverlapsPrefix) {
 				dests = append(dests, dest)
+				continue
+			}
+
+			// Exit-route advertisers need rules targeting the
+			// public internet so the kernel filter accepts
+			// traffic forwarded by autogroup:internet sources.
+			if hasExitRoutes && ipSetSubsetOf(expanded, util.TheInternet()) {
+				dests = append(dests, dest)
 			}
 		}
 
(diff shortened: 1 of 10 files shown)


hscontrol/policy/policyutil/reduce.go
```diff
@@ -18,6 +19,15 @@ import (
...
 	subnetRoutes := node.SubnetRoutes()
+	hasExitRoutes := node.IsExitNode()
+
+	// Cache TheInternet() lookup outside the per-rule loop. The IPSet
+	// is constant for the lifetime of the process; ContainsRange runs
+	// for every dest checked when the node advertises exit routes.
+	var internet *netipx.IPSet
+	if hasExitRoutes {
+		internet = util.TheInternet()
+	}
```
kradalby: Is this necessary? TheInternet returns a once, so we can just call it when needed?

juanfont (approved):

nblock (approved):

PR #3224, opened 2026-04-28
state: fix HA tracking losing secondary after all routers offline

### LINKED ISSUE #3203: [Bug] HA tracking loses secondary once all subnet routers are offline

### Current Behavior

The new HA tracking is losing track of secondary nodes as soon as all subnet routers are offline at once.

### Expected Behavior

To be able to recover as soon as one of the subnet routers comes back online.

### Steps To Reproduce

Policy:

```json5
{
  "autoApprovers": {
    "routes": {
      "10.0.0.0/24": [
        "alice@"
      ]
    }
  },
  "acls": [
    {
      "action": "accept",
      "src": [
        "100.64.0.3"  //n3
      ],
      "dst": [
        "10.0.0.1:8000"
      ]
    }
  ]
}
```

- Prepare two subnetrouters (n1, n2) that offer the same route: 10.0.0.0/24
  ```
  sysctl -w net.ipv4.ip_forward=1
  ip link add dummy0 type dummy 
  ip addr add 10.0.0.1/24 dev dummy0 
  ip link set dummy0 up
  python3 -m http.server -b 10.0.0.1
  ```
- register n1, n2 for alice: `tailscale up --login-server https://headscale.example.com --advertise-routes=10.0.0.0/24`
- register regular node (n3) for alice: `tailscale up --login-server https://headscale.example.com --accept-routes=true`
- on n3: `while :; do curl -m 2 http://10.0.0.1:8000/ ; sleep 1 ; done`

---

All nodes are up- and running. The following sequence triggers the issue:

- n1 is primary, n1 is serving traffic: OK
- pull network cable on n1
- failover to n2, n2 is serving traffic: OK
- pull network cable on n2
- failover to n1 (but this node is offline): probably not OK, it should probably stay with n2
- connect network cable on n2
- n2 is not recognised as online: not OK
- From this point onwards, HA tracking of n2 is lost until Headscale is restarted.

### Environment

```markdown
- OS: Debian 13
- Headscale version: v0.0.0-20260416141056-1b6ab52f9e24
- Tailscale version: 1.96.4
```

### Runtime environment

### Debug information

None.

kradalby:
When all subnet routers serving a prefix go offline simultaneously and one returns, headscale could end up with the returning node marked online in `NodeStore` but absent from `PrimaryRoutes` for the prefix, with HA tracking stuck until daemon restart.

The generation counter added in b09af384 closed the obvious window but left the gen check non-atomic with the subsequent `NodeStore.UpdateNode` and `PrimaryRoutes.SetRoutes` calls. A concurrent `Connect` that bumped the counter and rewrote primary routes could be silently overtaken by a stale `Disconnect`'s `SetRoutes(empty)`. The grace-period flow in `poll.go` hides this on most cable-pull cadences, which is why three earlier reproductions (servertest graceful, integration graceful, integration cable-pull-via-iptables) all pass on main.

Replace the standalone counter with a per-node `nodeLock` pairing the gen counter and a mutex. `Connect` and `Disconnect` now run under the lock and re-check gen inside the critical section, making gen evaluation and the mutations atomic. Per-node so unrelated nodes never contend; same-node steady state has effectively no contention.

`TestConnectDisconnectRace` drives the race directly through `State`, fails deterministically on main, and passes under the lock. The earlier servertest, integration and cable-pull reproductions stay as regression coverage for the user-reported sequence. A property test on `PrimaryRoutes` joins as a safety net for future changes to primary route selection.

Fixes #3203


hscontrol/types/node.go
```diff
@@ -154,6 +154,20 @@ type Node struct {
...
+	// Default zero value (false) means "healthy or not yet probed", matching
+	// the legacy primaryRoutes.unhealthy set semantics.
+	Unhealthy bool `gorm:"-"`
+
+	// SessionEpoch is bumped by State.Connect inside the NodeStore writer
+	// goroutine and captured by the caller. State.Disconnect refuses to
+	// flip IsOnline=false unless the captured epoch still matches the
+	// current value, rejecting stale grace-period disconnects from old
+	// poll sessions. Runtime-only; replaces the external connectGen
+	// counter (issue #3203).
```
kradalby: brevity in the comments, focus on why, not what, not history.

hscontrol/state/node_store.go
```diff
@@ -144,6 +147,15 @@ type Snapshot struct {
...
 	allNodes          []types.NodeView
+
+	// primaries holds the per-prefix primary route assignment derived
+	// from the current set of online, healthy advertisers. It is
+	// recomputed on every snapshot rebuild and inherits the previous
+	// snapshot's assignments for anti-flap (preserve the current
+	// primary if it is still a valid advertiser and not unhealthy).
+	// Replaces routes.PrimaryRoutes (issue #3203 follow-up).
+	primaries map[netip.Prefix]types.NodeID
+	isPrimary map[types.NodeID]bool
```
kradalby: primary works as a name when we are inside the route package, but now we are inside the node store, and primary could mean many things and not only routes. So we we should probably do something like `routes` and `isPrimaryRoute`.

kradalby: Same note about comments, brevity, focus on why.

hscontrol/state/state.go
```diff
@@ -632,50 +596,27 @@ func (s *State) Connect(id types.NodeID) ([]change.Change, uint64) {
...
-			Msg("stale disconnect rejected, newer session active")
-
-		return nil, nil
-	}
-
+// The check and the IsOnline write happen inside the same NodeStore
+// UpdateNode closure, so they are atomic against any concurrent
+// Connect for the same node (the writer goroutine serialises batches).
+// Primary routes derive from the snapshot rebuild; no separate
+// primaryRoutes mutation is needed (issue #3203).
```
kradalby: Same note about comments, brevity, focus on why.

juanfont (approved):

nblock (changes requested):
The issue can be reproduced as described in the ticket. "Pulling the network cable" means a disconnect of the host in the hypervisor

nblock (approved):

PR #3228, opened 2026-04-29
cmd/vendorhash: track vendor SRI in flakehashes.json

kradalby:
Replaces the manual `flake.nix` vendor hash literal with a content-
addressed `flakehashes.json` file maintained by a small Go tool that
imports `tailscale.com/cmd/nardump/nardump.SRI` directly. Schema and
fingerprint algorithm mirror upstream's `tool/updateflakes` from
[tailscale/tailscale#19551](https://github.com/tailscale/tailscale/pull/19551)
so a future shared library extraction is trivial.

Today the hash lives as a string in `flake.nix:30` and the only
divergence detector is the CI step that scrapes `specified:` /
`got:` lines from `nix build` stderr. The round-trip is push → wait
for CI → read PR comment → paste new hash. Brittle and slow.

`vendorhash check` does a sha256 over `go.mod` and `go.sum`. If the
fingerprint matches `flakehashes.json`, exit silently. Only an actual
drift triggers `go mod vendor` + `nardump.SRI`. That makes the new
prek hook cheap enough to run on every commit touching `go.mod`,
catching staleness locally instead of at CI time. CI itself now runs
the same tool and reads structured output, dropping the grep/awk.

Pulls in `tailscale.com v1.97.0-pre.0.20260429005429-40088602c960`
to get the nardump library split (post-`v1.96.4`); cascades into a
go directive bump to `1.26.2` and a nixpkgs lock advance for the
matching toolchain (which incidentally renamed
`nodePackages.prettier` to top-level `prettier`).


.pre-commit-config.yaml
```diff
@@ -60,3 +60,13 @@ repos:
         language: system
         types: [go]
         pass_filenames: false
+
+      # vendor-hash keeps flakehashes.json in sync with go.mod/go.sum.
+      # Hot path (no input change) is a sha256 over two small files;
+      # only fires `go mod vendor` when the fingerprint actually drifts.
```
nblock, line 66: The comment is misleading as it does not run `go mod vendor` in any case.

nblock (approved):

PR #3229, opened 2026-04-29
policy/v2: add support for policy tests

### LINKED ISSUE #1803: Support ACL Tests

kradalby:
Tailscale policies have a `tests` block where the operator says, for a given source, what should be reachable. v2 ignored it — a policy whose tests contradicted its own ACLs would still apply.

`SetPolicy` now compiles the new policy in a sandbox, runs the tests, and rejects on failure without touching the live `PolicyManager`. `NewPolicyManager` skips evaluation so a stale stored policy whose referenced user was deleted does not block boot. `headscale policy check` opens the database under `--bypass-grpc-and-access-database-directly` (same flag `set` and `get` already accept) and runs tests against live data.

Error messages match Tailscale SaaS so policies move between control planes without rewriting. Comparing against captured SaaS responses surfaced two engine bugs. `Protocol.UnmarshalJSON` stored whichever form the user wrote, so `proto: "6", dst: "host:443"` was rejected at parse though SaaS accepts it — `UnmarshalJSON` now canonicalises numeric IANA forms to the name. `srcReachesDst` passed empty Protocol through to `ruleMatchesProto`, which short-circuited any rule with an `IPProto` restriction, so a TCP-only grant could not satisfy a default-proto test — empty Protocol now expands to `{TCP, UDP, ICMP, ICMPv6}` at the call site.

Parse-time schema validation rejects test shapes SaaS rejects: single-port dst, proto in `{tcp, udp, sctp, ""}`, no `autogroup:internet` dst, no CIDR-typed dst, at least one of `accept`/`deny`.

Closes #3005

Fixes #1803

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,23 @@ A new `headscale auth` CLI command group supports the approval flow:
 [#1850](https://github.com/juanfont/headscale/pull/1850)
 [#3180](https://github.com/juanfont/headscale/pull/3180)
 
+### Policy tests (beta)
+
+Headscale now evaluates the `tests` block in a policy file. Tests assert reachability between
+named sources and destinations and cover the whole policy — both `acls` and `grants` rules
+contribute. They run on user-initiated writes via `headscale policy set`, on SIGHUP reload
+(`systemctl reload headscale` / `kill -HUP $(pidof headscale)`), and on `headscale policy check`.
+A failing test rejects the write before it is applied, with the same error message Tailscale SaaS
+would return for the same policy.
+
+At boot a stored policy whose tests no longer pass — for example because a referenced user was
+deleted while the server was offline — logs a warning and the server keeps running. Fix the
+policy and reload.
+
+This feature is **beta** while behavioural coverage against Tailscale SaaS broadens.
+
+[#3229](https://github.com/juanfont/headscale/pull/3229)
+
 ### Grants
 
 We now support [Tailscale grants](https://tailscale.com/docs/features/access-control/grants)
@@ -134,6 +151,7 @@ connected" routers that maintain their control session but cannot route packets.
 - Fix exit node approval not triggering filter rule recalculation for peers [#2180](https://github.com/juanfont/headscale/pull/2180)
 - Policy validation error messages now include field context (e.g., `src=`, `dst=`) and are more descriptive [#2180](https://github.com/juanfont/headscale/pull/2180)
 - Reject policies whose `user@` tokens match multiple DB users; rename the duplicate via `headscale users rename` to load [#3160](https://github.com/juanfont/headscale/issues/3160)
+- Evaluate the policy `tests` block on user-initiated writes across both `acls` and `grants`; reject policies whose tests fail (beta) [#1803](https://github.com/juanfont/headscale/issues/1803)
 
 #### Grants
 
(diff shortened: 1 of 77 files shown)


cmd/headscale/cli/policy.go
```diff
@@ -179,36 +187,70 @@ var checkPolicy = &cobra.Command{
...
-			defer d.Close()
 
-			users, err = d.ListUsers()
-			if err != nil {
-				return fmt.Errorf("loading users for policy validation: %w", err)
+			if policyHasTestsBlock(policyBytes) {
+				return errTestsRequireBypass
 			}
+
+			fmt.Println("Policy syntax is valid (run with --" + bypassFlag + " to also validate user references against the database)")
```
nblock: Invoking `headscale policy check --bypass-grpc-and-access-database-directly --file /path/to/policy.json` does not perform checks as it fails before during startup:

```
DO NOT run this command if an instance of headscale is running, are you sure headscale is not running? [y/n] y
Error: loading config: Fatal config error: headscale now requires a new `noise.private_key_path` field in the config file for the Tailscale v2 protocol
Fatal config error: the only supported values for tls_letsencrypt_challenge_type are HTTP-01 and TLS-ALPN-01
Fatal config error: server_url must start with https:// or http://
Fatal config error: node.ephemeral.inactivity_timeout (0s) is set too low, must be more than 1m5sFatal config error: tuning.node_store_batch_size must be positive, got 0
Fatal config error: tuning.node_store_batch_timeout must be positive, got 0s
```

Passing the global `--config /path/to/config.yaml` yields `Error: unknown flag: --config`.

NOTE: the existing `headscale policy get --bypass-grpc-and-access-database-directly` works fine.

The used config also works fine with `headscale serve`

nblock (changes requested):

CHANGELOG.md
```diff
@@ -27,6 +27,22 @@ A new `headscale auth` CLI command group supports the approval flow:
...
+
+Headscale now evaluates the `tests` block in a policy file. Tests assert reachability between
+named sources and destinations and cover the whole policy — both `acls` and `grants` rules
+contribute. They run on user-initiated writes via `headscale policy set`, the file watcher, and
+`headscale policy check`. A failing test rejects the write before it is applied, with the same
+error message Tailscale SaaS would return for the same policy.
+
+Tests do not run at boot. An already-stored policy that no longer passes — for example because a
+referenced user was deleted while the server was offline — logs a warning and the server keeps
+running.
```
nblock: Starting with a `tests` block that should fail does not emit a warning during startup.

CHANGELOG.md
```diff
@@ -27,6 +27,22 @@ A new `headscale auth` CLI command group supports the approval flow:
 [#1850](https://github.com/juanfont/headscale/pull/1850)
 [#3180](https://github.com/juanfont/headscale/pull/3180)
 
+### Policy tests (beta)
+
+Headscale now evaluates the `tests` block in a policy file. Tests assert reachability between
+named sources and destinations and cover the whole policy — both `acls` and `grants` rules
+contribute. They run on user-initiated writes via `headscale policy set`, the file watcher, and
```
nblock: The file watcher does not watch the policy file. Headscale has to be explicitly informed about a changed policy file with `kill -HUP $(pidof headscale)` or `systemctl reload headscale`.

nblock (changes requested):

nblock (approved):

Issue #2912, opened 2025-11-22
[Feature] IP pool


Tillerino:
### Use case

Tailscale has had the [IP pool](https://tailscale.com/kb/1304/ip-pool) feature in beta for a while now. I was wondering if there is interest to implement this at all and if it would be implemented before it leaves beta in tailscale.

### Description

```yaml
    {
        "grants": ["..."],
        "nodeAttrs": [
            {
                "target": ["autogroup:admin"],
                "ipPool": ["100.81.0.0/16"],
            },
            {
                "target": ["group:dev"],
                "ipPool": ["100.85.0.0/16"],
            },
        ],
    }
```

The way that I understand the feature, you can make sure that certain nodes receive IPs from subranges within your IP prefix. In the example (copied from the linked doc), IP pools are only assigned to groups. I don't know which targets would be valid here. I am also not certain if new devices that are _not_ in one of those groups are guaranteed to receive IPs from _outside_ of those pools, though that would certainly make sense.

### Contribution

- [x] I can write the design doc for this feature
- [x] I can contribute this feature

kradalby:
As I am working on classifying features and figuring out when and how to work on it, I had a quick look at what likely would be how this works:

First we will need to implement the `nodeAttrs` support: https://github.com/juanfont/headscale/issues/2319

Then ultimately I think this means that we need to make IP allocator understand multiple smaller subnets, if a node has not `nodeAttr` with `ippool` it should allocate from a specific subnet, and if it does not have one, it should allocate from "the rest".

This means that when we allocate an IP, we need to not only ask for a new IP, but for a new IP for a given "nodeAttr identifier". It also means that the IP allocator needs to be initiated with information from the policy, maybe even have access to it while running. 

This will make the currently "independent" (only reads the database on startup), component that is IP allocator more tied in with the rest of the system.

These are notes for myself/others for later, and they might be incorrect, or the implementation might differ, but it will need some thought.

kradalby:
Related: #3133 (manual node IP assignment). Both touch node IP allocation and should be designed together — whichever approach we pick likely settles the other. Not scheduled for a milestone yet; leaving linked for future triage.

PR #3245, opened 2026-05-03
policy: implement nodeAttrs support

### LINKED ISSUE #2319: [Feature] Support for node attributes (nodeattrs)

ghost:
Closes #2319

This PR adds support for `nodeAttrs` in the Tailscale ACL policy, allowing operators to attach arbitrary capabilities to nodes via the policy file. These capabilities are distributed to clients through the CapMap field in MapResponse, enabling Tailscale client-side features that depend on node attributes.

## What is implemented

- **HuJSON parsing**: `nodeAttrs` array is parsed from the policy file alongside existing `acls` and `grants`.
- **Target matching**: Supports user emails, tags (`tag:server`), IP addresses (IPv4/IPv6), and wildcard (`*`).
- **Tag ownership semantics**: Respects the existing "tags XOR user ownership" rule—tagged nodes are owned by `TaggedDevices`, not the creating user.
- **CapMap distribution**: Compiled `NodeCapMap` is included in `MapResponse` for all affected nodes on policy changes.
- **Self-node invalidation**: When a policy change affects nodeAttrs, the mapper correctly invalidates the self node so clients receive updated CapMap without reconnecting.

## Test coverage

- `hscontrol/policy/v2/nodeattrs_test.go`: Unit tests for all target types (user, tag, IP, wildcard) and multiple-grant merge semantics.
- `hscontrol/servertest/nodeattrs_test.go`: End-to-end server-level tests verifying CapMap appears in Netmap after policy reload.
- `integration/nodeattrs_test.go`: Docker-based E2E tests verifying real tailscale clients receive and apply nodeAttrs dynamically.
- `BenchmarkCompileNodeAttrs`: Micro-benchmark for regression tracking of CapMap compilation cost.

## Compatibility

- Old clients (CapVer < 81) ignore CapMap; behavior is unchanged.
- Policies without `nodeAttrs` behave exactly as before.
- No database migration required.

## Future work (out of scope)

Some nodeAttrs require server-side features beyond CapMap distribution (e.g., `ippool` in #2912). This PR only covers the "string-only" CapMap path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hscontrol/mapper/builder.go
+++ b/hscontrol/mapper/builder.go
@@ -90,6 +90,16 @@ func (b *MapResponseBuilder) WithSelfNode() *MapResponseBuilder {
 		return b
 	}
 
+	// Merge policy nodeAttrs CapMap into the self node.
+	if capMap := b.mapper.state.NodeCapMap(nv); capMap != nil {
+		if tailnode.CapMap == nil {
+			tailnode.CapMap = make(tailcfg.NodeCapMap)
+		}
+		for k, v := range capMap {
+			tailnode.CapMap[k] = v
+		}
+	}
+
 	b.resp.Node = tailnode
 
 	return b
@@ -264,6 +274,16 @@ func (b *MapResponseBuilder) buildTailPeers(peers views.Slice[types.NodeView]) (
 		}, b.mapper.cfg)
 		if err != nil {
 			return nil, err
+		}
+
+		// Merge policy nodeAttrs CapMap into the peer node.
+		if capMap := b.mapper.state.NodeCapMap(peer); capMap != nil {
+			if tn.CapMap == nil {
+				tn.CapMap = make(tailcfg.NodeCapMap)
+			}
+			for k, v := range capMap {
+				tn.CapMap[k] = v
+			}
 		}
 
 		tailPeers = append(tailPeers, tn)
--- a/hscontrol/policy/v2/compiled.go
+++ b/hscontrol/policy/v2/compiled.go
@@ -1,6 +1,7 @@
 package v2
 
 import (
+	"maps"
 	"net/netip"
 	"slices"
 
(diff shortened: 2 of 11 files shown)


hscontrol/policy/v2/compiled.go
```diff
@@ -94,6 +95,78 @@ func buildUserNodeIndex(
 	return idx
 }
 
+// compileNodeAttrs resolves nodeAttrs into a per-node CapMap.
+// Each target string is parsed as an Alias and resolved against the current
+// nodes. Matching nodes receive all capabilities listed in Attrs.
+func (pol *Policy) compileNodeAttrs(
+	users types.Users,
+	nodes views.Slice[types.NodeView],
+) map[uint]tailcfg.NodeCapMap {
```
kradalby, line 104: `uint` -> `types.NodeID`

hscontrol/policy/v2/policy.go
```diff
@@ -531,6 +537,23 @@ func (pm *PolicyManager) FilterForNode(node types.NodeView) ([]tailcfg.FilterRul
 	return pm.filterForNodeLocked(node), nil
 }
 
+// NodeCapMap returns the CapMap compiled from nodeAttrs for the given node.
+// It includes capabilities granted via the policy file's nodeAttrs section.
+func (pm *PolicyManager) NodeCapMap(node types.NodeView) tailcfg.NodeCapMap {
```
kradalby, line 542: Is there a distinction between PeerCapMap and NodeCapMap? can we safely give the same map to both?

kradalby, line 542: Probably sufficient to only send `id types.NodeID` here.

hscontrol/policy/v2/types.go
```diff
@@ -1898,6 +1898,14 @@ type Grant struct {
 	Via []Tag `json:"via,omitzero"`
 }
 
+// NodeAttrGrant maps a set of node attributes to nodes matching a target.
+// Targets are strings that can be resolved as aliases (tags, users, groups,
+// autogroups, or wildcard).
+type NodeAttrGrant struct {
+	Targets []string `json:"target"`
```
kradalby, line 1905: Could Targets here be a stricter type than `string`? Something like `Alias`?

hscontrol/state/state.go
```diff
@@ -311,7 +311,10 @@ func (s *State) ReloadPolicy() ([]change.Change, error) {
 	s.nodeStore.RebuildPeerMaps()
 
 	//nolint:prealloc // cs starts with one element and may grow
-	cs := []change.Change{change.PolicyChange()}
+	c := change.PolicyChange()
+	c.IncludeSelf = true // nodeAttrs may affect self CapMap
```
kradalby, line 315: We should not set `IncludeSelf` unconditionally, we should only set it if the actual node has changed.

integration/tsic/tsic.go
```diff
@@ -519,7 +519,7 @@ func New(
 				)
 			}
 		}
-	case "unstable":
```
kradalby, line 522: Unrelated change, please remove.

integration/nodeattrs_test.go
```diff
@@ -0,0 +1,434 @@
+package integration
```
kradalby, line 1: We are trying to avoid adding new integration tests when `servertest` would be sufficient for the test case, which I think is the case for the test bellow. 

However, what would be the good use of `nodeAttrs` and integration test here would be to correctly wire up so we handle Taildrive, peer relay, SSH and other nodeAttrs correctly instead of unconditionally enabling them:
https://github.com/juanfont/headscale/blob/76ee29352b46659fc3ef16d8464bb50eb23ba497/hscontrol/types/node.go#L1184-L1200

Implementing them correctly and adding them to their respective integration tests would be great use of integration tests in synergy with this work.

If this PR is to go in, we need to implement the correct cap behaviour for the example above anyways so we do not get a split configuration.

hscontrol/mapper/builder.go
```diff
@@ -266,6 +276,16 @@ func (b *MapResponseBuilder) buildTailPeers(peers views.Slice[types.NodeView]) (
 			return nil, err
 		}
 
+		// Merge policy nodeAttrs CapMap into the peer node.
+		if capMap := b.mapper.state.NodeCapMap(peer); capMap != nil {
```
kradalby, line 280: Same question as earlier, is it safe to dump the same NodeCapMap into peers and self node?

kradalby (changes requested):
This is refreshing and might be the most promising looking LLM PR that anyone has sent us and it fits quite well with the current release. It seems _reasonable_, but still has some way to go. I've posted comments on some changes, some questions that needs investigation and some missing parts. 

One part that is going to be tricky is to "get it correct". This does require some extended testing and actually running nodes agents Tailscale SaaS to validate that we have the same behaviour so we do not have to make breaking changes in the future.

In addition, as mentioned in the comment about taildrive, peer relay and so on, we need to align all of these into the new support and that does mean we are making a breaking change as internal behaviour is changing and it will require users to explicitly enable the ones of these that are not new in the 0.29 release. Changelog must be updated. 

Please be mindful when you are steering the next LLM session and dont blindly ask it to "fix all my comments". LLMs work well when you steer them.

PR #3269, opened 2026-05-18
policy/v2: match via-grant destinations by prefix overlap

### LINKED ISSUE #3267: [Bug] via grant does not match subnet router when destination is a broader prefix than the advertised route

### Current Behavior

Headscale Policy:
```yaml
{
  "tagOwners": {
    "tag:subnet-router": ["exampleuser@"]
  },
  "hosts": {
    "example-4via6": "fd7a:115c:a1e0:b1a::/64"
  },
  "grants": [
    {
      "src": ["exampleuser@"],
      "dst": ["example-4via6"],
      "via": ["tag:subnet-router"],
      "ip": ["tcp:3389", "icmp:*"]
    }
  ]
}
```

The subnet router announces and has approved the subnet:
fd7a:115c:a1e0:b1a:0:13:ad2:7300/120

Client devices associated with the user `exampleuser` cannot ping devices within that subnet.

### Expected Behavior

Client devices associated with the user exampleuser should be able to ping devices in the range fd7a:115c:a1e0:b1a:0:13:ad2:7300/120

### Steps To Reproduce

- Build Headscale from the latest main branch.

- Create the following ACL file for Headscale:
```yaml
{
  "tagOwners": {
    "tag:subnet-router": ["exampleuser@"]
  },
  "hosts": {
    "example-4via6": "ffd7a:115c:a1e0:b1a::/64"
  },
  "grants": [
    {
      "src": ["exampleuser@"],
      "dst": ["example-4via6"],
      "via": ["tag:subnet-router"],
      "ip": ["tcp:3389", "icmp:*"]
    }
  ]
}
```
- Create the user exampleuser.

- Add a test node, advertise and approve the subnet:
fd7a:115c:a1e0:b1a:0:13:ad2:7300/120

- Add the tag subnet-router to the node.

Additionally, you have to add another grant for the node to become visible to the user’s node:

```yaml
{
  "src": ["exampleuser@"],
  "dst": ["tag:subnet-router"],
  "ip":  ["icmp:*"],
}
```
This should not be necessary, but I already saw some issues mentioning this. In my tests, the node was not visible to the source node without this grant, but this should not have anything to do with the issue

I think this issue comes down to exact prefix matching in the via path.

In `compileViaForNode()` and `ViaRoutesForPeer()`, Headscale appears to check whether the grant `dst` prefix is exactly contained in the subnet router's `SubnetRoutes()`:

```go
slices.Contains(nodeSubnetRoutes, dstPrefix)
```

This only works when the policy destination exactly matches the route advertised and approved by the subnet router.

If dst is a broader host alias or broader literal prefix, for example fd7a:115c:a1e0:b1a ::/64, and the router advertises a contained route such as fd7a:115c:a1e0:b1a:0:13:ad2:7300/120, the comparison fails even though the advertised route is inside the allowed destination prefix.

### Environment

```markdown
- OS: Debian 13
- Headscale version: v0.28.1-0.20260517161527-b50d593f2eaa
- Tailscale version: 1.96.3-t3ffddb134-g460d8764a
```

### Runtime environment

### Debug information

Not applicable

kradalby:
`compileViaForNode` (`hscontrol/policy/v2/compiled.go:686`) and `ViaRoutesForPeer` (`hscontrol/policy/v2/policy.go:998, 1089`) used `slices.Contains(subnetRoutes, dstPrefix)` to decide whether a node advertised a grant destination. Equality rejects any non-identical prefix pair, so a via grant with broader or narrower `dst` than the advertised subnet emitted no filter rule and added no route to the viewer's `AllowedIPs`. Tailscale SaaS uses containment in either direction.

Switch the filter-rule path to `slices.ContainsFunc(routes, dst.Overlaps)` and keep the literal `dst` in `DstPorts`. Append overlapping advertised routes to `ViaRoutesForPeer.Include` / `Exclude` rather than the dst. Rewrite the multi-router HA election and regular-grant overlap to key off the matched routes. Resolve `*Host` aliases to `*Prefix` once in `compileOneViaGrant` and at the top of `ViaRoutesForPeer` so the type switch reaches them.

Five new `testdata/grant_results/via-grant-v47..v51.hujson` lock the SaaS shape: broader literal dst, narrower literal dst, host-alias dst, disjoint dst (negative), and the reporter's exact 4via6 case. `TestGrantsCompat` replays them; existing v14-v46 captures stay green.

Fixes #3267

### DIFF
--- a/hscontrol/policy/v2/compiled.go
+++ b/hscontrol/policy/v2/compiled.go
@@ -66,12 +66,52 @@ type selfGrantData struct {
 }
 
 // viaGrantData holds data needed for per-node via-grant compilation.
-// Sources are already resolved into srcIPStrings.
+// Sources are already resolved into srcIPStrings; destinations are
+// pre-resolved into prefixes plus a flag for autogroup:internet, which
+// the consumers handle separately because its gate is per-node
+// (IsExitNode()) rather than per-prefix overlap.
 type viaGrantData struct {
-	viaTags           []Tag
-	destinations      Aliases
-	internetProtocols []ProtocolPort
-	srcIPStrings      []string
+	viaTags              []Tag
+	resolvedDsts         []netip.Prefix
+	hasAutoGroupInternet bool
+	internetProtocols    []ProtocolPort
+	srcIPStrings         []string
+}
+
+// resolveViaDestinations splits a via grant's destinations into the
+// flat list of IP prefixes they resolve to plus a flag for
+// autogroup:internet. Every alias kind goes through Alias.Resolve so
+// adding a new alias type to the policy parser does not silently
+// disappear from the via path. Non-IP alias kinds (tag, user, group,
+// wildcard) resolve to /32 host IPs that never overlap with subnet
+// route advertisements and therefore contribute nothing here.
+func resolveViaDestinations(
+	pol *Policy,
+	users types.Users,
+	nodes views.Slice[types.NodeView],
+	dsts Aliases,
+) ([]netip.Prefix, bool) {
+	var (
+		prefixes             []netip.Prefix
+		hasAutoGroupInternet bool
+	)
+
+	for _, d := range dsts {
+		if ag, ok := d.(*AutoGroup); ok && ag.Is(AutoGroupInternet) {
+			hasAutoGroupInternet = true
+
+			continue
+		}
+
+		ips, err := d.Resolve(pol, users, nodes)
+		if err != nil || ips == nil {
+			continue
+		}
+
+		prefixes = append(prefixes, ips.Prefixes()...)
+	}
+
+	return prefixes, hasAutoGroupInternet
 }
 
 // userNodeIndex maps user IDs to their untagged nodes. Built once per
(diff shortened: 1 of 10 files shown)


hscontrol/policy/v2/compiled.go
```diff
@@ -343,11 +343,29 @@ func (pol *Policy) compileOneViaGrant(
 	hasWildcard := sourcesHaveWildcard(grant.Sources)
 	hasDangerAll := sourcesHaveDangerAll(grant.Sources)
 
+	// Resolve *Host aliases to *Prefix here so compileViaForNode and
+	// ViaRoutesForPeer can run the overlap check against advertised
+	// subnet routes without re-entering the policy's Hosts map.
+	destinations := make(Aliases, 0, len(grant.Destinations))
+
+	for _, d := range grant.Destinations {
+		if h, ok := d.(*Host); ok {
```
kradalby: This feels like at trap where we might need to resolve every type of alias? sounds like there might be undiscovered edge cases? should this be a helper with more types?

hscontrol/policy/v2/compiled.go
```diff
@@ -770,7 +788,12 @@ func compileViaForNode(
 		switch d := dst.(type) {
 		case *Prefix:
 			dstPrefix := netip.Prefix(*d)
-			if slices.Contains(nodeSubnetRoutes, dstPrefix) {
+			// SaaS emits the literal grant dst in the filter rule
+			// whenever at least one advertised subnet route overlaps
+			// it (containment in either direction). See the captures
+			// under testdata/grant_results/via-grant-v47..v51 for the
```
kradalby: we dont reference testdata as it might always change. Explain why, not what. This applies to ALL comments.

hscontrol/policy/v2/policy.go
```diff
@@ -1028,6 +1034,22 @@ func (pm *PolicyManager) ViaRoutesForPeer(viewer, peer types.NodeView) types.Via
 				break
 			}
 		}
+
+		resolved := make(Aliases, 0, len(grant.Destinations))
+		for _, d := range grant.Destinations {
+			if h, ok := d.(*Host); ok {
```
kradalby: same comment as above.

juanfont (approved):

PR #3272, opened 2026-05-19
nix/release-issue: add helper for opening release-tracking issues

kradalby:
Adds `headscale-release-issue` to the `nix develop` shell. Given a stable version and an optional milestone number, it renders the embedded checklist template, resolves the milestone title via `gh api`, then creates the issue with the `no-stale` label, attaches the milestone, and locks the conversation so the checklist isn't derailed by drive-by comments.

Each release cycle needs the same tracking issue with the same surrounding configuration. Doing it by hand drops one of the steps (the lock, the label, the milestone) every time. `--dry-run` prints both `gh issue create` and `gh issue lock` commands without side effects so the rendered body is reviewable before opening anything.

The script lives inside the Nix package next to its template; the template path is injected at build time so the binary works regardless of cwd.

### DIFF
--- a/nix/release-issue/release-checklist.md
+++ b/nix/release-issue/release-checklist.md
@@ -1,0 +1,36 @@
+<!--
+Create this issue from inside `nix develop`:
+  headscale-release-issue --version X.Y.Z [--milestone N]
+Pass --dry-run to preview the rendered body without opening an issue.
+The script attaches the milestone (when given), adds the "no-stale-bot"
+label, and locks the issue for comments.
+-->
+
+# Release vX.Y.Z
+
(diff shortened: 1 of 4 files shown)


nix/release-issue/release-checklist.md
```diff
@@ -0,0 +1,35 @@
...
+- [ ] Review & tidy changelog
+
+## Per build (copy this block for each beta / rc / stable)
+
+### vX.Y.Z-beta.N
+
+- [ ] Tag & push
+- [ ] Verify CI / release artifacts
+- [ ] Announce on Discord
+- [ ] Comment on issues fixed in this build
```
nblock, line 28: Could you add another step to bump the version in `mkdocs.yml`?

Issue #3012, opened 2026-01-13
[Feature] Assign global DNS based on ip address or machine name


Firehawke:
### Use case

I'm using headscale for a non-commercial purpose, primarily for connection to my home network (and Pi-Hole).

I'm considering adding my brother to my headscale network, with his own Pi-Hole included. I'd prefer to be able to designate that machines on his side of the network prioritize his Pi-Hole before mine and vice versa for mine.

### Description

The ability to designate multiple global DNS servers based on where on the tailscale network you are, whether by tailscale IP address or machine name.

### Contribution

- [ ] I can write the design doc for this feature
- [ ] I can contribute this feature

### How can it be implemented?

```
nameservers:
  global:
    - 100.64.0.4
  split:
    - 100.64.0.4
      - 100.64.0.1
      - 100.64.0.2
      - 100.64.0.3
      - desktopa
      - laptopa
    - 100.64.0.100
      - 100.64.0.101
      - 100.64.0.102
      - 100.64.0.103
      - serverb
      - laptopb
```

Not a great design, but gets the idea across. The global would take precedence where a split entry is not present. With that in mind, I could have left the entire 100.64.0.1-3/desktopa/laptopa section out since they'd default to the global entry anyway.

IdanKoblik:
Hey, I thought of another approach called **DNS profiles**. Basically, as you mentioned, different IP addresses could have different nameservers. For example:
```yaml
# profiles: []
#   # Example: Client with IP 100.64.0.2 uses Cloudflare DNS
#   - ips:
#       - 100.64.0.2
#     nameservers:
#       - 1.1.1.1
#       - 1.0.0.1
#   # Example: Client with IP 100.64.0.3 uses Google DNS
#   - ips:
#       - 100.64.0.3
#     nameservers:
#       - 8.8.8.8
#       - 8.8.4.4
```

I’ve forked the repoistory and I’m actively working on implementing this feature. This way, each machine could prioritize its preferred DNS without interfering with others on the network.

I'm open to discussion on how to improve this approach or make it more flexible. any suggestions or feedback would be super helpful :)

ctschach:
I would love to see a similar feature that would allow to set DNS based on the user. So different users can have different DNS settings.   

My +1 for getting this implemented.

PR #3301, opened 2026-06-07
policy, types, state: add policy DNS profiles

jonscoresby:
This PR adds a per-identity DNS override mechanism to the policy file.

## Background

I've wanted this feature for a while, and it appears others have too:

- **#3012** — tracking issue: "Assign global DNS based on IP address or machine name."
- **#3127** — alternative draft PR by IdanKoblik.

I've implemented and run variations of this on my own tailnet for a couple of months. I didn't realize #3127 existed until I was getting ready to submit my own PR. It's encouraging to find we'd arrived at similar ideas independently.

#3127 proposes a minimal IP-based implementation: a `dns.profiles` block in `headscale.yaml` keyed by tailnet IP, replacing `Resolvers` on the wire for matched nodes. Narrow scope, ~200 lines added. This PR takes a broader shape: identity-based instead of IP-based.

## Description

The policy file gains a top-level `dns:` block: an ordered list of DNS
profiles that override the corresponding fields in `headscale.yaml`'s
tailnet-wide `dns:` block for matched nodes. Each profile carries an
assignment list (`groups`, `users`, or `tags`) and an optional set of
body fields (`nameservers`, `overrideLocalDNS`, `split`,
`searchDomains`).

When a node matches a profile, the profile's set fields replace the
corresponding fields from `headscale.yaml`'s tailnet-wide `dns:` block;
unset fields inherit from the yaml. A node that matches no profile
receives the yaml config unchanged. Profile changes are hot-reloaded
with the policy file.

Matching is by tier — **tag > user > group** — and by profile list
order within each tier. Tagged nodes consult only the tag tier.

```hujson
{
  "groups": {
    "group:admin": ["alice@"]
  },
  "dns": [
    {
      // Admins override their primary resolver.
      "nameservers": ["192.168.4.2"],
      "overrideLocalDNS": true,
      "groups": ["group:admin"]
    }
  ]
}
```

Result on the wire:

| User | Resolvers | from |
|---|---|---|
| `alice@` (admin) | `[192.168.4.2]` | profile match |
| anyone else | (base) | unchanged from `headscale.yaml` |

### DNS Profiles in policy

Putting DNS in the policy file felt natural for a couple of reasons:

- DNS config can be assigned per node, and the ACL identity definitions (groups, users, tags)
  in the policy file are a great fit for that.
- There's no reason DNS config can't be reassigned at runtime — and the policy file already
  has the hot-reload plumbing.

Putting it in `headscale.yaml` instead would require restarting the server to change DNS for
even a single node. It would also create a coupling problem: the base config would need to
reference identities defined in the policy file, but those identities change on policy
reload, which could leave dangling references in the yaml.

I initially tried to move DNS out of the yaml entirely (with MagicDNS settings becoming their
own block), which I think is cleaner — it keeps all DNS config in one spot, and the first
profile can serve as the default. But the backwards-compatibility footguns were a nightmare.
This diff is much simpler.

### Pointer-typed body fields

The body fields (`Nameservers`, `Split`, `SearchDomains`,
`OverrideLocalDNS`) are pointer-typed so the schema can distinguish
three states:

- **Absent** (field omitted in JSON) — inherits from base
- **Present and non-empty** — replaces base
- **Present but empty** (e.g. `"searchDomains": []`) — explicitly clears
  the inherited value

### `Nameservers` + `OverrideLocalDNS` matrix

The wire protocol has two slots: `Resolvers` (primary) and
`FallbackResolvers` (only when OS has no DNS). The base has nameservers
in one or the other depending on `OverrideLocalDNS`. Profiles set or
omit `Nameservers` and `OverrideLocalDNS` independently:

| Profile sets... | Effect |
|---|---|
| Both `Nameservers` and `OverrideLocalDNS` | New nameservers go to whichever wire field the explicit `OverrideLocalDNS` selects; the other wire field is cleared |
| Only `Nameservers` (not `OverrideLocalDNS`) | New nameservers go to whichever wire field base was using |
| Only `OverrideLocalDNS` (not `Nameservers`) | The inherited nameservers move from one wire field to the other per the explicit value |
| Neither | Both wire fields inherit from base unchanged |

Unit tests in `TestApplyDNSProfileFieldHandoff` cover all four cells.

### Why threading `baseDomain` explicitly

`tailcfg.DNSConfig.Domains` is a flat list — by convention `Domains[0]`
is the tailnet `base_domain` followed by search domains, but the type
itself doesn't distinguish them. When a profile replaces
`SearchDomains`, we want to preserve `base_domain` and replace the
rest. If we read `Domains[0]` heuristically that works in the common
case, but breaks when `base_domain` is unset (e.g. MagicDNS off + search
domains set in yaml) — `Domains[0]` is then a search domain and
preserving it would be wrong.

So `baseDomain string` is threaded through `NodeDNSConfig` →
`applyDNSProfile` explicitly. The state layer passes
`s.cfg.BaseDomain`; the mapper doesn't need to know. The
`search-domains-no-base-domain-fully-replaces-domains` subtest of
`TestApplyDNSProfileFieldHandoff` covers the edge case.

### `SetUsers` force-change carve-out

A policy can list a username in a Groups definition before that user
exists in the DB — validation doesn't reject it. So when the user is
later created (or renamed / deleted), the DNS profile resolution for
their node shifts. `SetUsers` normally returns `changed=true` only when
a tracked hash moves (filter / tag-owner / auto-approver / SSH), and
none of those catch DNS resolution. Without a carve-out the new DNS
mapping wouldn't be broadcast until something else triggered a map
response.

Mirrors the existing SSH carve-out:

```go
if pm.pol != nil && len(pm.pol.DNS) > 0 {
    return true, nil
}
```

`TestSetUsersForcesChangeWhenDNSProfilesExist` is the regression guard.

### Validation choices

The validator rejects:

- **Profiles with no assignment list.** A profile with no `groups` /
  `users` / `tags` would never match — it's almost certainly a
  misconfiguration. The opposite — a profile *with* an assignment
  list but no body fields — is allowed; it's the exemption pattern
  documented in `docs/ref/policy.md`.
- **Duplicate principals across profiles.** A group / user / tag may
  appear in at most one profile's assignment list. Each principal
  belongs to exactly one profile; list order only resolves ties within
  a tier, never which profile a principal is assigned to.
- **Undefined references.** Group references must exist in the policy's
  top-level `groups` map; tag references must be in `tagOwners`;
  usernames must contain `@`. Same conventions as ACL / SSH / TagOwners
  validation.
- **Ambiguous usernames** (multi-user resolution). `validateUserReferences`
  now walks `pol.DNS` alongside `Groups`, `TagOwners`, `AutoApprovers`,
  `ACLs`, and `SSH`. Without this, an ambiguous username in a DNS
  profile's `Users` list would silently fail to match at lookup time.

---

## Backwards compatibility

The yaml `dns:` block is unchanged. Operators who don't use
`policy.dns` see identical behavior to today — no deprecation warnings,
no migration path, no cross-file checks.

DB-mode policies are unaffected. Both file and db modes converge on
`hsdb.PolicyBytes` → `polMan.SetPolicy(pol)`; the DNS feature lives
entirely downstream of that fork.

---

## Maintenance surface

The feature is bounded and localized: `policy/v2/policy.go` matcher,
`policy/v2/types.go` schema + validation, a thin passthrough in
`state.go`. No new external dependencies, no new long-running
goroutines, no new lock layers. The tests I added should catch
regressions.

I'll engage with review comments and plan to run headscale with this
feature for the foreseeable future. I won't be offended if the PR
isn't accepted — I'll be maintaining this fork for myself either way,
so I figured I'd throw it out there for anyone else that might
benefit. No hard maintenance commitments, but if anything ever breaks
my setup, I'll be fixing it.

---

### Unit (`hscontrol/policy/v2/dns_test.go`, 11 functions)

- Tier precedence (tag > user > group)
- List order within tier (group + user)
- Multi-tag profile list order — regression guard for an earlier bug
  where iterating node tags first picked the wrong profile
- No match returns base unchanged
- Empty DNS policy returns base
- Nil base returns nil
- Split + SearchDomains override applied to base
- `applyDNSProfile` field handoff matrix:
  - `Nameservers` + `OverrideLocalDNS` four-cell matrix
  - Re-route on `OverrideLocalDNS`-only profile (both directions)
  - Present-empty `Nameservers` clears both wire fields
  - Absent fields inherit base
  - Split replace / inherit / present-empty-clear
  - SearchDomains with `base_domain` / without `base_domain` /
    present-empty (with and without `base_domain`)
- Validation: profile with no assignment list rejected
- Validation: duplicate principal across profiles rejected (groups,
  users, tags)
- Validation: undefined group / tag / malformed username rejected
- `TestSetUsersForcesChangeWhenDNSProfilesExist` — the SetUsers
  carve-out

### Existing validation test parity

`TestValidateUserReferences_AllSites` (`policy_test.go`) gains a `dns`
case for symmetry with `groups`, `tagOwners`, `autoApprovers.routes`,
`autoApprovers.exitNode`, `acls.src`, `acls.dst`, `ssh.src`, `ssh.dst`.

### Mapper

`hscontrol/mapper/mapper_test.go` updated for the
`generateDNSConfig` signature change. Tests now construct
`*tailcfg.DNSConfig` directly (no test stub for the dropped
`dnsConfigStater` interface).

### Integration (`integration/dns_test.go`, `TestPolicyDNSProfiles`)

Three phases against real Tailscale clients on a real headscale
server:

1. **No policy.dns overrides** → neither admin nor guest has the
   override resolver in `Resolvers` or `FallbackResolvers`.
2. **Hot-reload via `SetPolicy`** to add a `group:admin` profile with
   `Nameservers` + `OverrideLocalDNS` + `Split`.
3. **Post-reload** — admin gets the override `Resolvers` AND the split
   `Routes`; guest stays clean on `Resolvers`, `FallbackResolvers`, and
   `Routes`.

Verified passing locally.

---

## Checklist (from PR template)

- [x] raised a GitHub issue or discussed it on the projects chat beforehand — #3012
- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

---

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -282,6 +282,7 @@ jobs:
           - TestDERPVerifyEndpoint
           - TestResolveMagicDNS
           - TestResolveMagicDNSExtraRecordsPath
+          - TestPolicyDNSProfiles
           - TestDERPServerScenario
           - TestDERPServerWebsocketScenario
           - TestPingAllByIP
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ to understand how the packet filter should be generated. We discovered a few dif
 overall our implementation was very close.
 [#3036](https://github.com/juanfont/headscale/pull/3036)
 
+### DNS profiles (per-tag / per-user / per-group) [#3012](https://github.com/juanfont/headscale/issues/3012)
+
+The policy file gains a top-level `dns:` block: an ordered list of DNS
+profiles that override fields of the tailnet-wide `dns:` config in
+`headscale.yaml` for matched nodes. Each profile carries an assignment
+list (`groups`, `users`, or `tags`); fields set on the profile replace
+the base, unset fields inherit. Nodes matching no profile receive the
+base unchanged. Matching is by tier (tag > user > group) and by
+profile list order within each tier. Hot-reloaded with the policy file.
+
 ### SSH check action
 
 SSH rules with `"action": "check"` are now supported. When a client initiates a SSH connection to a node
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -318,6 +318,12 @@ policy:
 #
 # If you want stop Headscale from managing the DNS configuration
 # all the fields under `dns` should be set to empty values.
+#
+# For per-identity DNS overrides (different resolvers / split DNS /
+# search domains for specific groups, users, or tagged nodes), use the
+# policy file's top-level `dns:` block — see
+# https://headscale.net/stable/ref/policy/#dns-profiles
+
 dns:
   # Whether to use MagicDNS
   magic_dns: true
--- a/docs/ref/dns.md
+++ b/docs/ref/dns.md
@@ -3,6 +3,10 @@
 Headscale supports [most DNS features](../about/features.md) from Tailscale. DNS related settings can be configured
 within the `dns` section of the [configuration file](configuration.md).
 
+For **per-identity DNS overrides** (different resolvers, override flags,
+or split-DNS routes for specific groups, users, or tagged nodes), see
+[DNS profiles in the policy file](policy.md#dns-profiles).
+
 ## Setting extra DNS records
 
 Headscale allows to set extra DNS records which are made available via
--- a/docs/ref/policy.md
+++ b/docs/ref/policy.md
@@ -226,6 +226,71 @@ configuration and attributes. At least the following node attributes are current
 }
 ```
 
+## DNS profiles
+
+The top-level `dns:` block in the policy file is an ordered list of
+DNS profiles that override fields of the tailnet-wide `dns:` config in
+`headscale.yaml`. Each profile carries an assignment list (`groups`,
+`users`, or `tags`) and an optional set of DNS fields; a matched node
+gets the profile's set fields, with unset fields inheriting from the
+base. A node matching no profile receives the base unchanged.
+
+### Profile shape
+
+| Field | Type | Effect |
+|---|---|---|
+| `nameservers` | `[]string` | Replaces the base nameservers (routed by `overrideLocalDNS` to either `Resolvers` or `FallbackResolvers`). |
+| `overrideLocalDNS` | `bool` | Sends nameservers to `Resolvers` (`true`) or `FallbackResolvers` (`false`). Set without `nameservers` to re-route the inherited nameservers between the two wire fields. |
+| `split` | `map[string][]string` | Per-domain restricted resolvers (`Routes`); replaces base `Routes`. |
+| `searchDomains` | `[]string` | DNS search suffixes; appended after the `base_domain` in `Domains`. |
+| `groups` | `[]Group` | Assignment list — matches nodes whose user is a member of any listed group (group tier). |
+| `users` | `[]Username` | Assignment list — matches nodes whose user is listed (user tier). |
+| `tags` | `[]Tag` | Assignment list — matches tagged nodes (tag tier). |
+
+Every profile must have at least one of `groups` / `users` / `tags`.
+
+### Resolution model
+
+For each node, the lookup walks tiers in this order and returns the
+first profile whose assignment list matches, with profile list order
+breaking ties within a tier:
+
+1. **Tag tier** — any of the node's tags appears in the profile's
+   `tags`. Tagged nodes consult ONLY the tag tier.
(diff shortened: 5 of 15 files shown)


liorsl:
Doesnt it make more sense to specify the dns settings in the configuration as a named profile, and reference it within the acls file as a grant? This is what I did, based on #3127

jonscoresby:
If it's in settings, then any profile additions or modifications require server restart which may not be a big deal, but I think optimally all the dns settings would be in the policy doc, but that creates a backwards compatibility mess.

PR #3304, opened 2026-06-08
all: more fixes

kradalby:


hscontrol/mapper/mapper.go
```diff
@@ -440,6 +441,49 @@ func (m *mapper) buildFromChange(
 	return builder.Build()
 }
 
+// filterVisiblePeerPatches drops peer-change patches whose target peer the
+// recipient cannot see under the ACL policy, mirroring the [policy.ReduceNodes]
+// visibility filter [MapResponseBuilder.buildTailPeers] applies to full peer
+// objects. Without it, online/offline, endpoint, and key-expiry patches
+// disclose the existence, presence, and addresses of peers the recipient's
+// policy forbids it from accessing.
+func (m *mapper) filterVisiblePeerPatches(
```
kradalby: An interesting and good find, is there a way we could maintain this code in the same place as the other filter / reduce? Just to make sure we avoid the case where one path is updated and another is not.

hscontrol/mapper/mapper.go
```diff
@@ -484,6 +485,37 @@ func (m *mapper) filterVisiblePeerPatches(
...
 }
 
+// filterVisibleNodes restricts a peer slice to the nodes the recipient can see
+// under the ACL policy, mirroring the policy.ReduceNodes filter that
+// buildTailPeers applies to full peer objects. It guards UserProfiles on the
+// incremental PeersChanged path: unlike the full-map path (whose ListPeers
+// returns the BuildPeerMap-filtered set), that path receives an unfiltered
+// node slice and would otherwise leak the identities of users whose nodes the
+// recipient cannot access.
+func (m *mapper) filterVisibleNodes(
```
kradalby, line 523: Same questions, it feels like we are adding more and more paths for reducing / limited visibility. Its good we are adding it, but having a single place would be cleaner and easier to reason about.

juanfont (approved):

PR #3308, opened 2026-06-10
authflow flakes

kradalby:
In-memory only, like SessionEpoch. Generated clone/view regenerated.


hscontrol/types/node.go
```diff
@@ -181,6 +181,14 @@ type Node struct {
 	// online. Written by the HA prober. Runtime-only.
 	Unhealthy bool `gorm:"-"`
 
+	// ActiveSessions counts live poll sessions for this node.
+	// [State.Connect] increments it and every session release
+	// ([State.Disconnect]) decrements it, so the node goes offline
+	// exactly when its last session ends — regardless of the order in
+	// which overlapping sessions' cleanups run. Never persisted, like
+	// SessionEpoch.
+	ActiveSessions int `gorm:"-"`
```
kradalby, line 190: Should this be an atomic int?

hscontrol/state/connect_test.go
```diff
@@ -207,10 +202,75 @@ func TestConnectDisconnectSubnetRouterEmitsPolicyChangeNotFull(t *testing.T) {
...
+
+// drainSessions releases poll sessions on nodeID until the node goes
+// offline, returning the changes from the final release (the one that
+// emits the offline notifications). Fails the test if the node is
+// still online after releasing as many sessions as Connect calls could
+// plausibly have acquired.
+func drainSessions(t *testing.T, s *State, nodeID types.NodeID, epoch uint64) []change.Change {
+	t.Helper()
+
+	const maxSessions = 16
```
kradalby, line 222: Is this arbitrarily chosen? comment.

hscontrol/state/state.go
```diff
@@ -611,6 +611,7 @@ func (s *State) Connect(id types.NodeID) ([]change.Change, uint64) {
 	node, ok := s.nodeStore.UpdateNode(id, func(n *types.Node) {
 		n.SessionEpoch++
```
kradalby, line 612: Does this mean SessionEpoch is no longer used? do we need both?

juanfont (approved):

PR #3311, opened 2026-06-13
ci: add nix flake checks for build and test

kradalby:
Adds two inline Nix flake checks and matching CI jobs:

- `checks.build` — compiles the headscale module via buildGoModule over a minimal, vendor-free source fileset (Go sources + embeds + testdata).
- `checks.gotest` — runs the pure unit-test subset with `go test`, excluding the `integration` package (Docker; self-skips outside a container) and `hscontrol/servertest` (binds loopback sockets / race-stress harness), and skipping `TestPostgres*` (needs a postgres server; the SQLite equivalents still run).

Both checks are Linux-gated (`lib.optionalAttrs stdenv.isLinux`) and validated in CI on `x86_64-linux`. The derivations are inlined directly in `flake.nix` with no extra dependencies.

golangci-lint (`--new-from-rev`) and the Docker integration tests are intentionally left unchanged.

### DIFF
--- a/.github/workflows/nix-checks.yml
+++ b/.github/workflows/nix-checks.yml
@@ -1,0 +1,76 @@
+name: Nix Flake Checks
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-$${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


kradalby:
Folded into #3310. After rebasing onto main (which removed the proto/gRPC stack), the flake checks and nix-checks.yml now ride on top of the hestia migration in that PR — the fileset's stale gen/openapiv2 embed was dropped since that path no longer exists. Closing to keep it as a single CI change.

PR #3314, opened 2026-06-15
state: re-register converted tagged nodes with reused key

### LINKED ISSUE #3312: [Bug] Containerized tailscaled is logged out when reusing the same TS_AUTHKEY

### Current Behavior

A containerized tailscaled that was registered with a user PAK and later converted to a tagged node is logged-out upon container restart when the same `TS_AUTHKEY` is reused.

Together with @Nathanael-Mtd, we were able to identify this issue on containerized nodes that were tagged on older versions but hit the same code path when upgraded from 0.28.0 to 0.29.0-beta.3.

### Expected Behavior

To ignore the same `TS_AUTHKEY` sent again for nodes that are already registered.

This works in 0.28 as the node still has the `user_id` reference in the database. The upgrade to 0.29.0-beta.3 sets `node.user_id=NULL` and this causes a state mismatch during login between the PAK (still owned by the user) and the node which is now tagged).

### Steps To Reproduce

There are multiple ways to trigger this bug; one of the simpler variants:

1. Start Headscale 0.29.0-beta.3
1. Create user: `headscale user create alice`
1. Create PAK for user alice: `headscale preauthkeys create -u 1`
1. Register node with PAK:
    ```bash
    podman run --rm \
      -e TS_AUTHKEY=hskey-auth-... \
      -e TS_EXTRA_ARGS=--login-server=https://headscale.example.com \
      -e TS_STATE_DIR=/var/lib/tailscale \
      -v /tmp/ts-state:/var/lib/tailscale \
      docker.io/tailscale/tailscale
    ```
- Convert to tagged node on Headscale: `headscale nodes tag -t tag:foo -i 1`
- Restart container using the same podman command as above

This results in the following log lines on the containerized tailscaled:

```
2026/06/14 05:21:49 control: RegisterReq: got response; nodeKeyExpired=false, machineAuthorized=false; authURL=false
2026/06/14 05:21:49 Received error: authkey already used
2026/06/14 05:21:49 health(warnable=login-state): error: You are logged out. The last login error was: authkey already used
backend error: authkey already used
```

The actual log lines might differ a bit depending on how this issue is reproduced. For example if the auth key is expired:

```
2026/06/14 05:38:22 control: RegisterReq: got response; nodeKeyExpired=false, machineAuthorized=false; authURL=false
2026/06/14 05:38:22 Received error: authkey expired
2026/06/14 05:38:22 health(warnable=login-state): error: You are logged out. The last login error was: authkey expired
backend error: authkey expired
``` 

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.29.0-beta.3
- Tailscale version: 1.98 (containerized)
```

### Runtime environment

### Debug information

This policy is required to perform the conversion to a tagged node:

```json
{
  "tagOwners": {
    "tag:bar": ["alice@"],
    "tag:foo": ["alice@"]
  },
  "acls": [
    {
      "action": "accept",
      "src": ["*"],
      "dst": ["*:*"]
    }
  ]
}
```

---

PAK in the DB (after update to 0.29.0-beta.3):

```
sqlite> select * from pre_auth_keys where id=1;
        id = 1
       key = 
    prefix = ppBQfc61kStX
      hash = $2a$10$QoW/P/0/m26iHLp4xeTXk.JCZKE7k5o7EHUqcYJwNk4BgRdGuDr7.
   user_id = 1
  reusable = 0
 ephemeral = 0
      used = 1
      tags = []
created_at = 2026-06-14 05:21:06.639613424+00:00
expiration = 2026-06-14 07:38:14.350474576+02:00
```

Node in the DB (after update to 0.29.0-beta.3):

```
sqlite> select * from nodes;
             id = 1
    machine_key = mkey:8bbce1b4d4a3792b744f4046b1f03cf236577b16b5bac5793f005cc65c901f1f
       node_key = nodekey:46d7a0374e63f747f140c656923c8de5529f1bbcf1968ec53b1c0c9ae306b15b
      disco_key = discokey:7842848cedf6acf584e34e0a003ea42246cee2dba088ae7793dbd59889f13a4e
      endpoints = [..]
      host_info = {"stripped"}
           ipv4 = 100.64.0.1
           ipv6 = fd7a:115c:a1e0::1
       hostname = cd9dce670fde
     given_name = cd9dce670fde
        user_id = 
register_method = authkey
           tags = ["tag:foo"]
    auth_key_id = 1
         expiry = 
      last_seen = 2026-06-14 07:59:15.434673756+02:00
approved_routes = 
     created_at = 2026-06-14 07:21:16.302874795+02:00
     updated_at = 2026-06-14 07:59:15.434904099+02:00
     deleted_at = 
```

---

Two possible workarounds:

- Restart the container without `-e TS_AUTHKEY=...`. The client uses its persistent state to login
- Create a tagged PAK (actual tags are not relevant) and use it as new value for `-e TS_AUTHKEY=...`: `headscale preauthkeys create --tags tag:foo`

kradalby:
Fixes https://github.com/juanfont/headscale/issues/3312

### DIFF
--- a/hscontrol/state/node_store.go
+++ b/hscontrol/state/node_store.go
@@ -804,12 +804,15 @@ func (s *NodeStore) GetNodeByMachineKey(machineKey key.MachinePublic, userID typ
 	return types.NodeView{}, false
 }
 
-// GetNodeByMachineKeyAnyUser returns the first node with the given machine key,
+// GetNodeByMachineKeyAnyUser returns a node with the given machine key,
 // regardless of which user it belongs to. This is useful for scenarios like
 // transferring a node to a different user when re-authenticating with a
 // different user's auth key.
-// If multiple nodes exist with the same machine key (different users), the
-// first one found is returned (order is not guaranteed).
+//
+// When more than one node shares the machine key (e.g. a tagged node and a
+// user-owned node), the choice is deterministic: the tagged node is preferred,
+// otherwise the node with the lowest ID. A stable pick keeps re-auth branch
+// selection (convert vs create) from depending on map iteration order.
 func (s *NodeStore) GetNodeByMachineKeyAnyUser(machineKey key.MachinePublic) (types.NodeView, bool) {
 	timer := prometheus.NewTimer(nodeStoreOperationDuration.WithLabelValues("get_by_machine_key_any_user"))
 	defer timer.ObserveDuration()
@@ -817,14 +820,20 @@ func (s *NodeStore) GetNodeByMachineKeyAnyUser(machineKey key.MachinePublic) (ty
 	nodeStoreOperations.WithLabelValues("get_by_machine_key_any_user").Inc()
 
 	snapshot := s.data.Load()
+
+	var best types.NodeView
+
 	if userMap, exists := snapshot.nodesByMachineKey[machineKey]; exists {
-		// Return the first node found (order not guaranteed due to map iteration)
 		for _, node := range userMap {
-			return node, true
+			if !best.Valid() ||
+				(node.IsTagged() && !best.IsTagged()) ||
+				(node.IsTagged() == best.IsTagged() && node.ID() < best.ID()) {
+				best = node
+			}
 		}
 	}
 
-	return types.NodeView{}, false
+	return best, best.Valid()
 }
 
 // DebugString returns debug information about the [NodeStore].
--- a/hscontrol/state/node_store.go
+++ b/hscontrol/state/node_store.go
@@ -787,53 +787,27 @@ func (s *NodeStore) GetNodeByNodeKey(nodeKey key.NodePublic) (types.NodeView, bo
 	return nodeView, exists
 }
 
-// GetNodeByMachineKey returns a node by its machine key and user ID. The bool indicates if the node exists.
-func (s *NodeStore) GetNodeByMachineKey(machineKey key.MachinePublic, userID types.UserID) (types.NodeView, bool) {
-	timer := prometheus.NewTimer(nodeStoreOperationDuration.WithLabelValues("get_by_machine_key"))
-	defer timer.ObserveDuration()
-
-	nodeStoreOperations.WithLabelValues("get_by_machine_key").Inc()
-
-	snapshot := s.data.Load()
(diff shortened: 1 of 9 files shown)


hscontrol/state/node_store.go
```diff
@@ -804,27 +804,36 @@ func (s *NodeStore) GetNodeByMachineKey(machineKey key.MachinePublic, userID typ
...
 // transferring a node to a different user when re-authenticating with a
 // different user's auth key.
-// If multiple nodes exist with the same machine key (different users), the
-// first one found is returned (order is not guaranteed).
+//
+// When more than one node shares the machine key (e.g. a tagged node and a
+// user-owned node), the choice is deterministic: the tagged node is preferred,
+// otherwise the node with the lowest ID. A stable pick keeps re-auth branch
+// selection (convert vs create) from depending on map iteration order.
 func (s *NodeStore) GetNodeByMachineKeyAnyUser(machineKey key.MachinePublic) (types.NodeView, bool) {
```
kradalby: Should machine key be unique? Maybe this method is creating more trouble than it should?

hscontrol/state/state.go
```diff
@@ -179,6 +179,23 @@ type State struct {
 	// persistNodeToDB so the database row always converges on [NodeStore]
 	// rather than being clobbered by a stale caller snapshot.
 	persistMu sync.Mutex
+
+	// registerLocks serialises registration per machine key so concurrent
+	// registrations of the same machine resolve to a single node instead of
+	// racing the find-then-create section and each creating their own.
+	// ponytail: entries are never pruned; bounded by distinct machine keys
+	// seen, add cleanup on node delete only if it ever matters.
+	registerLocks sync.Map // key.MachinePublic -> *sync.Mutex
```
kradalby: always use the typed xsync map.

hscontrol/state/node_store.go
kradalby: Or should they always return all in a stable order? then the caller has to decide?

kradalby: Confirmed it should not be unique.

juanfont (approved):

PR #3317, opened 2026-06-16
all: small fixups

kradalby:


kradalby:
Folded into #3318 — the four commits (NodeKey-rotation regression test, PKCE validation, verify Content-Type, server_url port) now ride on that branch.

PR #3325, opened 2026-06-18
db: preserve user_id on untagged nodes with tags='null'

### LINKED ISSUE #3323: [Bug] Nodes with tags='null' lose their assigned user upon upgrade to 0.29

### Current Behavior

The assigned user is lost during the upgrade from 0.28 to 0.29 for all nodes that have the string literal `'null'` for `nodes.tags` in their database. Such nodes end up with `node.user_id=null` (real NULL); no tags are involved.

This only shows up for relatively old nodes that haven't been connected for a long time. Its probably some issue with a rather old headscale version.

### Expected Behavior

To keep their user assigned during the upgrade.

### Steps To Reproduce

1. Have a 0.28 database with at least one node `nodes.tags='null'`
2. Upgrade to 0.29.0
3. Observe that the node has no user assigned.

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.29.0
- Tailscale version: 1.98.x
```

### Runtime environment

### Debug information

sqlite before the update:

```
sqlite> .nullvalue "!!THE-REAL-NULL!!"
sqlite> select id, user_id, tags, deleted_at from nodes where tags='null' LIMIT 1;
        id = 27
   user_id = 6
      tags = null
deleted_at = !!THE-REAL-NULL!!
```

sqlite after the update:

```
sqlite> .nullvalue "!!THE-REAL-NULL!!"
sqlite> select id, user_id, tags, deleted_at from nodes where tags='null' LIMIT 1;
        id = 27
   user_id = !!THE-REAL-NULL!!
      tags = null
deleted_at = !!THE-REAL-NULL!!
```

---

The workaround for this bug is to execute the following SQL statement *before* the upgrade: `UPDATE nodes SET tags=null WHERE tags='null'` or to re-assign the `nodes.user_id` after the update.

kradalby:
A nil tags slice marshals to the JSON literal `null`, so pre-0.29 databases hold untagged nodes whose `tags` column is the string `'null'`. The `clear-tagged-node-user-id` migration treated anything matching `tags != '[]' AND tags != ''` as tagged, so it cleared `user_id` on those untagged nodes and detached them from their owner on upgrade.

Excluding `'null'` from the predicate keeps fresh 0.28 upgrades safe. Databases already on 0.29.0 took the damage, so a follow-up migration re-derives `user_id` from each orphaned node's pre-auth key. Best-effort: CLI/OIDC nodes have no pre-auth key and need manual reassignment.

Fixes #3323

### DIFF
--- a/.github/workflows/nix-module-test.yml
+++ b/.github/workflows/nix-module-test.yml
@@ -52,4 +52,4 @@ jobs:
         if: steps.changed-files.outputs.nix == 'true' || steps.changed-files.outputs.go == 'true'
         run: |
           echo "Running NixOS module integration test..."
-          nix build .#checks.x86_64-linux.headscale -L
+          nix build .#checks.x86_64-linux.headscale -L --fallback
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 **Minimum supported Tailscale client version: v1.xx.0**
 
+## 0.29.1 (2026-06-18)
+
+**Minimum supported Tailscale client version: v1.80.0**
+
+### Changes
+
+- Fix nodes with `tags='null'` losing their assigned user on upgrade [#3325](https://github.com/juanfont/headscale/pull/3325)
+
 ## 0.29.0 (2026-06-17)
 
 **Minimum supported Tailscale client version: v1.80.0**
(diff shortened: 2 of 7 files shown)


nblock (approved):
Tested the update from 0.28 and the repair step (with a "faked" `auth_key_id`).

juanfont (approved):

PR #3329, opened 2026-06-19
api/v2: add Tailscale-compatible API with auth keys endpoint

kradalby:


juanfont (approved):

hscontrol/api/v2/api.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+}
+
+// Config returns the Huma configuration shared by the production API and tests:
+// the v2 OpenAPI/docs paths, the basic+bearer security schemes, and the
+// Tailscale error transform. Suppressing SchemasPath/CreateHooks keeps "$schema"
+// out of the emitted bodies, matching the Tailscale wire contract.
+func Config() huma.Config {
+	config := huma.DefaultConfig("Headscale API", "v2")
+	config.Info.Description = "Tailscale-compatible Headscale API."
```
kradalby: It is not the compatible API, its headscale's v2, but some endpoints are compatible with Tailscale.

hscontrol/api/v2/api.go
```diff
@@ -0,0 +1,271 @@
...
+		return huma.Error404NotFound("tailnet not found")
+	}
+
+	return nil
+}
+
+// Scope is a capability an operation requires and an API key grants. Today
+// every valid key is granted ScopeAll; the seam exists so OAuth tokens can
+// later grant a real, narrower set without reworking the operations.
+type Scope string
```
kradalby: we should model this in full after the tailscale api spec.

hscontrol/api/v2/keys.go
```diff
@@ -0,0 +1,413 @@
...
+	"github.com/juanfont/headscale/hscontrol/types"
+)
+
+func init() {
+	registrations = append(registrations, registerKeys)
+}
+
+// defaultExpirySeconds is the auth-key lifetime when the request omits
+// expirySeconds: 90 days, matching Tailscale and the Terraform provider default.
+const defaultExpirySeconds = 7776000
```
kradalby: write it as readable time math time.Hours * this and that

hscontrol/api/v2/keys.go
```diff
@@ -0,0 +1,413 @@
...
+// findKeyByID looks up a stored pre-auth key by its (stringified) id. There is
+// no by-id getter on State — GetPreAuthKey looks up by secret — so filter the
+// list, which the key set is small enough for.
+func findKeyByID(b Backend, rawID string) (*types.PreAuthKey, error) {
+	id, err := parseKeyID(rawID)
+	if err != nil {
+		return nil, err
+	}
+
+	keys, err := b.State.ListPreAuthKeys()
```
kradalby: this should just be a direct query, not something that needs to loop through a list. And should be part of guidance, to adopt the lookup methods instead of making too many helpers.

hscontrol/api/v2/keys.go
```diff
@@ -0,0 +1,413 @@
...
+	return tags
+}
+
+func timeOrZero(t *time.Time) time.Time {
+	if t == nil {
+		return time.Time{}
+	}
+
+	return *t
+}
```
kradalby: make sure we reuse these across all api v2 as helpers

hscontrol/api/v2/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+This package is the **Tailscale-compatible** Headscale API, served at `/api/v2`.
+It exists so the Tailscale ecosystem that cannot talk to Headscale today works:
+the [Terraform/OpenTofu provider], [tscli], and the official [Go client]
+(`tailscale.com/client/tailscale/v2`).
+
+It is **not** a port of the whole Tailscale API. Endpoints are added one at a
+time, only as we need them, reusing Tailscale's wire shapes so the existing
+tooling drives Headscale unchanged. The headscale-native admin API stays at
+`/api/v1` (`hscontrol/api/v1`).
```
kradalby: The package is Headscale's v2 API, but some endpoints are implementing Tailscale's API endpoints. The section here is to implement all the tests and conventions for the endpoints that are PORTED from Tailscale.

hscontrol/api/v2/README.md
```diff
@@ -0,0 +1,76 @@
...
+- Operations derived from Tailscale carry the `Tailscale compat` tag.
+- The `{tailnet}` path segment must be `-` (the single Headscale tailnet);
+  anything else is `404`. See `requireDefaultTailnet`.
+- Errors use **Tailscale's** body (`{"message","data","status"}`), installed as
+  a per-API transform (`tailscaleErrorTransformer` in `errors.go`). A future
+  headscale-native v2 operation would keep Huma's RFC 9457 problem+json.
+- Auth accepts the API key as **HTTP Basic** (key as username — what the SDK
+  sends) or **Bearer**. See `authMiddleware`.
+- Scopes are scaffolded: each operation declares a required scope; every key is
+  granted `all` today. OAuth fills in a real scope set later.
```
kradalby: Implement with the same scope as tailscale for every action, but not enforce anything.

hscontrol/api/v2/README.md
```diff
@@ -0,0 +1,76 @@
...
+6. **Roundtrip the real clients.** Add a `t.Run` subtest to `TestAPIv2`
+   (`hscontrol/servertest/apiv2_test.go`) for each of the Go client, tscli, and
+   OpenTofu — full create→read→list→delete against one shared server on a real
+   loopback port (`servertest.WithRealListener`). tscli and tofu come from the
+   nix dev shell; a missing binary fails the test. _Acceptance: `nix develop -c
+go test ./hscontrol/servertest/ -run TestAPIv2` is green._
+
+7. **Update the CLI** only if the v2 operation fully replaces a v1 one. The keys
+   resource does **not** (it is tagged/user-owned auth keys, no expire verb), so
+   `headscale preauthkeys` stays on v1.
```
kradalby: Im surprised that the tailscale api does not support something similar to this.

hscontrol/api/v2/api.go
```diff
@@ -0,0 +1,271 @@
...
+
+	op.Metadata[scopeMetaKey] = s
+
+	return op
+}
+
+type scopeSet map[Scope]struct{}
+
+func (s scopeSet) has(required Scope) bool {
+	if _, all := s[ScopeAll]; all {
```
kradalby: We should just accept everything now with a todo

hscontrol/api/v2/devices.go
```diff
@@ -0,0 +1,497 @@
...
+
+	slices.SortFunc(approved, netip.Prefix.Compare)
+
+	return slices.Compact(approved), nil
+}
+
+// deviceFromView maps a Headscale node onto the Tailscale Device. allFields
+// gates the route slices, which Tailscale only returns for fields=all.
+func deviceFromView(view types.NodeView, allFields bool) Device {
+	node := view.AsStruct()
```
kradalby: Why do we have to convert this? Everything should be accessible on the view?
Make sure you fix this everywhere.

hscontrol/api/v2/devices.go
```diff
@@ -0,0 +1,497 @@
...
+	id := strconv.FormatUint(uint64(node.ID), 10)
```
kradalby: We should have a type helper for this on node and nodeview

hscontrol/api/v2/keys.go
```diff
@@ -0,0 +1,413 @@
...
+			return &keys[i], nil
+		}
+	}
+
+	return nil, huma.Error404NotFound("auth key not found")
+}
+
+// parseKeyID parses the opaque key id, which Headscale renders as the decimal
+// pre-auth-key uint64. A non-numeric id is simply unknown -> 404.
+func parseKeyID(rawID string) (uint64, error) {
```
kradalby: Type helper that's reusable across the project

PR #3338, opened 2026-06-22
oidc: fix NewProvider ignoring caller context timeout

fredrikekre:
NewAuthProviderOIDC received a context with a 30-second timeout from its caller, but immediately discarded it by passing context.Background() to oidc.NewProvider. This caused both `headscale serve` and `headscale configtest` to hang indefinitely if the OIDC issuer was unreachable, rather than timing out and returning an error.

Fix by passing the caller's ctx to oidc.NewProvider and remove the nolint:contextcheck annotation that was suppressing the linter warning about this bug. The annotation was introduced in ce580f82 as part of a bulk lint-suppression pass without addressing the underlying issue.

- [ ] added unit tests
- [ ] added integration tests
- [ ] updated CHANGELOG.md

### DIFF
--- a/hscontrol/oidc.go
+++ b/hscontrol/oidc.go
@@ -87,7 +87,7 @@ func NewAuthProviderOIDC(
 ) (*AuthProviderOIDC, error) {
 	var err error
 	// grab oidc config if it hasn't been already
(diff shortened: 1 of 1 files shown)


kradalby (approved):
Thank you!, looks reasonable, I'll let the checks pass first.

fredrikekre:
Thanks!

PR #3342, opened 2026-06-23
ci: run integration tests as nix flake checks

kradalby:
hypothetical pr to see if we can speed up and run the CI tests inside a Nix enclosure. The goal is to have to rebuild less things by using Nix caches for things that dont change, and to make them run on [Garnix](https://github.com/garnix-io/garnix-ci) which I hope to set up.


integration/dsic/dsic.go
```diff
@@ -221,34 +227,57 @@ func New(
...
-		buildOptions,
-		runOptions,
-		dockertestutil.DockerRestartPolicy,
-		dockertestutil.DockerAllowLocalIPv6,
-		dockertestutil.DockerAllowNetworkAdministration,
-	)
+	// In CI / nix checks a prebuilt derper image is provided so we neither build
+	// the image nor clone tailscale at runtime, mirroring the tailscale and
+	// headscale image injection. Only the head image is prebuilt.
+	prebuiltImage := os.Getenv("HEADSCALE_INTEGRATION_DERPER_IMAGE")
```
kradalby: use tailscale envknob
it looks like this pattern is going to be used a bit so we should implement it in a reusable way.

nix/tests/integration.nix
```diff
@@ -0,0 +1,124 @@
...
+    # /tmp/control; cmd/hi mounts this dir. Without it every Status() errors on
+    # the write and the suite never observes NeedsLogin.
+    machine.succeed("mkdir -p /tmp/control")
+
+    # Run the suite inside a container (docker-out-of-docker), mirroring cmd/hi:
+    #  - name matches scenario.go's testSuiteName so it can join test networks
+    #  - docker socket mounted so sibling containers spawn on this daemon
+    #  - host nix store mounted so the test binary's closure resolves
+    out = machine.succeed(
+        "docker run --rm --name ${runnerName} "
```
kradalby, line 132: do we need to run the test binary inside docker on Nix? since there is only one, can we not just run the `integration.test` binary directly? Would save us a whole lot of mounting things here and there and complicated code when we already wrap it in a vm.

nix/images.nix
```diff
@@ -0,0 +1,173 @@
...
+  # Tailscale HEAD client + daemon + derper, built once from the pinned source.
+  # Mirrors Dockerfile.tailscale-HEAD / Dockerfile.derper's `go install`.
+  tailscaleHead = buildGoModule {
+    pname = "tailscale-head";
+    version = "head";
+    src = tailscaleSrc;
+
+    # Refresh after `nix flake update tailscale-head`: set to lib.fakeHash,
+    # build, copy the "got:" hash here.
+    vendorHash = "sha256-w6MF0pKV6zfeVorfrxMm2AXR0uDgw5sG3zl5Ugm6SCU=";
```
kradalby: This isnt manageable, and tailscale is a flake https://github.com/tailscale/tailscale/blob/main/flake.nix so we should pull it in like that.

nix/images.nix
```diff
@@ -0,0 +1,173 @@
...
+    config.Env = commonEnv;
+  };
+
+  derperImage = pkgs.dockerTools.buildLayeredImage {
+    name = "derper";
+    tag = "nix";
+    contents = [
+      tailscaleHead # provides derper
+      caShim
+      pkgs.dockerTools.binSh # dsic entrypoint is /bin/sh -c
```
kradalby: is this something we can simplify?

nix/images.nix
```diff
@@ -0,0 +1,173 @@
...
+      ln -sf ${headscale}/bin/headscale usr/local/bin/headscale
+    '';
+    config.Env = commonEnv;
+  };
+
+  derperImage = pkgs.dockerTools.buildLayeredImage {
+    name = "derper";
+    tag = "nix";
+    contents = [
+      tailscaleHead # provides derper
```
kradalby: does this mean we could just build one tailscale head container and use it as derper?

nix/images.nix
```diff
@@ -0,0 +1,173 @@
...
+    '';
+    config.Env = commonEnv;
+  };
+
+  # Pinned registry images, as docker-loadable tarballs that keep their
+  # repo:tag so the suite finds them locally (its own pull fails offline, then
+  # RunWithOptions uses the loaded image).
+  pullImage = spec: pkgs.dockerTools.pullImage ({
+    os = "linux";
+    arch = "amd64";
```
kradalby: we should make a matrix so we can support arm64

flake.nix
```diff
@@ -9,17 +9,26 @@
...
     # Reusable Go flake checks (build/test/lint/format); CI runs them via
     # `nix build .#checks.<system>.<name>` instead of bespoke per-tool steps.
     flake-checks.url = "github:kradalby/flake-checks";
     flake-checks.inputs.nixpkgs.follows = "nixpkgs";
+
+    # Tailscale HEAD, built from source for the integration test client image.
+    # Bump with `nix flake update tailscale-head` to track the latest main.
+    tailscale-head = {
+      url = "github:tailscale/tailscale";
+      flake = false;
```
kradalby: This is a flake https://github.com/tailscale/tailscale/blob/main/flake.nix

flake.nix
```diff
@@ -259,6 +349,46 @@
                   cat go.mod | ${pkgs.ripgrep}/bin/rg "\t" | ${pkgs.ripgrep}/bin/rg -v indirect | ${pkgs.gawk}/bin/awk '{print $1}' | ${pkgs.findutils}/bin/xargs go get -u
                   go mod tidy
                 '')
+
+              # Refresh nix/tailscale-versions.nix to the current MustTestVersions
+              # (run after a capver bump so the offline integration checks pin the
+              # tailscale images the suite actually requests).
+              (pkgs.writeShellScriptBin
```
kradalby: This could just be a subcommand of `hi` so we can implement it in go. or nushell or something a bit easier to read, and put it inside integration test folder. 

the output should be json instead of nix, and read like we read ./flakehashes.json so we can generate it with a bit more simple tooling.

kradalby: The same json list could be consumed in the go tests so we dont have to keep the list in the golang part, all reads the json that is the source of truth, generated by capver?

nix/tests/integration.nix
```diff
@@ -0,0 +1,124 @@
...
+    exit "$rc"
+  '';
+
+  pgEnv = lib.optionalString postgres
+    "-e HEADSCALE_INTEGRATION_POSTGRES=1 -e HEADSCALE_INTEGRATION_POSTGRES_IMAGE=postgres:16 ";
+in
+{
+  name = "integration-${name}${lib.optionalString postgres "-pg"}";
+
+  nodes.machine = { ... }: {
```
kradalby, line 61: how fast can we make a nixos vm boot? is there something equivalent of qemu microvms, we dont need most things?

nix/images.nix
```diff
@@ -42,6 +42,22 @@ let
     done
   '';
 
+  # Minimal busybox-style `adduser <name>` for the SSH integration tests. The
+  # from-scratch head image has no busybox/shadow, but the suite runs
+  # `adduser ssh-it-user` (tsic.WithExtraCommands) to create a login the
+  # tailscale SSH server resolves via /etc/passwd. A uid + /bin/sh + home is
+  # enough to run the test's `hostname` over SSH. Released alpine images already
+  # ship their own busybox adduser, so this is head-image only.
+  adduserShim = pkgs.writeShellScriptBin "adduser" ''
```
kradalby: We need to rethink this, it is not going to be feasible to add shim versions of whole distros like alpine.

I think the containers that are modified on startup needs to change, so if they try to install packages, we have them already in the base image, it cant make too much of a difference anyways. We should make sure the tools we need are there, in both the alpine and the nix image.

nix/images.nix
```diff
@@ -149,6 +181,54 @@ let
...
   };
 
+  # `apk add` shim for the released alpine images. Tests install packages at
+  # runtime — most via tsic.WithPackages (non-fatal, `;`-joined entrypoint) but
+  # TestTaildrop runs an explicit `apk add curl` whose error fails the test.
+  # Offline that can never reach the alpine CDN, and the tools are already baked
+  # in (testTools below), so treat `apk add` as satisfied when every requested
+  # package resolves to a present binary; fall through to real apk otherwise.
+  # On /usr/local/bin (ahead of /sbin in the alpine PATH). Mirrors caShim.
+  apkShim = pkgs.writeShellScriptBin "apk" ''
```
kradalby: same comment as earlier, this isnt viable. We need to just make sure we dont need apk in either images. Lets discuss options.

nix/images.nix
```diff
@@ -149,6 +181,54 @@ let
...
+
+  # Tools the suite `apk add`s into tailscale containers at runtime
+  # (curl/dig/python3/ssh). A nixosTest has no network, so apk can't run; bake
+  # them in instead. Symlinked into /usr/local/bin (in the alpine PATH) pointing
+  # at the nix store rather than overlaying /bin and /lib, which would shadow
+  # the musl loader and break the alpine tailscaled. Dynamic (cached), and one
+  # shared layer across all released images (docker dedups it on load).
+  testTools = pkgs.runCommand "ts-test-tools" { } ''
+    mkdir -p $out/usr/local/bin $out/usr/bin
+    ln -s ${apkShim}/bin/apk $out/usr/local/bin/apk
```
kradalby: we also need to get path sorted so we dont have to do all this symlinking.

nix/images.nix
```diff
@@ -158,7 +238,16 @@ let
     finalImageName = spec.imageName;
   } // spec);
 
-  tailscaleVersionImages = map pullImage versions.tailscale;
+  # Released tailscale images with the test tools layered on, keeping the exact
+  # registry repo:tag the suite requests.
+  withTestTools = spec: pkgs.dockerTools.buildImage {
```
kradalby, line 213: would it be easier to just build them all? and then have them in the cache?

.github/workflows/gh-action-integration-generator.go
```diff
@@ -8,9 +8,24 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"slices"
 	"strings"
 )
 
+// excludedFromNixChecks lists tests the hermetic nix-VM checks cannot run, with
+// the reason. They still run via cmd/hi, where the missing resource exists.
+var excludedFromNixChecks = map[string]string{
```
kradalby, line 18: we need to exclude these, but that also means we need to maintain the old pipeline so they do run in gh actions CI.

integration/dsic/dsic.go
kradalby: Done — the prebuilt-image knobs are now read through `integrationutil.PrebuiltImage` (tailscale `envknob`) plus a shared `ParseImageRef`; tsic/tsric/hsic/dsic all go through it.

nix/tests/integration.nix
kradalby: Prototyped, but running `integration.test` on the VM host needed dual-mode skip gates plus `/etc/hosts` injection and a bind-mount for the suite to reach containers — net more complexity than the docker-out-of-docker runner, so it was kept (your call to drop B3).

nix/images.nix
kradalby: Done — the `tailscale-head` input is now `flake = true` and the vendorHash is read from tailscale `flakehashes.json` (`.vendor.sri`). We still run our own `buildGoModule` because the suite needs `derper` + `containerboot` beyond their default package set.

nix/images.nix
kradalby: Simplified: dropped the apk/adduser shims (tools baked in and asserted present), removed the separate derper image (the head image serves derper + websocket), and replaced the per-tool symlinks with one `buildEnv` on PATH.

nix/images.nix
kradalby: Yes — done. `tailscale-head:nix` now serves client/daemon/derper/websocket and the separate `derperImage` is deleted.

nix/images.nix
kradalby: Done — `pullImage` arch follows the build system and the per-arch digests+hashes are pinned in `tailscale-versions.json`. Validated by building the v1.80 image for arm64 on x86 (genuinely arm64). No forced cross-build; an arm64 runner gets arm64 images.

flake.nix
kradalby: Done — the input is `flake = true` and the vendorHash is read from tailscale `flakehashes.json`.

flake.nix
kradalby: Done, with one tweak after discussion: instead of a `hi` subcommand it is a standalone `go generate` generator (`tools/tailscale-versions`), matching the repo `tools/capver` + integration-matrix generators. Output is `integration/tailscale-versions.json`, shaped like `flakehashes.json`, read by both the Go suite (`go:embed`) and nix (`fromJSON`).

nix/tests/integration.nix
kradalby: Measured: 18.7s boot (1.7s kernel + 3.9s initrd + 13.1s userspace); the userspace critical chain is `dhcpcd -> network-online -> docker`, i.e. functional rather than trimmable. Trimmed documentation + timesyncd. microvm.nix was evaluated and rejected — it still needs a full docker daemon + nested KVM, and boot is only ~10-15% of the 2-3 min image-load + test, so it cannot repay a new flake input.

nix/images.nix
kradalby: Done — no distro shims. The tools the suite used to `apk add` (curl/ssh/dig/python3/getent) are baked into both images at build time, and tooling is asserted-present (fails loudly) instead of installed.

nix/images.nix
kradalby: Done — `apk` is never invoked: `TestTaildrop` no longer runs `apk add curl` (curl is baked in) and the apkShim is deleted.

nix/images.nix
kradalby: Done — one `pkgs.buildEnv` puts every tool `/bin` on PATH; the per-tool symlinks are gone and `doSSH` resolves `ssh` via PATH instead of a hardcoded path.

nix/images.nix
kradalby: Kept pulling the pinned digests instead of building: each released version built from source would need its own pinned tailscale source + vendorHash, whereas the multi-arch registry digests pull straight from cache. Building from source only pays off for HEAD, which we already do.

.github/workflows/gh-action-integration-generator.go
kradalby: Done — `excludedFromNixChecks` drives both the nix-check exclusion and a generated `excluded-tests.json` matrix that `test-integration-legacy.yaml` runs via `cmd/hi`. (Also just fixed that workflow to build the headscale binary before the image build.)

PR #3366, opened 2026-07-04
fix: lowercase DNS extra record names

### LINKED ISSUE #2782: [Bug] extra records (JSON) works only when names are in lowercase

larrasket:
DNS names are case-insensitive, but clients match extra records against the lowercased query name, so records with mixed-case names (for example "Printer.fritz.box" in an extra_records_path file) never resolved and queries fell through to the global nameserver.

Normalize record names to lowercase where the records enter the tailcfg DNS config, covering both dns.extra_records and extra_records_path.

Fixes #2782

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ keys remain all-access.
 
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
+- Lowercase DNS extra record names so mixed-case records resolve [#2782](https://github.com/juanfont/headscale/issues/2782)
 
 ## 0.29.2 (2026-07-01)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,14 @@ keys remain all-access.
 
 - `headscale users rename` now sends the identifier of the matched user instead of the raw `--identifier` flag value, so renaming by name works again [#3442](https://github.com/juanfont/headscale/pull/3442)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ keys remain all-access.
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
 - Headscale now requires Go 1.27 to build
+- Lowercase DNS extra record names so mixed-case records resolve [#2782](https://github.com/juanfont/headscale/issues/2782)
```
kradalby, line 61: Lets link the PR here in the changelog to follow the same style. Can you move it to under 0.29.4 and we get it out quicker.

kradalby (approved):

PR #3375, opened 2026-07-09
cli: redesign proposal 

kradalby:
This pr is not to be merged, its end result is to serve as a "design doc" for how we organise our cli. Breaking changes allowed.

Comments, suggestions and proposals from humans in the community encouraged.

### DIFF
--- a/docs/proposals/001-cli.md
+++ b/docs/proposals/001-cli.md
@@ -1,0 +1,283 @@
+# Headscale CLI redesign
+
+**Status:** draft, for discussion. This PR exists so we can shape the design together — it is not meant to be merged as-is.
+
+The current CLI addresses targets inconsistently: nodes take `--identifier` (ID only), users take `--identifier` or `--name`, API keys take `--id` or `--prefix`. Positional arguments are only used in a few places. `nodes register` is deprecated in favour of `auth register` but still around. Tables have a fixed set of columns. This proposal restructures the command tree and its conventions.
+
+## Requirements
+
+- `--force` override for all commands requiring prompts
+- `--output` on every command offering `json`, `json-line` / `jsonl`, drop `yaml`
+- `--help` auto generated for all commands with good documentation
+    - usage examples
+    - good command enumeration and descriptions
+    - the full nested command tree must be presented so people can find commands
+- Any irreversible command should ask for confirm (or `--force`): `delete`, `expire`, `reject`
+- In principle, the entire API should be covered by the CLI. Any operation that can be done with the API should be available. Each entity in Headscale should have CRUD where it makes sense.
+
+## Current root
+
+```
+❯ headscale --help
+
+headscale is an open source implementation of the Tailscale control server
+
+https://github.com/juanfont/headscale
+
+Usage:
+  headscale [command]
+
+Available Commands:
+  apikeys     Handle the Api keys in Headscale
+  auth        Manage node authentication and approval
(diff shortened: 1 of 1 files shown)


docs/proposals/001-cli.md
```diff
@@ -0,0 +1,283 @@
...
+
+headscale nodes tags 12 tag:test,tag:example
+
+headscale nodes set-expiry 12 2025-08-27T10:00:00Z # support more formats
+headscale nodes set-expiry 12 now
+headscale nodes expire 12 # same as set-expiry now
+headscale nodes disable-expiry 12
+```
+
+An alternative shape is a top-level `expire` group instead of `nodes set-expiry`. We want reviewer input on which to pick:
```
kradalby: I currently like this, where we split out the very much used node ones, but expiry is a bit weird because it also exist in different scopes.

Maybe only routes and tags?

nblock:
> `auth` is the authentication of nodes to headscale

Authentication of nodes and also SSH check.

> headscale routes approve 12 "" # remove all from 12
> headscale routes disapprove-all 12 # remove all from 12

An approve subcommand with an empty last target argument is confusing to some users. Because its unclear whether the user forgot the last argument or all routes should really be disapproved. We also have this now when using `-r ""`. It came up a few times.

Maybe add a pair of subcommands to solve this in general? `approve` / `disapprove` or `approve` / `no-approve` and always require the last argument (in this case the route(s)).

> headscale users set <name> # optional fields are flags

I like the "optional fields as flags". Its flexible and can be extended as needed.

> We will support outputting json, json-line and a human readable table. Default output mode is always the human
> readable table. json formats will always show all data.
 
Do we even need json-line? If not, we could only provide json as universal machine readable output format and combine it with jsonpath somehow? Otherwise we'd have four formats:

- text (mangled by `--column` and `--filter`)
- json, json-line (mangled by `--filter`)
- json-path (arbitary data in json format)

> expire vs. nodes expire

I like the "nodes expire" form better. Adding one more word to the argument list does not really matter but it imho
improves readability and discoverability:

- Which operations can I perform for a node/user/…: `headscale nodes --help`
- The target - in this case node needs to be supplied anyways. If there are more possible target groups, we'd need the distinction anyways. An example would be nodes and apikeys for expire: `headscale expire node <...>` / `headscale expire apikeys <...>`.

Other questions:

- Are global arguments kept as is? `--config` comes to mind.
- Are environment variables still supported as-is (`HEADSCALE_CONFIG`, …)

nblock:
How should we do the formatting of the text output? Some columns are single valued (node name, user name) and some are lists (ip addresses, routes)? Should we keep it fixed (without stability guarantees) and direct users to `jsonpath` for script processing?

The text version should favor human readability and allow for some flexibility for upcoming changes (add, remove, reorder default columns). While its  nice to use grep and friends to work with the CLI, there's json and jsonpath which is better suited for scripting.

The "one line per list entry" output is good for variable long lists such as routes. For the node's IP addresses its a bit hard to read as its almost always two lines which produces alternating rows and make it visually harder to read.

ArcticLampyrid:
> Do we even need json-line? If not, we could only provide json as universal machine readable output format and combine it with jsonpath somehow?

Perhaps we could omit the `jsonpath` option too. Providing only table and json formats is sufficient. When users require specific data extraction, they can handle it in the pipeline using tools such as `jq`. In modern IT operations, tools like `jq` are virtually always pre-installed.

PR #3383, opened 2026-07-15
fix(poll): do not cancel ephemeral GC until Connect succeeds

iSerganov:
### The issue
---------

With `node.ephemeral.inactivity_timeout` set (e.g. 15m), ephemeral nodes are usually deleted after they go offline, but under reconnect churn some departed nodes stay in the node list as disconnected indefinitely until they are removed manually or Headscale is restarted.

### The root cause
--------------
Ephemeral cleanup is timer-based via EphemeralGarbageCollector, not a periodic LastSeen scan. serveLongPoll cancelled any pending GC timer at the very start of a long-poll attempt, and only rescheduled on a clean disconnect after Connect.
If a reconnect cancelled the timer and then failed before Connect (for example `UpdateNodeFromMapRequest` error), the deferred cleanup saw `connectGen == 0` and returned without Schedule. The node remained offline with no deletion timer and no reconciler to recover it.

### The fix
-------
Move `beforeServeLongPoll` (GC Cancel) to after a successful Connect so a failed reconnect leaves an already-armed inactivity timer intact. Successful reconnects still cancel GC once the node is online; a later disconnect reschedules as before.

### Tests 
-------
Add `TestFailedReconnectDoesNotCancelEphemeralGC` to lock the ordering + IsScheduled and `DeleteNodeFromStoreForTest` helpers for the test.

### DIFF
--- a/hscontrol/poll.go
+++ b/hscontrol/poll.go
@@ -96,12 +96,6 @@ func (m *mapSession) stopFromBatcher() {
 	}
 }
 
-func (m *mapSession) beforeServeLongPoll() {
-	if m.node.IsEphemeral() {
-		m.h.ephemeralGC.Cancel(m.node.ID)
-	}
-}
-
 // afterServeLongPoll is called when a long-polling session ends and the node
 // is disconnected.
 func (m *mapSession) afterServeLongPoll() {
@@ -144,8 +138,6 @@ func (m *mapSession) serve() {
 //
 //nolint:gocyclo
 func (m *mapSession) serveLongPoll() {
-	m.beforeServeLongPoll()
-
 	m.log.Trace().Caller().Msg("long poll session started")
 
 	// connectGen is set by [state.State.Connect] below and captured by the deferred cleanup closure.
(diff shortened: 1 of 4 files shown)


hscontrol/poll.go
```diff
@@ -248,6 +246,11 @@ func (m *mapSession) serveLongPoll() {
 
 	connectChanges, connectGen = m.h.state.Connect(m.node.ID)
 
+	// Cancel ephemeral GC only after Connect succeeds. Cancelling at the start
+	// of serveLongPoll left departed nodes without a deletion timer when a
+	// reconnect attempt failed before Connect (issue #3382).
+	m.beforeServeLongPoll()
```
kradalby: considering this is not called as a hook before the long poll, we would need to rethink the naming, or if this hook is valuable at all. Maybe the content, or the relevant content to this issue needs to move here instead of this function.

iSerganov: Makes sense. Renamed it to `cancelEphemeralNodeGC`. Please let me know if any other option is preferable. Thank you!

hscontrol/poll.go
```diff
@@ -248,6 +246,11 @@ func (m *mapSession) serveLongPoll() {
 
 	connectChanges, connectGen = m.h.state.Connect(m.node.ID)
 
+	// Cancel ephemeral GC only after Connect succeeds. Cancelling at the start
+	// of serveLongPoll left departed nodes without a deletion timer when a
+	// reconnect attempt failed before Connect (issue #3382).
+	m.cancelEphemeralNodeGC()
```
kradalby: Can you actually drop the function and just inline the check?

```
	if m.node.IsEphemeral() {
		m.h.ephemeralGC.Cancel(m.node.ID)
	}
```

We dont use it for anything else and the original idea was that the hook would expand but it never did.

kradalby (approved):
Looks good, before merge, can you squash the commit into on commit that follows https://go.dev/wiki/CommitMessage.

Thanks!

hscontrol/poll.go
iSerganov: Done. The function has been removed, now PR consists of a single commit where commit message follows  https://go.dev/wiki/CommitMessage.

PR #3387, opened 2026-07-17
Add AI policy

juanfont:
This PR adds a new AI policy.

We fully support (and use) LLMs for coding, but we still like talking to humans.

Copied from https://github.com/BurntSushi/bstr/pull/232


kradalby (approved):
Needs a run of `prettier` formatting.

PR #3388, opened 2026-07-17
.github: gate PRs via mitchellh/vouch

juanfont:
PRs keep arriving without the prior discussion CONTRIBUTING.md asks
for. We are not able to review or triage PRs fast enough, specially now that everyone has super human code generation capabilities.

Adopt [mitchellh/vouch](https://github.com/mitchellh/vouch) to
auto-close PRs from authors not on .github/VOUCHED.td. Issues stay
open to everyone — the close message points contributors back at
CONTRIBUTING.md to start that discussion.

Less aggressive than ghostty, which gates issues too.

Seed list is past PR contributors with merged code plus authors of
detailed issues; bots and CODEOWNERS are auto-vouched. Anyone with an
open PR that hasn't been through review yet was left off — !vouch
later if it lands.

!vouch @user / !denounce @user / !unvouch @user on any issue or PR
comment from a maintainer manages the list. Weekly cron resyncs
CODEOWNERS.

Workflows need a GitHub App with Contents/Issues/PRs write installed on
the repo, plus VOUCH_APP_ID and VOUCH_APP_PRIVATE_KEY secrets. Until
those exist nothing is gated (fail-open).


.github/workflows/vouch-check-pr.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          sparse-checkout: .github/pr-unvouched-message
+
+      - uses: mitchellh/vouch/action/check-pr@d66fa29a64600490892131ad87597c30c91fcac4 # v1.5.0
+        with:
+          pr-number: ${{ github.event.pull_request.number }}
+          auto-close: true
+          template-file: .github/pr-unvouched-message
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```
kradalby, line 30: Need to run `prettier` formatter on all files.

.github/workflows/vouch-check-pr.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+        id: app-token
+        with:
+          app-id: ${{ secrets.VOUCH_APP_ID }}
+          private-key: ${{ secrets.VOUCH_APP_PRIVATE_KEY }}
```
kradalby, line 18: This step isnt needed, see secrets comment

.github/workflows/vouch-check-pr.yml
kradalby, line 30: This already exists `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`, I think that will be correct. Needs to be on all

.github/VOUCHED.td
```diff
@@ -0,0 +1,119 @@
...
+TotoTheDragon
+tsujamin
+unreality
+vbrandl
+vdovhanych
+victorhooi
+vsychov
+yaneony
+yaroslavkasatikov
+Zottelchen
```
kradalby, line 119:
```suggestion
Zottelchen
iSerganov
arpitjain099
fredrikekre
```

PR #3400, opened 2026-07-28
hscontrol: cancel tailsql on graceful shutdown

iSerganov:
Serve stored a Background context for tailsql and called context.Done() during signal shutdown. Done only returns the done channel and does not cancel, so tailsql never stopped when Headscale shut down.

Create a cancellable child of the Serve context, run `runTailSQLService` in the listener errgroup so its error is surfaced, and call the cancel func on shutdown - tailsql unblocks on ctx.Done.

### DIFF
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -752,6 +752,7 @@ func (h *Headscale) Serve() error {
 		}
 
 		var tailsqlCtx context.Context
+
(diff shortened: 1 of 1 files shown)


kradalby:
Looks reasonable, could you rebase it from main, and fix the lint error?

iSerganov:
> Looks reasonable, could you rebase it from main, and fix the lint error?

Done.

kradalby (approved):

PR #3409, opened 2026-08-02
chore: fix tagged exit node visibility when only matcher is with via

### LINKED ISSUE #3408: [Bug] Exit node not visible when rule is matched with `via` tag

### Current Behavior

# Exit node not offered to viewers whose only matching grant uses `via` (0.29.2)
 
## Environment
 
- **Headscale version:** 0.29.2
- **Policy mode:** Database
- **Database:** sqlite
- **Deployment:** bare metal
- **Clients:** Tailscale on macOS, `1.98.5` (admin) and `1.98.5` (affected user)
## Summary
 
A grant of the form:
 
```json
{
  "src": ["autogroup:member"],
  "dst": ["autogroup:internet"],
  "via": ["tag:exit"],
  "ip": ["*"]
}
```
 
does not cause a tag-carrying exit node to be offered to viewers who match it. The affected clients report "No Exit Nodes Available".
 
Users who reach the same exit node through a *separate, non-via* rule (in our case `group:admins` → `dst: ["*"]`) see and use it without any problem. Removing `via` from the grant makes the exit node appear for all members immediately.

### Expected Behavior

Users who match the rule should see all exit nodes that carrie that tag.

## Reproduction
 
Minimal policy:
 
```json
{
  "tagOwners": {
    "tag:exit": ["admin@example.com"]
  },
  "autoApprovers": {
    "exitNode": ["tag:exit"]
  },
  "groups": {
    "group:admins": ["admin@example.com"]
  },
  "grants": [
    {
      "src": ["group:admins"],
      "dst": ["*"],
      "ip": ["*"]
    },
    {
      "src": ["autogroup:member"],
      "dst": ["autogroup:internet"],
      "via": ["tag:exit"],
      "ip": ["*"]
    }
  ]
}
```
 
1. Register a node, tag it `tag:exit`, advertise `0.0.0.0/0` and `::/0`. Confirm both are approved (`headscale nodes list-routes`).
2. Log in as `admin@example.com` (matches grant 1). The exit node is listed and works.
3. Log in as any other user. Their node is untagged, so it resolves into `autogroup:member` and matches grant 2 only.
4. **Expected:** the exit node is offered, restricted to nodes carrying `tag:exit`.
5. **Actual:** "No Exit Nodes Available". The exit node is not selectable.
6. Replace grant 2 with the same rule minus `via` — the exit node appears for that user immediately.

### Environment

```markdown
- OS: NixOS
- Headscale version: 0.29.2
- Tailscale version: 1.98.5
```

### Runtime environment

### Debug information

Can't share debug files, it's from the "production" server. Happy to check anything, though.

vdovhanych:
- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md

This should fix https://github.com/juanfont/headscale/issues/3408

If you decide to merge this, I need to update the changelog so it links to pr instead of the issue, and put it somewhere else in the list if needed.


kradalby:
ok, so I looked more into this, and I did a capture of the upstream control server too to validate.

What I discovered was that the function you "chained" with more context was kind of not needed anymore as we actually capture all the relationships inside the new state. This means that the code we used to determine these relationships was outdated and we could actually trim them away.

I will merge this, but would appreciate if you have the opportunity to test it. I want to cherry pick this onto the 0.29.4 patch release when confirmed.

vdovhanych:
> ok, so I looked more into this, and I did a capture of the upstream control server too to validate.
> 
> What I discovered was that the function you "chained" with more context was kind of not needed anymore as we actually capture all the relationships inside the new state. This means that the code we used to determine these relationships was outdated and we could actually trim them away.
> 
> I will merge this, but would appreciate if you have the opportunity to test it. I want to cherry pick this onto the 0.29.4 patch release when confirmed.

Hi, no problem, just ping me here or on discord when its ready ill deploy it to the test env.

kradalby:
Its all ready! let me know how it goes!

kradalby (approved):

vdovhanych:
Seems to be working as intended, could not replicate the original issue anymore. But I haven't tested it very thoroughly, I just checked the original issue.

kradalby:
Perfect thank you!

Issue #2272, opened 2024-12-06
[Feature] get remote IP


celevra:
### Use case

it would be great to get the real IP (not the tailscale IP) of a node

### Description

the ip should be listet in the cli and in the api

### Contribution

- [ ] I can write the design doc for this feature
- [ ] I can contribute this feature

kradalby:
This is a bit relative, not saying we can't but we have to decide which or all.

There will be the IP that is connected to headscale, which we likely can get from the go http server.

Then there is all the discovered endpoints that the tailscale client discovers as part of magicsock. This list might include old endpoints as it is quite opportunistic.

It is a bit non obvious to which ones are useful.

What is the use case for consuming them?

celevra:
for us the connected ip from go http server is what we are looking for.
it is not our main use case for tailscale but as the information exists we want to set dyndns names for devices that don't support dyndns but support tailscale

candidco-engineering:
It would be useful to get the real ip of the node. Mostly for audit purposes.

celevra:
so this means it gets not implemented?

kradalby:
No it could be, but likely not possible until I get around to reworking the connection map, which will be many releases out as there is quite a lot of other stuff in the pipeline atm.

kradalby:
@danya0123 please post on the relevant issue, you even found it #2152.

kradalby:
Keeping this under #3363 (API v2) as primary since the field doesn't exist in any response yet — but the original ask is explicit that it needs both: "the ip should be listet in the cli and in the api." Once Endpoints is added to the v2 Device response, #3364 (CLI redo) needs a matching nodes-list column. Flagging here so the CLI half doesn't get dropped.

kradalby:
I posted a comment on https://github.com/juanfont/headscale/pull/3419#issuecomment-5421531990, now that we are doing the v2 api, I think there is a nice slot for this.

PR #3419, opened 2026-08-11
api/v2: expose current or last known confirmed IP

bigbozza:
I've ammended this PR to replace `otherKnownIPs` with the Tailscale-compatible device posture endpoint:

` GET /api/v2/device/{id}/attributes`

It exposes `ip:publicAddress`, containing the current or last known (likely public) address from which the device most recently made an authenticated connection to Headscale. The value is persisted so it remains available while the device is offline.

The address comes from Headscale’s web server connection information, not from the device’s reported Magicsock endpoints. Existing `trusted_proxies` configuration is respected when Headscale is behind a reverse proxy.

The endpoint returns empty attribute and expiry objects until an address has been observed. This change also adds the corresponding OAuth scopes, database migration, generated client support, and tests.

### DIFF
--- a/hscontrol/api/v2/devices.go
+++ b/hscontrol/api/v2/devices.go
@@ -53,6 +53,13 @@ type DeviceRoutes struct {
 	Enabled    []string `json:"enabledRoutes"    nullable:"false"`
 }
 
+// DevicePostureAttributes is the Tailscale-compatible posture attribute
+// response. Headscale currently exposes only the server-derived public address.
+type DevicePostureAttributes struct {
+	Attributes map[string]any       `json:"attributes" nullable:"false"`
+	Expiries   map[string]time.Time `json:"expiries"   nullable:"false"`
+}
+
 // Request bodies match the Tailscale SDK wire shapes.
 type (
 	setAuthorizedRequest struct {
@@ -80,6 +87,9 @@ type (
 		DeviceID string `doc:"Device id (the decimal node id)." path:"id"`
 		Fields   string `doc:"Set to \"all\" for route fields." query:"fields"`
 	}
+	deviceAttributesInput struct {
+		DeviceID string `doc:"Device id (the decimal node id)." path:"id"`
+	}
 	listDevicesInput struct {
 		Tailnet string `doc:"Tailnet; must be \"-\"." path:"tailnet"`
 		Fields  string `query:"fields"`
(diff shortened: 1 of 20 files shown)


hscontrol/api/v2/devices.go
```diff
@@ -448,6 +450,34 @@ func deviceFromView(view types.NodeView, allFields bool) Device {
 	return d
 }
 
+// otherKnownIPs returns the unique IP addresses from the node's last reported
+// endpoint set, excluding its Headscale-assigned tailnet addresses.
+func otherKnownIPs(view types.NodeView) []string {
```
kradalby: The endpoint set also contains a lot of potentially stale issues, and I believe that #2272 asks for the IP the node last used, and the most reliable one would be the one used to contact the control server. 

This can be gotten from the web server part of headscale.

kradalby:
Sorry for the long wait, I've been on holiday.

Now that we are implementing an API that is largely aiming to be compatible with Tailscale, we should also put this IP in the right place. 

Tailscale has this notion called "device attributes" which has a field for this particular IP that is requested in #2272, and it would fit nicely to wire up that single attribute as our "initial device attribute": https://tailscale.com/docs/features/device-posture#geolocation-attributes

`ip:publicAddress` is used to represent the IP that the node last connected to the control server and can then be fetched from the API: https://tailscale.com/api-docs#tag/devices/GET/device/{deviceId}/attributes.

If you are up for giving it a go, that would be welcome, it would fit nicely with this planned release. If not I will get to it eventually. Since this will be some foundational work for working out the device attributes, please keep it to this one to avoid scope creep. 

We can do the basic ones (`node:`) and expiries in a follow up, while the `custom:` ones can be more of a future thing:

```
{
  "attributes": {
    "custom:myScore": 80,
    "custom:diskEncryption": true,
    "custom:myAttribute": "my_value",
    "node:os": "linux",
    "node:osVersion": "5.19.0-42-generic",
    "node:tsReleaseTrack": "stable",
    "node:tsVersion": "1.40.0",
    "node:tsAutoUpdate": false,
    "node:tsStateEncrypted": false
  },
  "expiries": {
    "custom:myScore": "2024-04-23T18:25:43.511Z"
  }
}
```

bigbozza:
ok i've reworked it @kradalby

Issue #3415, opened 2026-08-07
[Bug] macOS/iOS - no exit nodes available


alexls74:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I'm not sure if this is a headscale issue, but tailscale 1.102.2 on macOS and iOS does not display the exit nodes list.
If start a connection from the terminal, it occurs and the list of available ones appears. After restarting the application, the list is not visible again.

### Expected Behavior

The exit nodes list should be visible in the GUI.

### Steps To Reproduce

1. MacOS or iOS. Tailscale 1.102.2
2. GUI->Exit Nodes
3. Tailnet Exit Nodes should be visible

### Environment

```markdown
- OS: Debian 13
- Headscale version: 0.29.3 (docker) 
- Tailscale version: 1.102.2
```

### Runtime environment

- [ ] Headscale is behind a (reverse) proxy
- [x] Headscale runs in a container

### Debug information

No errors in tailscale debug daemon-logs

GenkaOk:
I confirmed. After update Tailscale to 1.102.2, it hide any exit-nodes.

alexls74:
Tailscale support responded: "This is a known class of issue when using a custom/third-party control server like headscale. A previous similar report noted that the macOS GUI showed no exit nodes while the CLI worked fine. A Tailscale team member suggested the root cause was headscale not correctly setting the online/offline status bits for peer nodes in the MapResponse."

toni-cerise:
Same issue here, `tailscale exit-node list` correctly lists the node, `tailscale set --exit-node=<ip>` connects the macOS machine, and the exit node gets visible in the GUI (Tailscale for macOS, Version 1.102.1) and routes traffic correctly. On iOS, the node is invisible as described. 

See output from `tailscale version` below:

```
1.102.1
  tailscale commit: 8ebe8f7c36d79b792714ed8360ca55dd1b38cc23
  long version: 1.102.1-t8ebe8f7c3-gda6192991
  other commit: da61929913f0135ffb1f0d10f16858700f9652f5
  go version: go1.26.5 (tailscale/go 63ae404c82)
```

TAT-Hins:
Same issue here, any workaround solutions for iPhone and iPad right now?

alexls74:
> Same issue here, any workaround solutions for iPhone and iPad right now?

Nothing at all ☹️

arthur-lbchn:
[...]

PR #3420, opened 2026-08-11
state,change,mapper: report LastSeen for online peers per tailcfg protocol

zicochaos:
tailcfg documents `Node.LastSeen` as "when the node was last online … nil if never been online", and `PeerChange.LastSeen` as accompanying online-status changes. headscale only sent `LastSeen` for offline peers and never included it in online/offline patches, so clients on the new InitialStatus + peer-delta code path (Tailscale 1.102 on Apple platforms, tailscale/tailscale@`6bf05cb63`) saw online peers as never-online — rendering exit nodes as offline in the GUI.

- emit `LastSeen` for online peers in `TailNode`, per tailcfg semantics
- set `LastSeen` on connect/disconnect in `State` and carry it in the `NodeOnline`/`NodeOffline` change patches

Covered by a new servertest regression test and updated change/mapper tests. Note: this alone does not fix the empty exit-node list from #3415 — that turned out to be the GUI requiring a suggested exit node; see the issue for analysis and a policy (`nodeAttrs`) workaround.

Updates #3415

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ keys remain all-access.
 - Fix `headscale users rename` sending the raw `--identifier` flag value instead of the matched user's identifier, so renaming by name works again [#3442](https://github.com/juanfont/headscale/pull/3442)
 - Fix tailsql not shutting down with headscale, leaving the process hanging on graceful shutdown [#3400](https://github.com/juanfont/headscale/pull/3400)
 - Fix tvOS setup instructions: install the VPN configuration before setting the coordination server URL [#3431](https://github.com/juanfont/headscale/pull/3431)
+- The peer patches sent when a node connects or disconnects now carry `LastSeen`, so clients learn when a peer was last seen without waiting for a full map [#3420](https://github.com/juanfont/headscale/pull/3420)
 
 ## 0.29.3 (2026-07-29)
 
--- a/hscontrol/types/change/change.go
+++ b/hscontrol/types/change/change.go
@@ -395,27 +395,34 @@ func DNSConfig() Change {
 	}
 }
 
-// NodeOnline creates a patch response for a node coming online.
-func NodeOnline(nodeID types.NodeID) Change {
+// NodeOnline creates a patch response for a node coming online. lastSeen is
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ keys remain all-access.
 
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
+- Report `LastSeen` for online peers (frozen at connect time) and carry it on online/offline transition patches, matching the tailcfg protocol and the hosted control plane; the Tailscale 1.102 macOS/iOS GUI hid exit nodes whose LastSeen was never set [#3415](https://github.com/juanfont/headscale/issues/3415)
```
kradalby: changelog should refer to the PR. 
The line is really long, and it includes implementation detail, which we typically avoid as we link to the PR so we can focus on what has changed.

hscontrol/types/change/change.go
```diff
@@ -395,27 +395,34 @@ func DNSConfig() Change {
 	}
 }
 
-// NodeOnline creates a patch response for a node coming online.
-func NodeOnline(nodeID types.NodeID) Change {
+// NodeOnline creates a patch response for a node coming online. lastSeen is
+// the time the node connected; per the tailcfg protocol, a [tailcfg.PeerChange]
```
kradalby, line 399: these comments are unnecessary verbose, parts like "; per the tailcfg protocol" doesnt add a lot of value. Please do a pass of all the comments and focus on why something is the way it is, lean on brevity.

kradalby (approved):
Some comments to address. 
It is funny because in my years working on this project, we have probably added and removed this multiple times as over time it seems to have been unused, then used. Lets get it in again 😆 

You probably have to rebase main.

CHANGELOG.md
kradalby: Lets also get this into 0.29.4 in the changelog.

PR #3424, opened 2026-08-14
Reject unknown OAuth client scopes at creation

### LINKED ISSUE #3406: [Bug] OAuth client creation accepts and stores undocumented scopes

### Current Behavior
OAuth client creation accepts scope strings that are not part of the documented scope vocabulary and stores them verbatim.

The CLI help lists the expected scope values:

    auth_keys, oauth_keys, devices:core, devices:routes, policy_file, feature_settings, read variants, all, all:read

However, the create path only requires at least one scope and does not validate the submitted values against that vocabulary.

Relevant code in cmd/headscale/cli/oauth_client.go:

    if len(scopes) == 0 {
        return fmt.Errorf("at least one --scope is required: %w", errMissingParameter)
    }

    resp, err := client.CreateKeyWithResponse(ctx, oauthTailnet, clientv2.CreateKeyRequest{
        KeyType:     &keyType,
        Scopes:      &scopes,
        Tags:        &tags,
        Description: &description,
    })

The database creation path validates tags with validateACLTags, but scopes are only deduplicated and sorted before being stored.

Relevant code in hscontrol/db/oauth.go:

    tags, err := validateACLTags(tags)
    if err != nil {
        return "", nil, err
    }

    scopes = set.SetOf(scopes).Slice()
    slices.Sort(scopes)

    client := types.OAuthClient{
        Scopes:      scopes,
        Tags:        tags,
        Description: description,
        UserID:      creatorUserID,
        CreatedAt:   &now,
    }

Local validation output:

    #### STEP 1: malicious/arbitrary scopes + VALID tag  (expect SUCCESS) ####
    OAuth client b54801670ee2 created.

    #### STEP 2: list -> confirm bogus scopes persisted verbatim ####
    ID           | Scopes                               | Tags      | Description              | Created
    b54801670ee2 | *::*,../../etc/passwd,EVIL:superuser | tag:legit | repro-unvalidated-scopes | 2026-07-30 04:43:58

    #### STEP 3: CONTRAST — valid scope + INVALID tag 'notagprefix' (expect FAIL) ####
    Error: api error (400): creating oauth client

This shows that tags are rejected when malformed, while undocumented scope strings are accepted and stored.

### Notes:

    hscontrol/scope.Parse keeps unknown scope strings as-is, and the code comments state that unknown strings do not satisfy required scopes. So this report is about input validation and stored OAuth client configuration consistency, not a demonstrated privilege escalation.

### Suggested fix:

    Add scope validation next to validateACLTags in CreateOAuthClient. The validator should deduplicate, sort, and reject any scope that is not in hscontrol/scope.Known().

### Expected Behavior

OAuth client creation should reject scope values that are not part of the known/documented scope vocabulary.

If the intended behavior is to allow forward-compatible unknown scopes, that behavior should be explicit in the CLI help and API documentation.

### Steps To Reproduce

1. Start headscale in a local container or development environment.
2. Create an OAuth client with undocumented scope strings and a valid tag:

       headscale oauth-clients create --scope "EVIL:superuser" --scope "*::*" --scope "../../etc/passwd" --tag "tag:legit" -d "repro-unvalidated-scopes"

3. List OAuth clients:

       headscale oauth-clients list

4. Observe that the undocumented scope strings are stored and displayed.
5. As a contrast, create an OAuth client with a valid scope but an invalid tag:

       headscale oauth-clients create --scope "all" --tag "notagprefix" -d "repro-bad-tag"

6. Observe that the invalid tag is rejected.

### Environment

```markdown
### Environment

- OS: Linux / Docker
- Headscale version: tested at commit 565fd254d06c4c7f9a8cad1714a43445c79ba420
- Tailscale version: not applicable
- Runtime environment: Headscale runs in a container
```

### Runtime environment

### Debug information

Tested commit:

    565fd254d06c4c7f9a8cad1714a43445c79ba420

MsfPablo:
Fixes #3406

- [x] raised a GitHub issue or discussed it on the projects chat beforehand (#3406)
- [x] added unit tests
- [ ] added integration tests — the behaviour is fully covered at the `db.CreateOAuthClient` chokepoint every caller (CLI and v2 API) goes through; happy to add an `integration/cli_oauth_clients_test.go` case if you'd prefer one.
- [ ] updated documentation if needed — the CLI `--scope` help text now also lists `users`, which was already part of the vocabulary
- [x] updated CHANGELOG.md

## Problem

`CreateOAuthClient` validated tags via `validateACLTags` but not scopes: any string was deduplicated, sorted and stored verbatim, so `headscale oauth-client create -s 'EVIL:superuser' -s '../../etc/passwd'` succeeded. Since an unknown scope satisfies nothing in `scope.Grants`, this silently mints a client with fewer permissions than the operator asked for, and a typo like `auth_key` fails at request time rather than creation time.

## Change

- `scope.Scope.Valid()` reports membership in the existing `scope.Known()` vocabulary — the canonical list already used by the grant predicate, so validation cannot drift from enforcement.
- `db.validateScopes` replaces the inline dedup/sort in `CreateOAuthClient`, rejecting unknown values with `ErrOAuthClientScopeInvalid` and an error naming the accepted set.
- `mapError` maps that error to 400, so an invalid scope surfaces exactly like an invalid tag does today.
- Validation lives in the db layer rather than the CLI so the v2 API path (`POST` keys) is covered too.

## Verification

```
go build ./...                       # clean
go vet ./...                         # clean
gofmt -l .                           # no output
go test ./hscontrol/... ./cmd/...
```

Test suite is clean apart from two failures that also reproduce on an unmodified checkout in this environment: `TestAPIv2OAuthScopes` (needs `tofu`, provided by the nix dev shell) and `TestDialHeadscaleSocketRetriesUntilPresent` (macOS unix socket path length).

---
Disclosure: this patch was prepared with the assistance of Claude Code. I have reviewed the change and the test results above myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,6 @@ keys remain all-access.
 
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
-- Creating an OAuth client now rejects scopes outside the known vocabulary instead of storing them verbatim [#3406](https://github.com/juanfont/headscale/issues/3406)
 
 ## 0.29.3 (2026-07-29)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ keys remain all-access.
 
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
+- Creating an OAuth client now rejects scopes outside the known vocabulary instead of storing them verbatim [#3406](https://github.com/juanfont/headscale/issues/3406)
 
 ## 0.29.3 (2026-07-29)
 
--- a/hscontrol/db/oauth.go
+++ b/hscontrol/db/oauth.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/juanfont/headscale/hscontrol/scope"
 	"github.com/juanfont/headscale/hscontrol/types"
 	"golang.org/x/crypto/argon2"
 	"gorm.io/gorm"
@@ -37,6 +38,7 @@ var (
 	ErrOAuthClientNotFound      = fmt.Errorf("oauth client not found: %w", gorm.ErrRecordNotFound)
 	ErrOAuthClientFailedToParse = errors.New("failed to parse oauth client secret")
 	ErrOAuthClientRevoked       = errors.New("oauth client revoked")
+	ErrOAuthClientScopeInvalid  = errors.New("oauth client scope invalid")
 
 	ErrAccessTokenNotFound      = fmt.Errorf("oauth access token not found: %w", gorm.ErrRecordNotFound)
 	ErrAccessTokenFailedToParse = errors.New("failed to parse oauth access token")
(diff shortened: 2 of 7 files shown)


hscontrol/db/oauth.go
```diff
@@ -132,6 +134,39 @@ func verifySecret(encoded []byte, secret string) error {
 	return nil
 }
 
+// validateScopes deduplicates and sorts scopes, rejecting any value outside the
+// vocabulary in [scope.Known]. An unknown scope satisfies no requirement, so
+// storing one would silently mint a client with fewer permissions than asked
+// for.
+func validateScopes(scopes []string) ([]string, error) {
```
kradalby: I think this function should validate the scopes, but not _also_ dedupe and sort. 

It would be great if this would just validate, then build a single multierr (tailscale has one in the lib) so we can return one error which gives the user full feedback instead of having to call again and potentially having more with errors. 

Then keep the dedupe and sort as the call after we have determined there is no errors. Or it can be done before so we have to validate less.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ keys remain all-access.
 
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
+- Creating an OAuth client now rejects scopes outside the known vocabulary instead of storing them verbatim [#3406](https://github.com/juanfont/headscale/issues/3406)
```
kradalby: We dont want to update the changelog since this is a change to an unreleased feature. The feature will be released correctly when it is finished thanks to this PR, so this line should be removed.

kradalby (commented):
Good catch! thank you, two small comments, otherwise good.

MsfPablo:
Both addressed, thanks.

`validateScopes` now only validates and collects every bad scope into one `multierr.New`, so you get all of them back in a single call. The dedupe/sort moved up into `CreateOAuthClient`, after validation. Dropped the changelog line too, and rebased off the conflict.

kradalby (approved):
Great, thanks, can you rebase it?

PR #3414, opened 2026-08-06
fix(metrics): collect metrics for non-OPTIONS requests

PauloLuna:
Fix the inverted HTTP metrics skip condition, which caused the collector to ignore all requests except OPTIONS.

Apply the correction to both the main and Noise routers so regular HTTP traffic is included in http_requests_total and http_request_duration_seconds.

- [x] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated documentation if needed
- [ ] updated CHANGELOG.md

Fixes #3412

PR #3426, opened 2026-08-17
Fix inverted HTTP metrics skip condition

### LINKED ISSUE #3412: [Bug] Inverted HTTP metrics skip condition

### Current Behavior

The HTTP metrics collector skips every request whose method is not OPTIONS. As a result, regular requests such as GET, POST, PUT, and DELETE are not recorded in http_requests_total or http_request_duration_seconds.

Only OPTIONS requests are currently collected.

### Expected Behavior

The collector should skip only OPTIONS requests and collect metrics for all regular HTTP traffic.

This behavior should apply consistently to both the main and Noise routers, ensuring that requests are properly represented in http_requests_total and http_request_duration_seconds.

### Steps To Reproduce

1. Start Headscale v0.29.x with the HTTP and metrics listeners enabled:

   ```yaml
   listen_addr: 127.0.0.1:8080
   metrics_listen_addr: 127.0.0.1:9090
   ```

2. Check the current HTTP metrics:

   ```bash
   curl -s http://127.0.0.1:9090/metrics |
     grep -E 'http_requests_total|http_request_duration_seconds'
   ```

3. Send multiple regular requests to the `/apple` endpoint:

   ```bash
   for i in $(seq 1 10); do
     curl -s http://127.0.0.1:8080/apple >/dev/null
   done
   ```

4. Check the metrics again:

   ```bash
   curl -s http://127.0.0.1:9090/metrics |
     grep -E 'http_requests_total|http_request_duration_seconds'
   ```

5. Observe that the `GET /apple` requests were not recorded.

6. Send an `OPTIONS` request to the same endpoint:

   ```bash
   curl -i -X OPTIONS http://127.0.0.1:8080/apple
   ```

7. Check the metrics again:

   ```bash
   curl -s http://127.0.0.1:9090/metrics |
     grep -E 'http_requests_total|http_request_duration_seconds'
   ```

8. Observe that the `OPTIONS` request was recorded, demonstrating that the skip condition is inverted.

### Environment

```markdown
- OS: N/A — server-side issue independent of the operating system
- Headscale version: 0.29.3
- Tailscale version: N/A — reproducible directly against the Headscale HTTP endpoint
```

### Runtime environment

### Debug information

N/A

harriiinnii:
### Problem

The `Skip` function passed to `metrics.Collector` in both routers had the condition backwards:

```go
Skip: func(r *http.Request) bool {
    return r.Method != http.MethodOptions  // skips everything except OPTIONS
},
```

A `Skip` function that returns `true` tells the middleware to **not** record metrics for that request. With `!=`, the function returns `true` for every method that is not OPTIONS — meaning GET, POST, PUT, DELETE and all other real traffic is silently excluded from `http_requests_total` and `http_request_duration_seconds`. Only `OPTIONS` preflight requests were ever counted.

### Fix

Change `!=` to `==` so OPTIONS (a CORS preflight with no application value in metrics) is the one method skipped, and all other traffic is recorded as intended.

The same inverted condition existed in both the main router (`createRouter` in `app.go`) and the Noise upgrade router (`NoiseUpgradeHandler` in `noise.go`). Both are fixed in this commit.

### Verification

```bash
# send regular requests to any endpoint
for i in $(seq 1 5); do curl -s http://127.0.0.1:8080/health >/dev/null; done

# metrics now show http_requests_total with method="GET"
curl -s http://127.0.0.1:9090/metrics | grep http_requests_total
```

Fixes #3412

### DIFF
--- a/hscontrol/app.go
+++ b/hscontrol/app.go
@@ -447,7 +447,7 @@ func (h *Headscale) createRouter(apiV1Mux, apiV2Mux http.Handler) *chi.Mux {
 		Host:  false,
 		Proto: true,
 		Skip: func(r *http.Request) bool {
-			return r.Method != http.MethodOptions
+			return r.Method == http.MethodOptions
 		},
 	}))
 	r.Use(middleware.RequestID)
--- a/hscontrol/noise.go
+++ b/hscontrol/noise.go
@@ -150,7 +150,7 @@ func (h *Headscale) NoiseUpgradeHandler(
 		Host:  false,
 		Proto: true,
 		Skip: func(r *http.Request) bool {
-			return r.Method != http.MethodOptions
+			return r.Method == http.MethodOptions
 		},
 	}))
 	r.Use(middleware.RequestID)


nblock:
Duplicate #3414

PR #3427, opened 2026-08-17
hscontrol: allow changing a user's display name, email and profile picture

### LINKED ISSUE #2166: [Feature] Setting DisplayName and ProfilePicURL for headscale users
### Use case

Mostly for niceness and completeness. I would like Headscale to know users by name, so that `tailscale whois` (and related solutions such as `tailscale-nginx-auth`) can know the user's display name and profile picture.

The [OIDC rework document](https://docs.google.com/document/d/1X85PMxIaVWDF6T_UPji3OeeUqVBcGj_uHRM5CI-AwlY/edit) mentioned that CLI-based login should be able to populate these fields, so I hope the maintainers won't mind me filing this as an issue so it could be easily trackable (and to show that this is in fact desirable, and bikeshed over the minute details of the implementation).

Additionally, I may try my hand at implementing this. (I don't really know Go that well, but how hard could it be?)

### Description

One or two commands in the Headscale CLI to set the user's display name and the profile picture. Optionally a way to set these fields on user creation.

### Contribution

Exact subcommand/option names subject to bikeshedding.

```console
$ headscale users set -i 1 --display-name "Vika" --profile-pic-url "https://example.com/vika.png"
$ # Additionally, consider the following, to create a user and set their personal data in one step:
$ headscale users create vika --display-name "Vika" --profile-pic-url "https://example.com/vika.png"
```

Alternatively:
```console
$ headscale users set-display-name -i 1 Vika
$ headscale users set-profile-pic-url -i 1 https://example.com/vika.png
```

albedev:
Fixes #2166. Design document posted on the issue as required by CONTRIBUTING.md:
https://github.com/juanfont/headscale/issues/2166#issuecomment-5320904646

## What is missing today

`DisplayName`, `Email` and `ProfilePicURL` already exist end to end — the DB
carries them, `User.TailscaleUserProfile()` fills them, the mapper emits them,
and `headscale users create` accepts all three. The only gap is that they cannot
be changed afterwards: an operator who created a user before deciding on an
avatar has to destroy and recreate it, which is impossible once that user owns
nodes.

## What this adds

**`PUT /api/v1/user/{id}`** with pointer semantics: an absent field is left
alone, a present field is set — including the empty string, which clears it.
Clearing an avatar is a real operation, and GORM's struct-based `Updates()`
silently skips zero values, so the DB layer builds an explicit column map
instead. An empty body is a `400`, not a silent no-op.

`pictureUrl` is validated to be an absolute `http`/`https` URL, server-side and
in the CLI. This value is handed to every client's UI, so `javascript:` and bare
strings are rejected. (This also replaces the existing `url.Parse` check in
`users create`, which accepted almost anything.)

**`headscale users set`**, following the flag style of `rename`/`destroy`:

```console
$ headscale users set -i 1 --display-name "Vika" --picture-url "https://example.com/vika.png"
$ headscale users set -n vika --picture-url ""   # clear
```

`Flags().Changed` is what distinguishes "not passed" from "passed empty".

**OIDC users are refused**, reusing `ErrCannotChangeOIDCUser` exactly as
`RenameUser` already does. `User.FromClaim` rewrites all three fields on every
login — including to `""` when a claim is absent — so a manual edit would be
silently reverted. The provider stays the source of truth.

## Two bugs this surfaced

Both predate this feature and affect `RenameUser` and OIDC re-login as well: any
user mutation was invisible to already-connected nodes until a restart.

1. **Stale `NodeStore`.** `types.Node` embeds the owner and the mapper reads it
   from the snapshot via `NodeView.Owner()`, not from the DB. `UpdateUser` wrote
   the row and left the cached copies alone — there is an explicit
   `TODO(kradalby): We might want to update nodestore with the user data` at the
   site. Now every node of that user is re-pointed at the updated record in a
   single `NodeStore.UpdateNodes` batch.
2. **No map update was emitted.** `UpdateUser` returned only the policy
   manager's change, which is empty when the policy does not reference the user.
   It now returns `change.UserAdded()` — but *only* when a client-visible field
   actually changed, because OIDC calls `UpdateUser` on every login and
   broadcasting a full update to every node on every login would be a serious
   regression.

## Relationship to #3107

#3107 asks for the same capability but is written against the gRPC/protobuf
stack that was removed in #3324; the API is now code-first with huma and the Go
client is generated from OpenAPI via `make client`. This is a fresh
implementation against the current architecture rather than a rebase, and it
also covers the NodeStore/broadcast defects above. Happy to defer if @Razano26
would rather bring theirs forward instead.

## Tests

- `hscontrol/db/users_test.go` — set each field, clear with `""`, leave absent
  fields untouched, reject OIDC users and unknown ids, reject invalid URLs
- `hscontrol/state/user_profile_test.go` — the cached `Owner()` in the NodeStore
  reflects the new values, no broadcast on a no-op, rename covered too
- `hscontrol/types/users_test.go` — `ProfilePicURL` in all three tailcfg shapes
- `hscontrol/apiv1_users_test.go` — golden parity plus 400/404/OIDC guards
- `cmd/headscale/cli/users_test.go` — flag semantics
- `integration/cli_users_test.go` — `TestUserSetCommand`, registered in
  `test-integration.yaml`

Verdicts on this machine:

```
go test -race ./hscontrol/types/... ./hscontrol/db/... ./hscontrol/state/... ./hscontrol/api/...   ok
go test -race ./hscontrol/ -run TestAPIV1                                                          ok
go test ./cmd/headscale/cli/ -run TestSetUserRequestFromFlags                                      ok
golangci-lint run (v2.12.2, the version pinned in flake.nix)                                       0 issues
```

Each of the four commits builds and passes `go vet` on its own. Disabling the
NodeStore refresh makes `TestSetUserProfilePropagatesToNodeStore` and
`TestRenameUserPropagatesToNodeStore` fail, so they are real regression guards.

**Caveat:** the integration suite does not run locally here —
`Dockerfile.tailscale-HEAD` fails to build tailscale from source. I confirmed
this is environmental by running the pre-existing, unmodified
`TestUserCreateCommand`, which fails identically. `TestUserSetCommand` is written
but needs CI to validate.

## Verified end to end

Built this branch, ran it against a real `tailscale` client, and read the profile
the client received from the control server:

```
 id 1 | alice@example.com | DisplayName: Alice Rossi | ProfilePicURL: https://avatars.githubusercontent.com/u/9784476?v=4
```

Changing another user's picture propagated to that already-connected client
without a restart (confirming fix 2), clearing it returned the field to empty,
and `javascript:alert(1)` was rejected:

```
Error: profile picture URL must be an absolute http or https URL: "javascript:alert(1)"
```

## Maintenance

Three string columns and one endpoint, no external dependency and no background
work. The risk area is the NodeStore refresh in `UpdateUser`, which the state
tests cover. I am happy to maintain this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -290,6 +290,7 @@ jobs:
           - TestUserCommand
           - TestUserCreateCommand
           - TestUserCommandValidation
+          - TestUserSetCommand
           - TestDERPVerifyEndpoint
           - TestResolveMagicDNS
           - TestResolveMagicDNSExtraRecordsPath
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ keys remain all-access.
 
 ### Changes
 
+- Add `headscale users set` and `PUT /api/v1/user/{id}` to change the display name, email and profile picture URL of an existing user [#2166](https://github.com/juanfont/headscale/issues/2166)
+- Fix user profile changes (rename, OIDC re-login) not reaching connected nodes until the server restarted [#2166](https://github.com/juanfont/headscale/issues/2166)
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
 
--- a/cmd/headscale/cli/users.go
+++ b/cmd/headscale/cli/users.go
@@ -5,10 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"net/url"
 	"strconv"
 
 	clientv1 "github.com/juanfont/headscale/gen/client/v1"
+	"github.com/juanfont/headscale/hscontrol/types"
 	"github.com/juanfont/headscale/hscontrol/util"
 	"github.com/juanfont/headscale/hscontrol/util/zlog/zf"
 	"github.com/rs/zerolog/log"
@@ -19,6 +19,9 @@ import (
 var (
 	errFlagRequired       = errors.New("--name or --identifier flag is required")
 	errMultipleUsersMatch = errors.New("multiple users match query, specify an ID")
+	errNoUserFieldsToSet  = errors.New(
+		"at least one of --display-name, --email or --picture-url is required",
+	)
 )
 
 func usernameAndIDFlag(cmd *cobra.Command) {
@@ -96,6 +99,11 @@ func init() {
 	usernameAndIDFlag(renameUserCmd)
 	renameUserCmd.Flags().StringP("new-name", "r", "", "New username")
 	mustMarkRequired(renameUserCmd, "new-name")
+	userCmd.AddCommand(setUserCmd)
+	usernameAndIDFlag(setUserCmd)
+	setUserCmd.Flags().StringP("display-name", "d", "", "Display name, empty to clear")
+	setUserCmd.Flags().StringP("email", "e", "", "Email, empty to clear")
+	setUserCmd.Flags().StringP("picture-url", "p", "", "Profile picture URL, empty to clear")
 }
 
 var userCmd = &cobra.Command{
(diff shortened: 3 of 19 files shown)


kradalby:
Thanks for having a look at this.

The release now focuses on implementing a v2 api that has the same shape as Tailscale's API, I am not certain yet, but the goal is to _not_ make any changes to our v1 API and focus on adopting the Tailscale API where it fits. 
If there are things that cannot be implemented using the Tailscale API, then we will have to rethink a bit, but I am not sure yet if we will go the route of extending v1 or something else. 

Feel free to investigate if this would be possible to implement with [Tailscale's API](https://tailscale.com/api-docs), tho I think it might be challenging.

PR #3430, opened 2026-08-19
build: update tailscale.com to v1.103.0-pre.0.20260819232550-0e84b4a3a0a0

sfllaw:
PR tailscale/tailscale#20646 changed the representation of DERP region IDs from plain ints to a tailcfg.DERPRegionID type. This patch requires that version of the tailscale.com library and updates our code to also use the new type.

Updates: tailscale/tailscale#20165

- [X] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [ ] raised a GitHub issue or discussed it on the projects chat beforehand
- [ ] added unit tests
- [ ] added integration tests
- [ ] updated documentation if needed
- [ ] updated CHANGELOG.md

PR #3433, opened 2026-08-22
Resync the GivenName DNS label when a node re-registers

### LINKED ISSUE #3432: [Bug] GivenName is not resynced when a node re-registers with a changed hostname

### Current Behavior

`GivenName` is resynced from a changed client hostname on the **MapRequest** path, but not on the **register/reauth** path. Because the reauth path already overwrites `node.Hostinfo`, the later MapRequest never observes a hostinfo change, so the resync that exists never runs. The node keeps its old DNS label indefinitely.

`hscontrol/state/state.go`, `UpdateNodeFromMapRequest` — the resync exists and is correctly guarded against clobbering an admin rename:

```go
if hostinfoChanged {
    currentNode.Hostinfo = req.Hostinfo
    if req.Hostinfo != nil && req.Hostinfo.Hostname != "" {
        // Preserve an admin-renamed GivenName: only auto-derive when the
        // current GivenName is still what SanitizeHostname of the old
        // Hostname would produce (possibly with a "-N" collision bump).
        autoDerived := isAutoDerivedGivenName(currentNode.GivenName, currentNode.Hostname)

        currentNode.Hostname = req.Hostinfo.Hostname
        if autoDerived {
            currentNode.GivenName = dnsname.SanitizeHostname(req.Hostinfo.Hostname)
        }
    }
    ...
```

`hscontrol/state/state.go`, `applyAuthNodeUpdate` — same field, no resync:

```go
updatedNodeView, ok := s.nodeStore.UpdateNode(params.ExistingNode.ID(), func(node *types.Node) {
    node.NodeKey = regData.NodeKey
    node.DiscoKey = regData.DiscoKey
    node.Hostname = params.Hostname          // <-- hostname updated

    node.Hostinfo = params.ValidHostinfo     // <-- hostinfo updated too
    ...
    // GivenName is never considered
})
```

The second assignment is what makes this permanent rather than merely delayed. `UpdateNodeFromMapRequest` computes `hostinfoChanged` by comparing the incoming `Hostinfo` against the stored one. Reauth has already stored the new `Hostinfo`, so the first MapRequest after reauth sees no change, takes the early-return branch, and the `autoDerived` block is never reached.

Net effect: `node.Hostname` and `node.GivenName` diverge permanently for any node that re-registers under a new hostname. MagicDNS keeps serving the old label (`GetFQDN()` uses `GivenName`), so the name a device answers to and the name it reports are different, with no warning surfaced anywhere.

This is self-latching. Once the two fields diverge, `isAutoDerivedGivenName(GivenName, Hostname)` returns false forever, so even a subsequent genuine MapRequest hostname change will no longer resync. The anti-clobber guard cannot distinguish "admin renamed this" from "reauth desynced this", and treats the desync as an intentional rename.

### Expected Behavior

`applyAuthNodeUpdate` applies the same `isAutoDerivedGivenName` guard that `UpdateNodeFromMapRequest` already applies: when the stored `GivenName` is still auto-derived from the stored `Hostname`, re-derive it from the incoming hostname; when it has been admin-renamed, leave it alone.

Both paths mutate the same two fields for the same reason, so they should agree on the policy.

### Steps To Reproduce

1. Register a node with hostname `host-a`. `Hostname` and `GivenName` are both `host-a`.
2. Change the client's hostname to `host-b` and re-register with an auth key (`tailscale up --force-reauth`, or logout/login). This is the normal path for automated (re)provisioning, where a device is re-imaged or renamed and re-enrolled with the same machine key and user.
3. `headscale nodes list` shows `Hostname=host-b`, `GivenName=host-a`.
4. `host-a.<base_domain>` still resolves; `host-b.<base_domain>` does not.
5. Wait for any number of MapRequests. The state never converges — step 2 already wrote the new `Hostinfo`, so `hostinfoChanged` is false from then on.

### Environment

```markdown
- Headscale version: v0.29.3 (also present on main @ 3390-line hscontrol/state/state.go)
- Tailscale version: multiple, 1.60-1.80
```

### Runtime environment

### Anything else?

**Proposed fix** — mirror the existing MapRequest logic in `applyAuthNodeUpdate`:

```go
updatedNodeView, ok := s.nodeStore.UpdateNode(params.ExistingNode.ID(), func(node *types.Node) {
    node.NodeKey = regData.NodeKey
    node.DiscoKey = regData.DiscoKey

    // Keep GivenName in sync with a client-side hostname change, using the
    // same rule as UpdateNodeFromMapRequest: only re-derive when the current
    // GivenName is still auto-derived, so an admin rename is preserved.
    if params.Hostname != "" && params.Hostname != node.Hostname {
        if isAutoDerivedGivenName(node.GivenName, node.Hostname) {
            node.GivenName = dnsname.SanitizeHostname(params.Hostname)
            // NodeStore.UpdateNode auto-bumps GivenName on collision.
        }
    }

    node.Hostname = params.Hostname
    ...
```

`NodeStore.UpdateNode` already bumps on collision via `resolveGivenName`, so the uniqueness constraint on `GivenName` is handled by the existing machinery.

**Note on existing deployments.** Nodes that have already desynced cannot be repaired by this fix alone, because the guard now reads them as admin-renamed. A detection helper would let operators find them:

```go
// Nodes where GivenName is neither auto-derived from the current Hostname
// nor from any plausible earlier one are indistinguishable from an admin
// rename, so surfacing them is the most that can be done automatically.
```

Two options that would each be useful on their own:

1. Extend the existing `hscontrol/state/node_health.go` checks with a `givenNameMatchesHostname` warning, alongside the current `givenNameMapsToValidFQDN`. Purely advisory, no behaviour change, and it makes the drift visible without an operator having to diff two columns by hand across the whole tailnet.
2. Add a `--reset-given-name` flag to `headscale nodes rename` (or a distinct subcommand) that re-derives `GivenName` from the current `Hostname`, so the repair does not require the operator to retype a name that headscale already knows.

Happy to open a PR for the `applyAuthNodeUpdate` change plus a regression test asserting that reauth with a changed hostname updates `GivenName` when auto-derived and preserves it after an explicit rename. Let me know if the health-check or CLI additions are wanted in the same PR or separately.

yzxcj797:
## What?

A node that re-registers under a new hostname keeps serving its old MagicDNS label forever: `GivenName` is resynced from the hostname on the **MapRequest** path, but not on the **register/re-auth** paths.

Closes #3432

## Why?

Both re-registration paths overwrite the node's hostname and hostinfo in one step, and neither touches `GivenName`:

- `applyAuthNodeUpdate` sets `node.Hostname = params.Hostname` and `node.Hostinfo = params.ValidHostinfo`;
- `HandleNodeFromPreAuthKey`'s in-place update does the same (and already carries a TODO asking for the map-request merge semantics).

The resync that exists in `UpdateNodeFromMapRequest` is guarded by `hostinfoChanged` — but re-registration has already stored the new `Hostinfo`, so the first MapRequest afterwards compares equal, takes the early-return branch, and never reaches the `autoDerived` block. The divergence is self-latching: once `Hostname` and `GivenName` differ, `isAutoDerivedGivenName` returns false forever, so even the MapRequest path stops helping. `GetFQDN()` uses `GivenName`, so MagicDNS keeps answering the old name indefinitely.

## How?

Mirror `UpdateNodeFromMapRequest`'s GivenName handling on both register paths, inside the existing `NodeStore.UpdateNode` mutation (so the collision-bump logic in `UpdateNode` applies as it does elsewhere):

```go
givenNameAutoDerived := isAutoDerivedGivenName(node.GivenName, node.Hostname)
node.Hostname = <new hostname>
if givenNameAutoDerived && <new hostname> != "" {
    node.GivenName = dnsname.SanitizeHostname(<new hostname>)
}
```

The `isAutoDerivedGivenName(node.GivenName, node.Hostname)` check runs against the **old** hostname (before it is overwritten), exactly as the MapRequest block does, so an admin rename keeps winning on every path. The empty-hostname guard matches the MapRequest block's `Hostname != ""` condition (relevant for the pre-auth-key path, where a client can omit hostinfo entirely).

## Testing

`TestReauthResyncsGivenName` (new, `hscontrol/state`), covering both paths and the rename guard:

- a node registered as `old-host`, re-registered via `HandleNodeFromAuthPath` as `new-host`: `GivenName` becomes `new-host`;
- an admin `SetGivenName("custom-label")` survives both a subsequent re-auth **and** a pre-auth-key re-registration under yet another hostname;
- a second node with an auto-derived label, re-registered via `HandleNodeFromPreAuthKey` as `pak-host`: `GivenName` becomes `pak-host`.

**Differential**: on the previous code the test fails with `#3432: GivenName must follow the new hostname when re-registering via re-auth`. Full `go test ./hscontrol/state/` (56s, the whole package) passes.

### DIFF
--- a/hscontrol/state/auth_givenname_resync_test.go
+++ b/hscontrol/state/auth_givenname_resync_test.go
@@ -85,7 +85,7 @@ func TestReauthResyncsGivenName(t *testing.T) {
 
 	// The rename survives pre-auth-key re-registration too.
 	pakRegRenamed := tailcfg.RegisterRequest{
-		Auth:    &tailcfg.RegisterResponseAuth{AuthKey: pakKeyFor(s, t, user)},
+		Auth:    &tailcfg.RegisterResponseAuth{AuthKey: pakKeyFor(t, s, user)},
 		NodeKey: key.NewNode().Public(),
 		Expiry:  clientExpiry,
 		Hostinfo: &tailcfg.Hostinfo{
@@ -100,7 +100,7 @@ func TestReauthResyncsGivenName(t *testing.T) {
 
 	// --- Pre-auth-key re-registration on a node with an auto-derived label ---
 	pakReg := tailcfg.RegisterRequest{
-		Auth:    &tailcfg.RegisterResponseAuth{AuthKey: pakKeyFor(s, t, user)},
+		Auth:    &tailcfg.RegisterResponseAuth{AuthKey: pakKeyFor(t, s, user)},
 		NodeKey: key.NewNode().Public(),
 		Expiry:  clientExpiry,
 		Hostinfo: &tailcfg.Hostinfo{
@@ -115,7 +115,7 @@ func TestReauthResyncsGivenName(t *testing.T) {
 		"#3432: GivenName must follow the new hostname when re-registering via pre-auth key")
 }
 
-func pakKeyFor(s *State, t *testing.T, user *types.User) string {
+func pakKeyFor(t *testing.T, s *State, user *types.User) string {
 	t.Helper()
 
 	userID := types.UserID(user.ID)
--- a/hscontrol/state/auth_givenname_resync_test.go
+++ b/hscontrol/state/auth_givenname_resync_test.go
@@ -1,0 +1,126 @@
+package state
+
+import (
+	"testing"
+	"time"
+
+	"github.com/juanfont/headscale/hscontrol/db"
+	"github.com/juanfont/headscale/hscontrol/types"
+	"github.com/stretchr/testify/require"
+	"tailscale.com/types/key"
+	"tailscale.com/tailcfg"
+)
+
+// TestReauthResyncsGivenName is the reproduction for #3432: the re-auth and
+// pre-auth-key re-registration paths overwrite Hostname (and Hostinfo) in one
+// step, so the hostinfoChanged resync in UpdateNodeFromMapRequest can never
+// fire afterwards — the GivenName (and with it the MagicDNS label) kept the
+// old hostname forever. Admin renames must keep winning over both paths.
+func TestReauthResyncsGivenName(t *testing.T) {
+	dbPath := t.TempDir() + "/headscale.db"
+	cfg := persistTestConfig(dbPath)
+
+	database, err := db.NewHeadscaleDatabase(cfg)
+	require.NoError(t, err)
+
+	user := database.CreateUserForTest("givenname-user")
+	node := database.CreateRegisteredNodeForTest(user, "old-host")
+	machineKey := node.MachineKey
+
+	pakNode := database.CreateRegisteredNodeForTest(user, "old-pak-host")
+	pakMachineKey := pakNode.MachineKey
+
+	require.NoError(t, database.Close())
+
+	s, err := NewState(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = s.Close() })
+
+	clientExpiry := time.Now().Add(24 * time.Hour)
+
+	// --- Interactive/web re-auth path (applyAuthNodeUpdate) ---
+	registrationID := types.MustAuthID()
+	regEntry := types.NewRegisterAuthRequest(&types.RegistrationData{
+		MachineKey: machineKey,
+		NodeKey:    key.NewNode().Public(),
+		Hostname:   "new-host",
+		Hostinfo: &tailcfg.Hostinfo{
+			Hostname: "new-host",
+		},
+		Expiry: &clientExpiry,
+	})
+	s.SetAuthCacheEntry(registrationID, regEntry)
+
+	afterReauth, _, err := s.HandleNodeFromAuthPath(
+		registrationID, types.UserID(user.ID), nil, "webauth",
+	)
+	require.NoError(t, err)
+	require.Equal(t, "new-host", afterReauth.Hostname())
+	require.Equal(t, "new-host", afterReauth.GivenName(),
+		"#3432: GivenName must follow the new hostname when re-registering via re-auth")
+
+	// --- Admin renames survive re-registration ---
+	_, err = s.nodeStore.SetGivenName(afterReauth.ID(), "custom-label")
+	require.NoError(t, err)
+
+	registrationID2 := types.MustAuthID()
+	regEntry2 := types.NewRegisterAuthRequest(&types.RegistrationData{
+		MachineKey: machineKey,
+		NodeKey:    key.NewNode().Public(),
+		Hostname:   "another-host",
+		Hostinfo: &tailcfg.Hostinfo{
+			Hostname: "another-host",
+		},
+		Expiry: &clientExpiry,
+	})
+	s.SetAuthCacheEntry(registrationID2, regEntry2)
+
+	afterRename, _, err := s.HandleNodeFromAuthPath(
+		registrationID2, types.UserID(user.ID), nil, "webauth",
+	)
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


yzxcj797:
The garnix failures on the last two pushes (gotest, golangci-lint, and everything downstream) are not from this change — the go-modules derivation dies downloading dependencies inside the sandbox:

```
go: ... github.com/klauspost/compress@v1.18.6: reading https://proxy.golang.org/github.com/klauspost/compress/@v/v1.18.6.zip: 403 Forbidden
go: ... modernc.org/sqlite@v1.52.0: reading https://proxy.golang.org/modernc.org/sqlite/@v/v1.52.0.zip: 403 Forbidden
error: build of '/nix/store/...-go-modules.drv' on 'ssh-ng://nix-ssh@gigabuilder' failed
```

proxy.golang.org is 403ing the builder for uncached modules (same signature on #3430, which also touches the module set, while PRs whose go-modules derivation is already cached stay green). This branch does not touch go.mod/go.sum.

Local verification: `go build ./...` clean, `CGO_ENABLED=0 go test -skip TestPostgres` over the garnix test set passes (the only local failures are `TestConfigLoading`/`TestConfigFileLoading`, which need Windows symlink privileges and are unrelated).

kradalby:
Garnix failure has been fixed if you rebase of main.

This _sounds_ reasonable, but we need to validate what behaviour the Tailscale SaaS has for this kind of scenario. Would you be willing to validate that?

Moep90:
Reviewed at 581cfd3 as the reporter of #3432. The fix itself is what I proposed and `TestReauthResyncsGivenName` passes locally, including the pre-auth-key path, which I had not covered in the issue. Two things unrelated to the logic.

**1. The diff deletes an unrelated golden file and breaks a passing test.**

`hscontrol/testdata/apiv1_golden/TestAPIV1ApiKeyDelete_path_prefix_with_id_query_is_both_->_400_parity.json` is removed. `assertAgainstGolden` hard-fails on a missing golden rather than regenerating, so:

```
$ go test ./hscontrol/ -run TestAPIV1ApiKeyDelete -count=1     # main
ok      github.com/juanfont/headscale/hscontrol 0.237s

$ go test ./hscontrol/ -run TestAPIV1ApiKeyDelete -count=1     # 581cfd3
--- FAIL: TestAPIV1ApiKeyDelete/path_prefix_with_id_query_is_both_->_400_parity
    missing golden testdata/apiv1_golden/TestAPIV1ApiKeyDelete_path_prefix_with_id_query_is_both_->_400_parity.json
    for DELETE /api/v1/apikey/... — run with HEADSCALE_UPDATE_GOLDEN=1 to generate
FAIL
```

Looks like a stray `HEADSCALE_UPDATE_GOLDEN` run picking up a filename with `>` in it. Restoring the file should be the whole fix.

**2. The new doc comment orphans the existing one.**

`state.go:3001` onwards is now a single comment block, with the `isAutoDerivedGivenName` text ending up on `resyncAutoDerivedGivenName` and `isAutoDerivedGivenName` left undocumented:

```go
// isAutoDerivedGivenName reports whether given matches what
// ... overwritten by client-side hostname changes).
// resyncAutoDerivedGivenName mirrors the GivenName handling of
// ...
func resyncAutoDerivedGivenName(node *types.Node, newHostname string) {
```

Needs a blank line and the original comment moved back down onto `isAutoDerivedGivenName`.

I ran the Tailscale SaaS validation kradalby asked for and posted the full results on #3432. Short version, since it affects how this PR gets justified rather than whether it is correct: SaaS misses the resync on the reauth path exactly like headscale does, but converges on the next client start. So this PR converges earlier than SaaS and lands on the same end state, which is a better argument for it than the one I originally wrote in the issue.

Issue #2289, opened 2024-12-14
[Bug] Filewatcher stops watching dns.extra_records_path


nblock:
### Is this a support request?

- [X] This is not a support request

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A read-edit-write cycle (with vim) causes headscale to lose track of the path in `dns.extra_records_path`.

### Expected Behavior

Pickup changes made to the file. Headscale logs detected writes with `log.level: trace` when different versions of the config file are written. The sequence:

```bash
cp extra1.json .headscale-dns/extra.json
cp extra2.json .headscale-dns/extra.json
cp extra2.json .headscale-dns/extra.json    # no changes, OK
```

yields (fine)
```
2024-12-14T16:39:47+01:00 TRC ../runner/work/headscale/headscale/hscontrol/dns/extrarecords.go:87 > extra records received filewatch event op=WRITE path=.headscale-dns/extra.json
2024-12-14T16:39:47+01:00 TRC ../runner/work/headscale/headscale/hscontrol/dns/extrarecords.go:134 > extra records updated from path, count old: 1, new: 1 records=[{"Name":"grafana.myvpn.example.com","Type":"A","Value":"100.64.0.1"}]
2024-12-14T16:39:56+01:00 TRC ../runner/work/headscale/headscale/hscontrol/dns/extrarecords.go:87 > extra records received filewatch event op=WRITE path=.headscale-dns/extra.json
2024-12-14T16:39:56+01:00 TRC ../runner/work/headscale/headscale/hscontrol/dns/extrarecords.go:134 > extra records updated from path, count old: 1, new: 1 records=[{"Name":"grafana.myvpn.example.com","Type":"A","Value":"100.64.0.2"}]
2024-12-14T16:40:01+01:00 TRC ../runner/work/headscale/headscale/hscontrol/dns/extrarecords.go:87 > extra records received filewatch event op=WRITE path=.headscale-dns/extra.json
```

### Steps To Reproduce

The following causes headscale to loose track of the file:
```bash
vim .headscale-dns/extra.json    # make some valid changes and save
```

yields sometimes:
```
2024-12-14T16:50:21+01:00 TRC ../runner/work/headscale/headscale/hscontrol/dns/extrarecords.go:87 > extra records received filewatch event op=RENAME path=.headscale-dns/extra.json
[...]

PR #3437, opened 2026-08-24
dns: cancel extra-records retry on shutdown

SebTardif:
When `dns.extra_records_path` is deleted or atomically replaced, the filewatcher retry can ignore shutdown for up to 15 minutes. This change cancels that retry on `Close` and closes the watcher if setup fails.

## Problem

`NewExtraRecordsManager` starts an `fsnotify` watcher and `Run` re-attaches after `Remove`/`Rename` with `backoff.Retry`. That retry used `context.Background()`, so `Close()` (called from `Serve` shutdown) could not abort it. backoff v5's default budget is 15 minutes.

If the file is still missing after that budget, the old loop continued without `watcher.Add`. A later recreate of the same path was never seen. That is a delayed return of the watch-loss bug fixed in #2298 for editor atomic saves.

Setup errors after `fsnotify.NewWatcher()` also returned without `watcher.Close()`, leaking the watcher FD and its internal goroutine.

## Change

- Cancel the path-exists retry when `closeCh` is closed.
- After a failed retry that is not shutdown, watch the parent directory so a later recreate is observed.
- Close the watcher on every `NewExtraRecordsManager` error path after `NewWatcher` succeeds.

## Validation

- Red: `go test ./hscontrol/dns/ -run TestWaitUntilPathExistsReturnsOnClose` failed in 2s without the cancel (`waitUntilPathExists did not return after Close`).
- Green: `go test -race ./hscontrol/dns/` passes.
- `gofmt` and `golangci-lint run ./hscontrol/dns/` are clean.

## Checklist

- [x] added unit tests
- [x] updated CHANGELOG.md

Ref #2289

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ keys remain all-access.
 - Expiring or deleting a non-existent pre-auth key now returns an error instead of silently succeeding [#3324](https://github.com/juanfont/headscale/pull/3324)
 - Improve systemd service file hardening [#3341](https://github.com/juanfont/headscale/pull/3341)
 - Headscale now requires Go 1.27 to build
+- Fix extra-records filewatcher hanging on shutdown after the watched file is deleted, and leaking the watcher when setup fails
 
 ## 0.29.3 (2026-07-29)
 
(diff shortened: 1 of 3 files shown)


kradalby (approved):
Looks reasonable, ignore garnix for now, it looks like it took a holiday too.

kradalby:
Could you rebase this off main?

SebTardif:
@kradalby

> Could you rebase this off main?

Rebased onto current `main` (`63123196`). The only conflict was `CHANGELOG.md`; the Go 1.27 note and the extra-records line are both kept. Local `go test -race ./hscontrol/dns/...` still passes.

SebTardif:
@kradalby

Integration failed once on `TestHASubnetRouterFailoverDockerDisconnect` (phase 5a, primary flap to an offline router; same class as #3203). That path is not the extra-records change. Pushed an empty commit to retrigger, since fork PRs cannot `gh run rerun --failed`.

Issue #3441, opened 2026-08-30
[Bug] renaming users by name does not work


lukas-runge:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When using `headscale users rename -n "current_username" -r new_username` the operation always fails with: `Error: renaming user: rpc error: code = Unknown desc = user not found`.

### Expected Behavior

The user called `current_username` shall be found and renamed to `new_username`.

### Steps To Reproduce

1. create new user called `current_username`
2. try to rename `current_username` by name to `new_username`

### Environment

```markdown
- OS: Ubuntu 24.04
- Headscale version: 0.29.3
- Tailscale version: 1.102.3
```

### Runtime environment

- [x] Headscale is behind a (reverse) proxy
- [x] Headscale runs in a container

### Debug information

There is no additional debug information required to understand or test this issue.

PR #3442, opened 2026-08-30
cli: fix users rename when resolved by name

lukas-runge:
`resolveSingleUser` returned the raw `--identifier` flag value instead of
the identifier of the matched user. Renaming with `--name` therefore sent
`OldId=0` to the API and failed with "user not found".

This PR returns the matched user's identifier from `resolveSingleUser` and
passes it straight to the rename endpoint.

fixes #3441

>**Transparency note:** This fix was generated with AI (GLM-5.3-Flash),
not hand-written. So don't assume someone has already stared at every line.
Happy to rework anything that doesn't hold up under review. 🙌

- [x] have read the [CONTRIBUTING.md](./CONTRIBUTING.md) file
- [x] raised a GitHub issue or discussed it on the projects chat beforehand
- [x] added unit tests
- [x] updated CHANGELOG.md

PR #3445, opened 2026-09-01
Fix: [Bug] renaming users by name does not work

goransh-walia:
This PR addresses #3441.

[Bug] renaming users by name does not work

Generated with AI assistance and validated against the original
file before submission (syntax check + change-scope check).
Please review carefully — happy to adjust based on feedback.


nblock:
Duplicate #3442.

Please have a look at open PRs before submission.

PR #3448, opened 2026-09-01
oidc: serve the registration confirmation page from a reloadable URL

### LINKED ISSUE #3365: [Bug] OIDC login issues after upgrading 0.28.0->0.29.1 with Ad Block Plus

### Current Behavior

With Ad Block Plus enabled, after upgrading to 0.29.x, interactive OIDC login fails with a "Forbidden — You are not authorized" (or "Bad Request") page; the node is never registered. There's a brief flash of the confirmation page before the error replaces it.  Disabling ABP allows the login to succeed.

AI suggests that this can probably be considered issues both on the ABP side (they probably shouldn't reload a page with query parameters, though they have reasons for doing it), but also on the Headscale side: serving a single-use OAuth code at URL that the user sits on.  The back button, pull-to-refresh, and browser prerendering would all break it.

/oidc/callback is hit twice on the same connection (simplified log messages):

    GET /oidc/callback   200   (~200ms — exchanges the code, renders the confirmation page)
    ERR user msg: state not found  code=400
    GET /oidc/callback   400   (<1ms)

It's not cache or prefetch. Comparing the two requests (DevTools → Copy as cURL):

|  | 1st hit (→200, real) | 2nd hit (→400, spurious) |
| --- | --- | --- |
| `Referer` | the IdP (`accounts.google.com`) | *(none)* |
| `Sec-Fetch-Site` | `cross-site` | `same-origin` |
| `Sec-Fetch-User` | `?1` (user gesture) | *(none)* |
| state cookie | present | gone (1st hit cleared it) |

The first hit is the genuine callback; the second is a same-origin, no-referrer, no-user-gesture navigation to the same URL — a programmatic reload of the confirmation page. The reloader is Adblock Plus (its content script caches an element-hiding value in sessionStorage then calls window.location.reload()); disabling ABP fixes it, and the DevTools Initiator points at ABP's content.js.

Root cause: the interstitial is rendered inline at /oidc/callback, which carries the single-use OAuth code. Reloading re-hits the callback, re-attempts the spent exchange, and the error page paints over the interstitial. The back button, mobile pull-to-refresh, and prerendering would all trigger it too. New in 0.29 (interstitial #3180 + styled error page #3182).

Suggested fix

Post/Redirect/Get: the callback exchanges the code, stores the pending confirmation, and 303-redirects to a new GET /register/confirm/{auth_id} that renders the (reloadable) interstitial; confirm cookie → SameSite=Lax to survive the redirect (CSRF still enforced by the double-submit token); and /oidc/callback answers 204 No Content on Sec-Fetch-Site: same-origin as defense-in-depth. Implemented and tested; PR to follow.

Claude has implemented this fix and I have verified it works with my setup, it is available on my branch "oidc-confirm-redirect" in my fork at https://github.com/linsomniac/headscale/tree/oidc-confirm-redirect

LMK if you want me to turn that into a PR, but I'm guessing that may be more of a implementation guide than a ready-to-submit PR; I'm not familiar with golang, OIDC, or headscale enough to be able to talk intelligently about it.

### Expected Behavior

Ideally, users with Ad Block Plus would be able to login via OIDC.

### Steps To Reproduce

1. Load Ad Block Plus
2. In tailscale client do "Renew Session..." or other similar way of logging in.

I've reproduced this both on MacOS with the Arc browser and Linux with the Zen browser.

### Environment

```markdown
- OS: Ubuntu 24.04 for server
- Headscale version: 0.23.1 (from deb package), configured to use Google OIDC
- Tailscale version: 1.98.5 on MacOS and NixOS.
- Number of nodes: ~200
```

### Runtime environment

### Debug information

No additional debug information, I believe I've fully debugged it above.

kradalby:
Fixes #3365


kradalby:
The `build` check is red on a blocker that is not this PR's, and it gates the whole integration matrix (all `sqlite`/`postgres` jobs show as skipping).

`Dockerfile.tailscale-HEAD:7` pins `golang:1.27.0-alpine` and then clones tailscale HEAD, whose `go.mod` now requires `go >= 1.27.1`:

```
#14 0.080 go: go.mod requires go >= 1.27.1 (running go 1.27.0; GOTOOLCHAIN=local)
ERROR: failed to solve: process "/bin/sh -c GOARCH=$TARGETARCH go install ..." exit code: 1
```

This same check passed on this exact tree earlier today — the last push changed only commit message text, not a byte of code (`git diff` between the two heads is empty). Nothing in the diff touches `Dockerfile.tailscale-HEAD`, and the failure is deterministic, so a re-run will not clear it. Every PR in the repo will hit it until the pin moves.

Proposed patch, same shape as `docker: head tailscale latest go`:

```diff
-FROM golang:1.27.0-alpine AS build-env
+FROM golang:1.27.1-alpine AS build-env
```

Verified locally: `docker build -f Dockerfile.tailscale-HEAD .` fails on the current pin and succeeds with `1.27.1-alpine`. Leaving it out of this PR rather than widening it — happy to port it here or send it separately.

juanfont (approved):

PR #3450, opened 2026-09-02
state,mapper: classify map requests and reuse peer adjacency

### LINKED ISSUE #3417: [Bug] 0.29 batcher emits reason="node added" for long-existing nodes, fanning a full netmap rebuild to every connected peer
### Description

After upgrading from 0.28.0 to 0.29.3, the control plane's steady-state CPU roughly tripled and has kept climbing. The debug log shows the batcher emitting `reason="node added"` changes for nodes that have existed for months, at a rate roughly 8–14x the number of actual client connections. Each one appears to trigger a netmap rebuild fanned out to every connected peer.

At our node count this saturates the server: map delivery to the subnet routers no longer completes inside the 5s HA probe timeout, both routers are marked unresponsive, and the primary never fails over — so subnet routes black-hole.

### Environment

- headscale 0.29.3 (previously 0.28.0), official image
- SQLite, single replica
- 617 node records, ~130 online, ~115 with an active long-poll connection
- ~139 advertised subnet routes across two HA subnet routers
- policy v2, ACLs + `autoApprovers`

### What I observe

90 seconds of debug log, 6648 lines. Change reasons:

```
5721  reason="node added"
 696  reason="node online"
  13  reason="ping node"
   2  reason="policy change"
   2  reason="node offline"
```

Actual client connections in that same 90s — `poll.go > node added to batcher`:

```
6
```

`node online` reconciles exactly: 6 connections × 116 connected peers = 696. That path behaves as I'd expect.

`node added` does not. 5721 fan-out lines from `batcher.go:168`, across **115 distinct recipient node IDs** — i.e. every connected peer. The most-visible peer received **86** separate `node added` changes in those 90 seconds, against 6 real connections.

Sample lines:

```
DBG hscontrol/mapper/batcher.go:168 > node change processing started node.id=896 reason="node added"
DBG hscontrol/mapper/batcher.go:168 > node change processing started node.id=670 reason="node added"
```

Crucially, these fire for nodes that were not added. Three of the most frequent recipients:

| node.id | created |
|---|---|
| 420 | 2025-08-08 |
| 646 | 2026-02-07 |
| 955 | 2026-08-10 |

Node 420 has existed for a year and is still receiving `reason="node added"`. My reading is that a routine nodestore write — `last_seen`, endpoint refresh — is being classified as a node addition rather than an update, and takes the full-rebuild path instead of a cheap one.

### Cost

From `/metrics`, on a process ~190s old:

```
headscale_nodestore_nodes                                    617
headscale_nodestore_peers_calculation_duration_seconds_sum   189.1
headscale_nodestore_peers_calculation_duration_seconds_count 23
headscale_nodestore_snapshot_build_duration_seconds_sum      189.2
headscale_nodestore_snapshot_build_duration_seconds_count    23
headscale_nodestore_batch_size_sum                           468
headscale_nodestore_batch_size_count                         23
headscale_nodestore_queue_depth                              6
process_cpu_seconds_total                                    529.6
```

8.2s per peer calculation, snapshot build accounting for essentially all of it, queue depth never draining, and 529 CPU-seconds burned in 190s of wall time.

### Downstream failures

```
WRN send: connection failed error="connection 99: timeout sending to channel (likely stale connection)" conn.id=99 node.id=875
ERR failed to apply change error="sending map response to node 875: node 875: all connections failed, last error: connection 99: timeout sending to channel (likely stale connection)" node.id=875 reason="node added" worker.id=2
WRN HA probe: node did not respond node.id=955 timeout=5000
WRN HA probe: node did not respond node.id=646 timeout=5000
```

Both nodes failing the HA probe are the two HA subnet routers. They aren't down — the batcher can't reach them in time.

### When it started

Container CPU, 6-hour buckets, cores, across the upgrade:

```
Jul 26-31  nights 0.50-0.65   days 0.83-1.11     0.28.0
Aug 1-2    0.36-0.52                             weekend
Aug 3      0.37 -> 0.76 -> 1.10 -> 1.02          upgraded to 0.29.3
Aug 4      0.98  1.45  1.81  1.35
Aug 5      1.07  1.90  2.22  1.64
Aug 6      1.27  1.98  2.23  1.77
Aug 7      1.54  2.07  2.46  1.51
Aug 8-9    1.25-1.43 flat                        weekend
Aug 10     1.32 -> 2.25 -> 3.07
```

The overnight floor is the clearest signal: with almost nothing connected it was 0.50–0.65 before the upgrade and has never dropped below 1.25 since. Client count overnight didn't change, so this is per-change cost, not load. Nothing else changed on Aug 3.

### Expected

A `last_seen` or endpoint update on an existing node should not be classified as `node added`, and should not force a full snapshot rebuild fanned out to every connected peer.

### Notes

I saw #3293 (`0.29 beta triggers PolicyChange on every reconnect causing CPU spikes`) and this looks like a different path — our `policy change` count is 2 per 90s, so that fix is working. Happy to supply more log or metrics detail.

kradalby:
Every MapRequest, LastSeen-only keepalives included, rebuilt the O(N²) peer map in `snapshotFromNodes` and returned `change.NodeAdded`, re-sending the whole node to every peer.

First ten commits are patch-shaped and apply to `release-branch/0.29` on their own: classify each request into the narrowest change, drop empty changes at the batcher, skip health probes that change nothing, fix two policy comparisons. No data structure touched.

Last four key the peer map by node ID and reuse adjacency for writes that cannot change peer visibility. That removes the rebuild, and changes `PeersFunc` and `BuildPeerMap`, so 0.30.0 only.

Changelog split to match.

Fixes #3417


kradalby:
The `build` check in the integration workflow was failing, and it was not caused by the map-request change.

`Dockerfile.tailscale-HEAD` pinned `golang:1.27.0-alpine`, but tailscale HEAD now declares `go 1.27.1` in its `go.mod`, so the image build stopped before any test ran:

```
go: go.mod requires go >= 1.27.1 (running go 1.27.0; GOTOOLCHAIN=local)
```

That blocked the `sqlite`, `postgres`, `build-postgres` and `build-tailscale-released` jobs, which all reported `skipping` — including the HA failover tests most worth running against this change. Folded the bump in so integration actually runs here:

```diff
-FROM golang:1.27.0-alpine AS build-env
+FROM golang:1.27.1-alpine AS build-env
```

Only `Dockerfile.tailscale-HEAD` builds tailscale at HEAD. `Dockerfile.derper` clones a released branch, and `Dockerfile.integration` and `Dockerfile.wasmclient` build headscale's own code, so none of them need it.

kradalby:
Integration run on the previous head: 175 passed, 2 failed.

`TestDERPVerifyEndpoint` was the same Go pin problem in a second file. `integration/dsic/dsic.go:230` maps DERP version `head` to `VERSION_BRANCH=main`, so `Dockerfile.derper` clones tailscale HEAD as well and fails identically at `go install ./cmd/derper`. Bumped it in the same commit. Verified locally in both directions: with `1.27.0` and the build stage cache busted the build reproduces `go: go.mod requires go >= 1.27.1`, and with `1.27.1` the image builds and the resulting `derper` starts.

`TestHASubnetRouterFailoverDockerDisconnect` fails the same way on `main` — `integration/route_test.go:4419`, `Condition satisfied`, ~290s — so it is not from this change.

Worth noting for review: `TestHASubnetRouterPingFailover` passed. That is the test that drives the health prober end to end, so route election still converges now that unchanged health writes are skipped. `TestHASubnetRouterFailoverBothOffline` passed here too, having failed on `main`.

juanfont (approved):

PR #3452, opened 2026-09-04
CHANGELOG: add 0.29.4 section

kradalby:
`CHANGELOG.md` had no section for the 0.29.4 patch release, so the `headscale
users rename` entry from #3442 opened a stray `### Fixes` heading at the end of
0.30.0. Fixes that ship first on `release-branch/0.29` now have somewhere to go.

Adds `## 0.29.4 (unreleased)` above 0.29.3 and moves the backported entries into
it: the extra-records filewatcher shutdown fix, the users rename fix, the HTTP
metrics collection fix, the tailsql shutdown fix, and the tvOS setup reorder.


kradalby:
Recreating from a correctly named branch.

PR #3453, opened 2026-09-04
cli: fix users rename when resolved by name

### LINKED ISSUE #3441: [Bug] renaming users by name does not work

kradalby:
`headscale users rename --name foo --new-name bar` failed with "user not found".
The command resolved the user with `ListUsers`, then built `RenameUserRequest`
with `OldId: id` — the raw `--identifier` flag, which is 0 when the user was
resolved by name. Send the matched user's ID instead.

Backport of #3442, which fixed the same bug in the HTTP client path on `main`.
The 0.29 CLI still talks gRPC, so the upstream patch does not apply.

`TestUserCommand` only ever renamed by `--identifier`, so it never saw this.
It now renames by `--name` before destroying by `--name`.

Fixes #3441

### DIFF
--- a/cmd/headscale/cli/users.go
+++ b/cmd/headscale/cli/users.go
@@ -231,7 +231,7 @@ var renameUserCmd = &cobra.Command{
 		newName, _ := cmd.Flags().GetString("new-name")
 
 		renameReq := &v1.RenameUserRequest{
-			OldId:   id,
+			OldId:   users.GetUsers()[0].GetId(),
 			NewName: newName,
 		}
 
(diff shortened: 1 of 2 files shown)


kradalby:
Recreating from a correctly named branch.

PR #3461, opened 2026-09-09
hscontrol: keep the ACME error body readable in acmeLogger

### LINKED ISSUE #3460: [Bug] acmeLogger drains the ACME error body, breaking badNonce retries and stopping certificate renewal permanently

### Current Behavior

`acmeLogger.RoundTrip` (added in #2933) reads and closes the response body of every ACME response with status >= 400, and then hands that response back to `golang.org/x/crypto/acme`:

```go
if resp.StatusCode >= http.StatusBadRequest {
	defer resp.Body.Close()

	body, _ := io.ReadAll(resp.Body)
	log.Error().Int("status_code", resp.StatusCode).Str("url", req.URL.String()).Bytes("body", body).Msg("acme request returned error")
}

return resp, nil
```

The acme client parses that body to classify the error. With an empty body, `responseError` (`acme/http.go`) cannot unmarshal the problem document, so `ProblemType` stays empty and `isBadNonce` (`acme/jws.go`) returns false. In `Client.post` (`acme/http.go`) `badNonce` is the only reason a 400 is retried, and it is also the only place that calls `clearNonces`:

```go
switch {
// Check for bad nonce before isRetriable because it may have been returned
// with an unretriable response code such as 400 Bad Request.
case isBadNonce(resErr):
	// Consider any previously stored nonce values to be invalid.
	c.clearNonces()
case !isRetriable(res.StatusCode):
	return nil, resErr
}
```

So a `badNonce` reply now terminates the renewal attempt instead of being retried, and the stale nonce pool is never cleared.

That turns into a permanent deadlock. autocert retries a failed renewal after roughly 30-60 minutes. The 400 response carries a fresh `Replay-Nonce`, which the client stores in its pool; by the time of the next attempt that nonce has expired on the CA side, so the next attempt fails with `badNonce` again. Our server was stuck in that loop for a month:

- 920 `acme request returned error` / `badNonce` log lines, ~32 per day
- exactly one line per attempt, i.e. no retry ever happened
- the certificate expired and every client dropped with `x509: certificate has expired`
- restarting headscale fixed it instantly, because a fresh process starts with an empty nonce pool and fetches a new nonce via `new-nonce`

Log sample, the same line repeating for a month:

```
2026-09-08T08:36:25Z ERR acme request returned error body="{\n  \"type\": \"urn:ietf:params:acme:error:badNonce\",\n  \"detail\": \"JWS has an invalid anti-replay nonce\",\n  \"status\": 400\n}" status_code=400 url=https://acme-v02.api.letsencrypt.org/acme/new-order
```

The ACME account, DNS, TLS-ALPN-01 reachability and the system clock were all fine — the very same setup issued a new certificate seconds after a restart.

There is some irony here: the middleware was added to make failing renewals debuggable, and it is what makes them permanent.

### Expected Behavior

`acmeLogger` logs the error response and leaves the body readable for the acme client, so `badNonce` is detected, the nonce pool is cleared and the request is retried.

### Steps To Reproduce

1. Run headscale >= v0.29 with `tls_letsencrypt_hostname` / autocert.
2. Let one ACME request fail with `urn:ietf:params:acme:error:badNonce`.
3. Every following renewal attempt fails the same way until the process is restarted, and the certificate eventually expires.

It also reproduces without a CA: serve a 400 `badNonce` problem document through `acmeLogger` and observe that the response body reaching the acme client is empty, so `acme.Error.ProblemType` is empty instead of `urn:ietf:params:acme:error:badNonce`.

### Environment

```markdown
- OS: Ubuntu 22.04
- Headscale version: v0.29.2 (code is unchanged in v0.29.3 and in main)
- golang.org/x/crypto: v0.53.0
- Tailscale version: 1.9x
```

### Runtime environment

- [x] Headscale is running as a systemd service

### Debug information

Certificate cache before and after the restart (`/var/lib/headscale/<hostname>`), the ECDSA entry had expired and was reissued within seconds of the restart with no configuration change.

### Contribution

- [x] I can write the fix, PR follows

shurshun:
- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md

Fixes #3460.

`acmeLogger.RoundTrip` drained and closed the body of every ACME error response before handing the response back to `golang.org/x/crypto/acme`. The client parses that body to classify errors, so `badNonce` was no longer recognised: `isBadNonce` returned false, `clearNonces` was never called, and `Client.post` treated the 400 as non-retriable.

The practical effect is that one `badNonce` reply stops certificate renewal for good. autocert retries every 30-60 minutes, each attempt reuses a nonce stored from the previous failed response, that nonce has already expired on the CA side, and the loop repeats. On our server that ran for a month — 920 `badNonce` log lines, exactly one per attempt, no retry ever — and ended with an expired certificate and every client dropping.

This restores the body with a fresh reader after logging it, so the acme client sees the problem document again.

The added test serves a 400 `badNonce` problem document through the middleware and reads the body afterwards. Without the fix it fails with `http: read on closed response body`.


hscontrol/app_test.go
```diff
@@ -24,3 +26,28 @@ func TestSecurityHeaders(t *testing.T) {
 	assert.Equal(t, "nosniff", h.Get("X-Content-Type-Options"))
 	assert.Equal(t, "no-referrer", h.Get("Referrer-Policy"))
 }
+
+func TestAcmeLoggerKeepsErrorBodyReadable(t *testing.T) {
```
kradalby:
```suggestion
func TestACMELoggerKeepsErrorBodyReadable(t *testing.T) {
```

hscontrol/app_test.go
```diff
@@ -24,3 +26,28 @@ func TestSecurityHeaders(t *testing.T) {
...
+	const problem = `{"type":"urn:ietf:params:acme:error:badNonce","detail":"JWS has an invalid anti-replay nonce","status":400}`
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/problem+json")
+		w.WriteHeader(http.StatusBadRequest)
+		_, _ = w.Write([]byte(problem))
+	}))
+	defer server.Close()
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, server.URL, nil)
```
kradalby:
```suggestion
	req, err := http.NewRequestWithContext(t.Context(), http.MethodGet, server.URL, nil)
```

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ keys remain all-access.
 - Fix `headscale users rename` sending the raw `--identifier` flag value instead of the matched user's identifier, so renaming by name works again [#3442](https://github.com/juanfont/headscale/pull/3442)
 - Fix tailsql not shutting down with headscale, leaving the process hanging on graceful shutdown [#3400](https://github.com/juanfont/headscale/pull/3400)
 - Fix tvOS setup instructions: install the VPN configuration before setting the coordination server URL [#3431](https://github.com/juanfont/headscale/pull/3431)
+- Fix ACME renewal stopping permanently after a `badNonce` reply, because the error logging middleware drained the response body the acme client needs to detect it [#3461](https://github.com/juanfont/headscale/pull/3461)
```
kradalby, line 67: lets move this to 0.29.4

kradalby (approved):
Some small fixups, otherwise looks good.

PR #3473, opened 2026-09-12
oidc: add option to retry if unavailable on start

### LINKED ISSUE #3462: [Feature] Lazy discovery of oidc provider
### Use case

Headscale does OIDC discovery exactly once at startup either stopping execution afterwards if `only_start_if_oidc_is_available: true` (default) or falling back to cli/web auth.

I use authentik as oidc provider which is only reachable internally via the headscale network. This poses a problem when the network map is stale and headscale can't reach authentik requiring me to restart headscale for oidc to work again.

### Description

Stop treating oidc discovery as a hard failure and try discovery again after startup when a user is trying to login. 

### Contribution

- Try oidc discovery once on startup
- When unsuccessful -> On first login make a new discovery check
- If oidc provider is genuinely down give a clear error message to the user 
- Try again on another login attempt (with backoff timeouts)
- When successful -> save result and properly setup oidc
- When 'only_start_if_oidc_is_available: true' only checking once at startup then exiting for users that prefer the current behavior

LiteApplication:
This is disabled by default to avoid changing the behavior of already-deployed instances, but can prove very useful for cases where the authentication server and headscale live on the same node that gets rebooted, and headscale starts before the authentication server (as it is the case for me).

Fixes #3462 

- [x] added unit tests
- [x] added integration tests
- [x] updated CHANGELOG.md


LiteApplication:
@copilot Fix the code for all comments in [this review thread](https://github.com/juanfont/headscale/pull/3473#pullrequestreview-5185995923).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

PR #3475, opened 2026-09-13
db: report blocking node count when destroying a user

YpNo:
Deleting a user that still owns nodes fails with a bare "user still has nodes".
The `headscale users destroy` prompt only mentions pre-auth keys, so the error
reads like something went wrong rather than a precondition. This wraps
`ErrUserStillHasNodes` with the number of nodes that block the deletion:

    deleting user: user still has nodes: 3 node(s) must be deleted or moved to another user first

`errors.Is` still matches the sentinel, so the API keeps returning 400 and the
existing tests pass unchanged.

While here, the `TestDestroyUserErrors` table now runs on Postgres as well as
SQLite. User deletion relies on foreign-key actions (cascade on `nodes.user_id`,
set-null on `pre_auth_keys.user_id`) that come from the hand-written SQLite
schema on one side and GORM AutoMigrate on the other, and nothing exercised the
Postgres side. The Postgres variant uses the existing `newPostgresTestDB`
helper and skips when no local server can be started, like the other Postgres
tests.

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md


hscontrol/db/users.go
```diff
@@ -61,7 +61,10 @@ func DestroyUser(tx *gorm.DB, uid types.UserID) error {
 	}
 
 	if len(nodes) > 0 {
-		return ErrUserStillHasNodes
+		return fmt.Errorf(
+			"%w: %d node(s) must be deleted or moved to another user first",
+			ErrUserStillHasNodes, len(nodes),
```
kradalby: Since we already have the nodes here, would be neat to get the ids as well? makes the tracking down even easier.

kradalby (approved):
One proposal

hscontrol/db/users.go
YpNo: done, the error now lists id (hostname) for each blocking node.

CHANGELOG.md
```diff
@@ -42,7 +42,7 @@ keys remain all-access.
 
 ### Changes
 
-- Deleting a user that still owns nodes now says how many nodes must be deleted or moved first
+- Deleting a user that still owns nodes now lists the nodes (ID and hostname) that must be deleted or moved first
```
nblock, line 45: Nodes can no longer be moved, please remove the last part of the sentence.

YpNo: Done

hscontrol/db/users.go
```diff
@@ -61,7 +62,15 @@ func DestroyUser(tx *gorm.DB, uid types.UserID) error {
...
 
 	if len(nodes) > 0 {
-		return ErrUserStillHasNodes
+		blocking := make([]string, len(nodes))
+		for i, node := range nodes {
+			blocking[i] = fmt.Sprintf("%d (%s)", node.ID.Uint64(), node.Hostname)
+		}
+
+		return fmt.Errorf(
+			"%w: %d node(s) must be deleted or moved to another user first: %s",
```
nblock:
```suggestion
            "%w: %d node(s) must be deleted first: %s",
```

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ keys remain all-access.
 
 ### Changes
 
+- Deleting a user that still owns nodes now lists the nodes (ID and hostname)
```
nblock:
```suggestion 
- Deleting a user that still owns nodes now lists the nodes (ID and hostname) that must be deleted first
```

kradalby: Also PR link

kradalby:
Looks like the linter is unhappy.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ keys remain all-access.
 
 ### Changes
 
+- Deleting a user that still owns nodes now lists the nodes (ID and hostname) that must be deleted first
```
kradalby, line 45: missing a link the PR.

kradalby (approved):

CHANGELOG.md
YpNo:
```suggestion
- Deleting a user that still owns nodes now lists the nodes (ID and hostname) that must be deleted first [#3475](https://github.com/juanfont/headscale/pull/3475)
```

kradalby: same here

Issue #3429, opened 2026-08-19
[Bug] able to create users with same `username`


debtquity:
### Is this a support request?

- [x] This is not a support request

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

I observe multiple users can have the same `username` if the users are later created/onboarded via OIDC.

```shell
# headscale users list
ID | Name           | Username  | Email          | Created
1  |                | debtquity |                | 2026-07-02 06:18:07
2  | Darren Rambaud | debtquity | me@example.com | 2026-08-16 03:29:59
3  |                | unused    |                | 2026-08-19 19:24:36
```

No issue observed when logging in and creating user via OIDC. However, after a restart of the headscale process. It prevented headscale from starting up:

```shell
# journalctl -u headscale.service -e
...
Aug 19 19:19:41 ... headscale-start[124195]: Error: initializing: creating new headscale: init state: initializing policy manager: validating policy user references: multiple users found: token "debtquity@" found: [ 1: debtquity, 2: debtquity,  ]
Aug 19 19:19:41 ... systemd[1]: headscale.service: Main process exited, code=exited, status=1/FAILURE
Aug 19 19:19:41 ... systemd[1]: headscale.service: Failed with result 'exit-code'.
Aug 19 19:19:44 ... systemd[1]: Stopped headscale coordination server for Tailscale.
```

Note: attempts to create or rename user via CLI will return an error:

```
# headscale users create debtquity
Error: creating user: rpc error: code = Internal desc = creating user: creating user: constraint failed: UNIQUE constraint failed: users.name (2067)

# headscale users rename -i 3 -r debtquity
Error: renaming user: rpc error: code = Unknown desc = updating user: constraint failed: UNIQUE constraint failed: users.name (2067)
```

### Expected Behavior

Should return unique name constraint error and prevent user from being created until username conflict resolved; or merge the 2 users.

### Steps To Reproduce

1. Create user via `headscale users create USERNAME`
2. Enable OIDC in headscale with `oidc.*` configuration
3. Setup policy file to reference the `USERNAME` in step 1 (if necessary, restart headscale)
```
# policy.path
{
  "groups": {
    "group:network-admin": [
      "USERNAME@"
    ]
  }
}
```
4. In IdP, create a user where the `preferred_username` claim will conflict with the `USERNAME` in step 1
5. Create user via OIDC flow in any tailscale client
6. Restart headscale process

### Environment

```markdown
[...]

PR #3476, opened 2026-09-13
cli: distinguish no match from ambiguous match when resolving users

YpNo:
`headscale users destroy` and `headscale users rename` resolve the target with
`resolveSingleUser`, which treated every result count other than one as
"multiple users match query, specify an ID". A name that matches nobody got the
same message, which points the operator in the wrong direction.

An explicit `--identifier 0` had a second problem: the flag default is -1, so 0
passed the "is a flag set" check and was sent to the API as `id=0`, which the
list endpoint treats as "no filter". The CLI then received every user and
failed with the same ambiguous-match error.

This change:

- returns "no user matches query" when the list comes back empty
- keeps "multiple users match query, specify an ID" for two or more matches
  (OIDC users can share a name, see #3429)
- rejects an explicit zero or negative `--identifier` before calling the API

`TestResolveSingleUser` now asserts the specific sentinel error for each case
instead of just "an error", and covers the `--identifier 0` regression.

The destroy prompt also now says that nodes owned by the user must be deleted
or moved first, since that is the precondition the API enforces and the prompt
only mentioned pre-auth keys.

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md


cmd/headscale/cli/users.go
```diff
@@ -74,11 +83,15 @@ func resolveSingleUser(
...
 	users := resp.JSON200.Users
-	if len(users) != 1 {
+
+	switch len(users) {
+	case 0:
+		return "", nil, errUserNotFound
+	case 1:
+		return users[0].Id, &users[0], nil
+	default:
 		return "", nil, errMultipleUsersMatch
```
kradalby: Same as with the multiple node errors, it would actually be nice to highlight which users match to help the user.

YpNo: done, the ambiguity error lists id, name, email and provider for each match.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ keys remain all-access.
 
 ### Changes
 
+- `headscale users destroy`/`rename` now report "no user matches query" when the name or ID matches nobody, instead of "multiple users match query", list the matching users when several match, and reject `--identifier 0`
```
kradalby, line 45: missing link to PR, a bit too long and windy, could be more consistent

kradalby (approved):
good, thanks, can you fixup the comment.

CHANGELOG.md
YpNo: Just to be sure, you want me to add the PR link to the changelog line for referencing ?

YpNo:
```suggestion
- `headscale users destroy`/`rename` now report "no user matches query" when the name or ID matches nobody, instead of "multiple users match query", list the matching users when several match, and reject `--identifier 0` [#3476](https://github.com/juanfont/headscale/pull/3476)
```

kradalby: yes

PR #3477, opened 2026-09-13
db: quote postgres connection parameters

YpNo:
The Postgres DSN was assembled by string concatenation:

    dbString += " password=" + cfg.Postgres.Pass

libpq's keyword/value format splits on whitespace, so a password containing a
space was cut off at the space and the remainder parsed as further parameters.
A value with a single quote or backslash broke parsing in similar ways. The
failure is silent: the connection simply fails to authenticate, or worse, a
password containing `sslmode=disable` turns TLS off.

This moves the DSN into a `postgresDSN` helper that quotes every value with
`pgQuote`, following the libpq rules: wrap in single quotes, escape `\` and `'`
with a backslash. Host, database name, user, password and a raw `ssl` value
such as `verify-full` all go through it. The `ssl` handling is unchanged:
`true` keeps the libpq default, `false` adds `sslmode=disable`, anything else
is passed through as an `sslmode` value.

The startup log line used to print the DSN before the password was appended.
It now logs the DSN built from a copy of the config with the password cleared,
so the redaction no longer depends on statement order.

`TestBuildPostgresDSN` covers the three `ssl` forms, a unix-socket host, an
omitted port and password, and a password containing a space, a quote, a
backslash and an embedded `sslmode=disable`. I also confirmed the quoted output
round-trips through pgx's `pgconn.ParseConfig` with the expected host, port,
database, user and password; that check is not committed to keep pgx an
indirect dependency.

- [x] added unit tests
- [ ] added integration tests
- [x] updated CHANGELOG.md


hscontrol/db/db.go
```diff
@@ -1157,6 +1140,48 @@ func openDB(cfg types.DatabaseConfig) (*gorm.DB, error) {
 	)
 }
 
+// postgresDSN builds a libpq keyword/value connection string. Every value is
+// quoted with [pgQuote] so a password (or host, user, database name) containing
+// spaces, quotes or backslashes is passed through intact instead of silently
+// truncating the DSN or injecting extra parameters.
+func postgresDSN(cfg types.PostgresConfig) string {
```
kradalby: Can you make this a function on the `types.PostgresConfig` so we can do `cfg.Postgres.DSN()` and `redacted.DSN()`?

It looks like if we are revisiting this, using the URL so we can use a proper URL builder, or the typed go config thing might be more meaningful than doing our own? https://github.com/lib/pq#connecting

YpNo: done, both suggestions. DSN() lives on PostgresConfig and builds a postgres:// URL via net/url. Worth mentioning that Unix socket paths go through the host query parameter because the URL form has no other way to carry them, and that you verified the output against pgx's parser.

